PR #8, opened 2016-01-26
Fixes for rds_enhanced_monitoring changes

jasonthomas:
I was testing this out and ran into the following issues.


tmichelet:
thanks for catching those mistakes!!

PR #14, opened 2017-10-30
Update handler section of the README

NBParis:


tmichelet:
From what I remember it was the default handler, is it not the case anymore?

NBParis:
Seems that it changed. As new lambda function are created with `index.handler`

tmichelet:
I think it depends on which template you start from, but mentioning this in the readme won't hurt, let's do it

PR #18, opened 2017-11-17
Correct source detection for lambda logs

NBParis:
Lambda logs were flagged as cloudwatch until now

### DIFF
--- a/Log/lambda_function.py
+++ b/Log/lambda_function.py
@@ -142,7 +142,8 @@ def awslogs_handler(s, event):
     # Get logs
     data = zlib.decompress(base64.b64decode(event["awslogs"]["data"]), 16 + zlib.MAX_WBITS)
     logs = json.loads(str(data))
-    source = logs["logGroup"]
+    #Set the source on the logs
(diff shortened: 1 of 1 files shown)


Log/lambda_function.py
```diff
@@ -142,8 +142,8 @@ def awslogs_handler(s, event):
     # Get logs
     data = zlib.decompress(base64.b64decode(event["awslogs"]["data"]), 16 + zlib.MAX_WBITS)
     logs = json.loads(str(data))
-    metadata[DD_SOURCE] = "cloudwatch"
-
+    source = logs["logGroup"]
```
tmichelet: is `logGroup` always present? Would it be safer to use `logs.get("logGroup", "DEFAULTVALUE")`?

NBParis: It should always be there from cloudwatch logs but will change as you suggest for safety.

NBParis:
@tmichelet updated (added the kinesis log source at the same time)

tmichelet:
👍

PR #22, opened 2017-12-11
Send functionname and memorysize as tag

NBParis:
We do want to have the lambda function parameters as tags instead of log attributes.


NBParis:
@tmichelet The lambda metrics are tagged with the functionname and memorysize which is why we must set those as tags and not log attributes so we can do the metrics to logs seamlessly

tmichelet:
ok thanks, 👍

PR #24, opened 2018-01-18
[Logs] Improve retry logic to catch exceptions in the socket

tmichelet:
As we're sometimes seeing connections resets, or SSLErrors

### DIFF
--- a/Log/lambda_function.py
+++ b/Log/lambda_function.py
@@ -47,11 +47,7 @@ def lambda_handler(event, context):
         )
 
     # Attach Datadog's Socket
-    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-
-    port = ssl_port
-    s = ssl.wrap_socket(s)
-    s.connect((host, port))
+    s = connect_to_datadog(host, ssl_port)
 
     # Add the context to meta
     if "aws" not in metadata:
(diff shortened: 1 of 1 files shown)


Log/lambda_function.py
```diff
@@ -62,28 +58,44 @@ def lambda_handler(event, context):
     #Add custom tags here by adding new value with the following format "key1:value1, key2:value2"  - might be subject to modifications
     metadata[DD_CUSTOM_TAGS] = "functionname:" + context.function_name+ ",memorysize:"+ context.memory_limit_in_mb
 
+    try:
+        logs = generate_logs(event)
+        for log in logs:
+            s = safe_submit_log(s, log)
+    except Exception as e:
+        print('Uncaught exception: {}'.format(str(e)))
```
ajacquemot: This exception is caught, you should change `Uncaught` to `Unexpected`. I would also print `event` for debugging purpose.

Log/lambda_function.py
```diff
@@ -62,28 +58,44 @@ def lambda_handler(event, context):
...
+    finally:
+        s.close()
```
ajacquemot: I think you should close this socket all the time because it's in the `lambda_handler` scope.

Log/lambda_function.py
```diff
@@ -62,28 +58,44 @@ def lambda_handler(event, context):
...
         elif event_type == "awslogs":
-            logs = awslogs_handler(s, event)
-
-        for log in logs:
-            send_entry(s, log)
-
+            logs = awslogs_handler(event)
     except Exception as e:
         # Logs through the socket the error
         err_message = 'Error parsing the object. Exception: {}'.format(str(e))
```
ajacquemot: Same here, I would print the event too.

Log/lambda_function.py
```diff
@@ -62,28 +58,44 @@ def lambda_handler(event, context):
...
-    finally:
-        s.close()
+        logs = [err_message]
+    return logs
 
+def safe_submit_log(s, log):
+    try:
+        send_entry(s, log)
+    except Exception as e:
+        err_message = 'Error sending the log line. Exception: {}'.format(str(e))
```
ajacquemot: I would print the log here.

tmichelet: We're actually sending it to datadog!

tmichelet:
@mnshdw would you mind explaining this?
>Ok, this will work reliably only if we can guarantee a single node can be unavailable at any given time.

Log/lambda_function.py
ajacquemot: I meant doing something like:
```
err_message = 'Exception: {} occured for log line: {}'.format(str(e), log)
```

mnshdw:
@tmichelet What I meant is that if your retry happens on a node that happens to be unavailable when you re-send the log (eg. it is being restarted), you'll face the same issue as before. This is unlikely but still possible to drop logs.

tmichelet:
We now have CW logs when we're unable to submit data, so if this happens people will be able to catch it. I'd rather merge it as is, knowing that we might not use TCP anymore in the near future

Log/lambda_function.py
tmichelet: Feedback from customer: we should not log when an error happened but we were able to retry

PR #27, opened 2018-02-22
Add source auto detection for mysql and mariadb in RDS

NBParis:
AWS released a way to push MySQL and MariaDB logs to Cloudwatch. Therefore it is now possible to collect them with the lambda function and to automatically set the source.

### DIFF
--- a/Log/lambda_function.py
+++ b/Log/lambda_function.py
@@ -216,7 +216,7 @@ def is_cloudtrail(key):
     return bool(match)
 
 def parse_event_source(event, key):
-    for source in ["lambda","elasticloadbalancing","redshift","cloudfront","kinesis", "mariadb","mysql","route53", "vpc", "rds"]:
+    for source in ["lambda","elasticloadbalancing","redshift","cloudfront","kinesis", "mariadb","mysql","apigateway", "route53", "vpc", "rds"]:
         if source in key:
             return source
     if is_cloudtrail(str(key)):
--- a/Log/lambda_function.py
+++ b/Log/lambda_function.py
@@ -216,24 +216,11 @@ def is_cloudtrail(key):
     return bool(match)
 
 def parse_event_source(event, key):
-    if "lambda" in key:
-        return "lambda"
+    for source in ["lambda","elasticloadbalancing","redshift","cloudfront","kinesis", "mariadb","mysql","route53", "vpc", "rds"]:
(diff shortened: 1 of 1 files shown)


Log/lambda_function.py
```diff
@@ -228,6 +228,12 @@ def parse_event_source(event, key):
         return "cloudfront"
     if "kinesis" in key:
         return "kinesis"
+    if "mysql" in key:
+        return "mysql"
+    if "mariadb" in key:
+        return "mariadb"
+    if "rds" in key:
+        return "rds"
```
tmichelet: could we refactor this a bit?
```
for source in [..., "mariadb", "rds", ]:
  if source in key:
    return source
```

NBParis:
@tmichelet done

tmichelet:
thanks, looks good

Log/lambda_function.py
```diff
@@ -216,18 +216,11 @@ def is_cloudtrail(key):
     return bool(match)
 
 def parse_event_source(event, key):
-    if "lambda" in key:
-        return "lambda"
+    for source in ["lambda","elasticloadbalancing","redshift","cloudfront","kinesis", "mariadb","mysql","apigateway", "route53", "vpc", "rds"]:
```
tmichelet: nit: you should install python linters, so we have a common way of formatting code. Usually we do `[x, y]` instead of `[x,y]`

NBParis:
@tmichelet Formatting done

Log/lambda_function.py
```diff
@@ -216,18 +216,11 @@ def is_cloudtrail(key):
...
 def parse_event_source(event, key):
-    if "lambda" in key:
-        return "lambda"
+    for source in ["lambda", "elasticloadbalancing", "redshift", "cloudfront", "kinesis", "mariadb", "mysql", "apigateway", "route53", "vpc", "rds"]:
+        if source in key:
+            return source
     if is_cloudtrail(str(key)):
         return "cloudtrail"
-    if "elasticloadbalancing" in key:
-        return "elb"
```
tmichelet: Is this expected? Do we want to rename `elb` into `elasticloadbalancing`?

NBParis: Indeed, our integration name is elb but AWS put elasticloadbalancing in the file path

PR #29, opened 2018-03-08
Update yaml files to add key parameter

DanielLanger:


garnermccloud:
Files are missing newlines at the end but otherwise this looks good. Were you able to test with cloudformation?

DanielLanger:
Do they need newlines at the end? Don't see that previously.

And how did you test using cloudformation in the past?

DanielLanger:
Tested both YAML templates in demo aws account and both successfully created correct stack

PR #30, opened 2018-03-12
Rename custom tag

DanielLanger:


tmichelet:
Why are we doing this?

NBParis:
@tmichelet Basically, the tag `functionname` on metrics is referencing the lambda function that generate the metric.

Whereas here the `functionname` tag was always referencing the Lambda used to forward logs which could have led to confusion.

tmichelet:
ok

PR #38, opened 2018-04-24
added functionality for custom tags from DD_TAGS environment variable

aldegoeij:
in order to pass tags like "prod", or "environment:prod" to all cloudwatch messages processed by the lambda, I've implemented similar functionality as for the DD docker container.
Using the DD_TAGS environment variable, which you can set in your CloudFormation or Terraform (like we do), as additional metadata["ddtags"].
Also minor whitespace on the '+'s.


NBParis:
Hello @dmulter, 

Indeed, sorry for forgetting about this one.
And it indeed handles it correctly in case it is not set (misread the code on my first review...)

We will do some tests and merge right after.

PR #46, opened 2018-08-03
Update function to support KMS encrypted keys

NBParis:
Add support for KMS encrypted keys

### DIFF
--- a/Log/README.md
+++ b/Log/README.md
@@ -46,7 +46,7 @@ metadata = {
 
 There are 3 possibilities to set your Datadog's API key (available in your Datadog platform):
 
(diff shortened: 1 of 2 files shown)


Log/README.md
```diff
@@ -36,17 +36,31 @@ At the top of the script you'll find a section called `#Parameters`, that's wher
...
+}
 ```
 
 - **API key**:
 
-Replace `<your_api_key>`: Your Datadog's API key is available in your platform.
-You can also set it thanks to the `DD_API_KEY` environment variable.
+There are 3 possibilities to set your Datadog's API key (available in your Datadog platform):
+
+1. **KMS Encrypted key**: Use the `DD_KMS_API_KEY` environment variable to use a KMS encrypted key. Make sure that the Lambda excution role is listed in the KMS Key user in https://console.aws.amazon.com/iam/home#encryptionKeys.
```
tmichelet: could we do `**KMS Encrypted key (Recommended)**`?

NBParis: Updated

PR #49, opened 2018-08-21
Added "proxy support" via environment variables

mspike:
Useful if you use proxy in your environment.

Please pull, thanks!
Br,
 Peter

### DIFF
--- a/Log/lambda_function.py
+++ b/Log/lambda_function.py
@@ -36,9 +36,16 @@ metadata = {
     "ddsourcecategory": "aws",
 }
 
+try:
+    host = os.environ['DD_URL']
+except Exception:
+    host = "lambda-intake.logs.datadoghq.com"
 
-host = "lambda-intake.logs.datadoghq.com"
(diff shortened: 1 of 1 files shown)


Log/lambda_function.py
```diff
@@ -36,9 +36,16 @@
     "ddsourcecategory": "aws",
 }
 
+try:
+    host = os.environ['DD_URL']
```
seittema: Just an observation, you can simplify the code by using os.getenv and setting the default if it's missing
`host = os.getenv("DD_URL", default="lambda-intake.logs.datadoghq.com")`

mspike:
Awwww, pretty awsome. Please do so and make a pull request!

NBParis:
@mspike Do you want to update the current PR or to merge this one and open a new one?

mspike:
@NBParis did it in the PR. Please check and approve.

@seittema thanks for the tip, pretty neat that we don't need an extra try & catch !

PR #57, opened 2018-10-04
Option to scrub ip addresses in forwarded logs

NBParis:
IP addresses can be considered as sensitive data and we added an option to redact them.

### DIFF
--- a/Log/lambda_function.py
+++ b/Log/lambda_function.py
@@ -59,6 +59,7 @@ try:
 except Exception:
     pass
 
+ip_regex = re.compile('\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}', re.I)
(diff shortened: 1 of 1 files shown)


Log/lambda_function.py
```diff
@@ -275,6 +280,14 @@ def send_entry(s, log_entry):
 
     # Send to Datadog
     str_entry = json.dumps(log_entry)
+    
+    #Scrub ip addresses if activated
+    if is_ipscrubbing:
+        try:
+            str_entry = re.sub('\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}',"xxx.xxx.xxx.xx",str_entry)
```
tmichelet: Could you `re.compile` this regex like we do with cloudtrail?

NBParis: @tmichelet done

PR #60, opened 2018-10-15
Fix bug where gzip would timeout

achntrl:
Found a fix on the file unzipping bug
It is not more memory intensive nor longer on files that weren't a problem


NBParis:
Works well on S3 regular files, CW events, CW logs and S3 zipped files.

Good to go @achntrl

PR #62, opened 2018-10-26
Log Functionname tag lowercase

NBParis:
The `functionname` tag that is set on metric is lowercase, so we need to do the same on the logs.


NBParis:
@mnshdw fixed the lower set according to @achntrl comment

PR #68, opened 2018-11-05
Ianb/rdsenhanced example

ian28223:
Added a script to generate a test event for the lambda function to run.
Based from same RDS Message example


rds_enhanced_monitoring/examples/README.md
```diff
@@ -0,0 +1,19 @@
+# Testing the Lambda function
+
+To test the Lambda function, you can use the script `generate_CWLogTestEvent.sh` to generate test event with current timestamps. The script requires the following files in the same directory:
+
+* `tmp_message.json` - This contains the metrics would be parsed and sent via Datadog API. Based from the [RDS Example message](../#rds-message-example)
+
+## Running the command
+```sh
+$ bash generate_CWLogTestEvent.sh
```
achntrl, line 9: The script is not portable, it doesn't work on macOS: 

There are at least 2 errors: 

```console
$ bash generate_CWLogTestEvent.sh
generate_CWLogTestEvent.sh: line 5: 1545047138N: value too great for base (error token is "1545047138N")
base64: invalid option -- w
Usage:	base64 [-hvD] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -D, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)

{
  "awslogs": {
    "data":""
  }
```

rds_enhanced_monitoring/examples/generate_CWLogTestEvent.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+SCRIPTPATH=$(cd `dirname $0` && pwd)
+
+ts=$(($(date +%s%N)/100000))
```
achntrl: I think you want:

```suggestion
ts=$(($(date +%s%N)/1000000))
```

You get the timestamp in nanosecond so it would make sense to have it in millisecond, to send it in second once it's divided by 1000 in the lambda function

rds_enhanced_monitoring/examples/generate_CWLogTestEvent.sh
```diff
@@ -0,0 +1,37 @@
...
+    "dd-rdsenhanced-filter"
+  ],
+  "logGroup": "RDSOSMetrics"
+}
+'''
+
+tmp_message=$(sed -e 's/"/\\"/g' $SCRIPTPATH/tmp_message.json)
+
+tmp_CWLogRaw=${tmp_data/__message__/$tmp_message}
+tmp_CWLogB64=$(echo ${tmp_CWLogRaw/__timestamp__/$ts} | gzip | base64 -w0)
```
achntrl: `base64 -w0` is not valid on macOS

rds_enhanced_monitoring/examples/generate_CWLogTestEvent.sh
achntrl: also `date +%s%N` is not valid on macOS

achntrl (changes requested):
Hi @ian28223, thanks for the example, it's always great to have a sample of the logs that are being processed!

We're almost there but the script is not portable: it does not work on macOS because macOS uses BSD flavored `coreutils` and there are some discrepancies between the flags. 

Hopefully, you can run 

```console
$ brew install coreutils
```

and you will have access to GNU `coreutils` with a `g` prefix. Example: `gdate` or `gbase64`

I would suggest adding something like 

```bash
#!/bin/bash

if [ $(uname) == "Darwin" ]; then
    datecmd=gdate
    base64cmd=gbase64
else
    datecmd=date
    base64cmd=base64
fi
ts=$(($(${datecmd} +%s%N)/100000))

# [...]

tmp_CWLogB64=$(echo ${tmp_CWLogRaw/__timestamp__/$ts} | gzip | ${base64cmd} -w0)
```

rds_enhanced_monitoring/examples/generate_CWLogTestEvent.sh
ian28223: I copied the changes you've made and updated my PR

PR #79, opened 2019-01-15
Allow KMS exceptions to bubble up

richid:
### What does this PR do?
If there is a problem decrypting the Datadog API key from KMS the Lambda should fail fast and alert he user. As it stands now the exception is swallowed and the user has no idea the operation failed.

### Motivation
I spent a bit of time tracking down a failing Lambda only to realize I had improper AWS IAM permissions that were preventing the decryption of the API key.

### Additional Notes
N/A

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -43,16 +43,13 @@ except Exception:
 
 # DD_API_KEY: Datadog API Key
 DD_API_KEY = "<your_api_key>"
-try:
-    if "DD_KMS_API_KEY" in os.environ:
-        ENCRYPTED = os.environ["DD_KMS_API_KEY"]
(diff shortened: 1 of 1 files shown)


NilsBunge (approved):
Thanks for submitting this PR.
Indeed any issue while setting the apikey should be blocking and raised in the lambda logs.

That said it should have been caught and raise there: https://github.com/DataDog/datadog-serverless-functions/blob/3593643f658fdb67f03d8600d47986cebe04e1f8/aws/logs_monitoring/lambda_function.py#L138
Seems that it was not the case so your change will definitely catch any issue while setting it.

richid:
Can this be merged? If not I'll just close it out.

NilsBunge:
Hey @richid 

Sorry about this, forgot to merge it. Doing it now.

richid:
No problem, thank you!

PR #81, opened 2019-01-23
Add ARN as host for S3 access logs

NilsBunge:
### What does this PR do?

S3 access logs now have the arn set as the hostname in order to properly retrieve the associated tags on that bucket.

### Motivation

Getting the same tags on logs and metrics is critical in order to link those two data sets.

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -425,4 +425,3 @@ def parse_service_arn(source, key, bucket):
         if bucket:
             return "arn:aws:s3:::"+ bucket
     return
-    return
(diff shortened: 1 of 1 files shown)


aws/logs_monitoring/lambda_function.py
```diff
@@ -420,4 +420,8 @@ def parse_service_arn(source, key):
             if len(idsplit) > 1:
                 idvalue = idsplit[1]
                 return "arn:aws:elasticloadbalancing:" + region + ":" + idvalue + ":loadbalancer/" + elbname
+    if source == "s3":
+        #For S3 access logs we use the bucket name to rebuild the arn
+        if bucket:
+            return "arn:aws:s3:::"+ bucket
```
achntrl: nit

```suggestion
            return "arn:aws:s3:::" + bucket
```

achntrl (approved):

PR #87, opened 2019-02-01
Created a HTTP client to send batch of logs over HTTP to Datadog

ajacquemot:
### What does this PR do?

Added the logic to send logs over HTTP to Datadog:
- Created a HTTP client on top of `requests`
- Created a batcher
- Refactor the TCP client to isolate the retry logic

### Motivation

Migrate the integration to th HTTP API

### Additional Notes

Need to add unit-tests

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -146,13 +146,13 @@ class DatadogHTTPClient(object):
     _POST = "POST"
     _HEADERS = {"Content-type": "application/json"}
 
-    def __init__(self, host, apiKey, timeout=1, max_retries=1, backoff=0.5):
+    def __init__(self, host, apiKey, timeout=10, max_retries=3, backoff=1):
         self._url = "https://{}/v1/input/{}".format(host, apiKey)
         self._session = requests.Session()
         self._session.headers.update(self._HEADERS)
-        self._timeout = timeout if timeout > 0 else 1
-        self._max_retries = max_retries if max_retries >= 0 else 1
-        self._backoff = backoff if backoff > 0 else 0.5
+        self._timeout = timeout
+        self._max_retries = max_retries
+        self._backoff = backoff
 
     def send(self, batch, metadata=None):
         """
@@ -162,7 +162,6 @@ class DatadogHTTPClient(object):
         body = json.dumps(batch)
         for retry in range(1 + self._max_retries):
             if retry > 0:
-                print("Retrying, retry: {}, max: {}".format(retry, self._max_retries))
                 time.sleep(self._backoff)
             try:
                 resp = self._session.post(
@@ -170,15 +169,9 @@ class DatadogHTTPClient(object):
                 )
             except Exception as e:
                 # most likely a network error
-                print("Unexpected exception: {}".format(str(e)))
                 continue
             if resp.status_code >= 500:
                 # server error
-                print(
-                    "Server error, status {}, reason: {}".format(
-                        resp.status_code, resp.reason
-                    )
-                )
                 continue
             elif resp.status_code >= 400:
                 # client error
@@ -210,10 +203,7 @@ class DatadogBatcher(object):
         each batch contains at most max_size logs.
         """
         batches = []
-        if len(logs) % self._max_size != 0:
-            nb_batchs = int(len(logs) / self._max_size) + 1
-        else:
-            nb_batchs = int(len(logs) / self._max_size)
+        nb_batchs = int(len(logs) / self._max_size) + 1
         for i in range(0, nb_batchs):
             l = i * self._max_size
             r = (i + 1) * self._max_size
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -22,13 +22,12 @@ import boto3
 
 # Change this value to change the underlying network client (HTTP or TCP),
 # by default, use the HTTP client.
-DD_USE_HTTP = os.getenv("DD_USE_HTTP", default=True)
+DD_USE_TCP = os.getenv("DD_USE_TCP", default="false").lower() == "true"
+
 
 # Define the destination endpoint to send logs to
 DD_SITE = os.getenv("DD_SITE", default="datadoghq.com")
-if DD_USE_HTTP:
-    DD_URL = os.getenv("DD_URL", default="lambda-http-intake.logs." + DD_SITE)
-else:
+if DD_USE_TCP:
     DD_URL = os.getenv("DD_URL", default="lambda-intake.logs." + DD_SITE)
     try:
         if "DD_SITE" in os.environ and DD_SITE == "datadoghq.eu":
@@ -37,6 +36,8 @@ else:
             DD_PORT = int(os.environ.get("DD_PORT", 10516))
     except Exception:
         DD_PORT = 10516
+else:
+    DD_URL = os.getenv("DD_URL", default="lambda-http-intake.logs." + DD_SITE)
 
 
 # Scrubbing sensitive data
@@ -297,12 +298,12 @@ def lambda_handler(event, context):
     logs = generate_logs(event, context, metadata)
 
     scrubber = DatadogScrubber(is_ipscrubbing)
-    if DD_USE_HTTP:
-        batcher = DatadogBatcher(128 * 1000, 1 * 1000 * 1000, 25)
-        cli = DatadogHTTPClient(DD_URL, DD_API_KEY, scrubber)
-    else:
+    if DD_USE_TCP:
         batcher = DatadogBatcher(256 * 1000, 256 * 1000, 1)
         cli = DatadogTCPClient(DD_URL, DD_PORT, DD_API_KEY, scrubber)
+    else:
+        batcher = DatadogBatcher(128 * 1000, 1 * 1000 * 1000, 25)
+        cli = DatadogHTTPClient(DD_URL, DD_API_KEY, scrubber)
 
     with DatadogClient(cli) as client:
         for batch in batcher.batch(logs):
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -94,7 +94,7 @@ class ScrubbingException(Exception):
 
 class DatadogClient(object):
     """
-    Client that implements a linear retrying logic to send a batch of logs.
+    Client that implements a exponential retrying logic to send a batch of logs.
     """
 
     def __init__(self, client, max_backoff=30):
@@ -258,8 +258,8 @@ class DatadogBatcher(object):
                 batch = []
                 size_bytes = 0
                 size_count = 0
+            # all logs exceeding max_log_size_bytes are dropped here
             if log_size_bytes <= self._max_log_size_bytes:
-                # all logs exceeding max_log_size_bytes are dropped here
                 batch.append(log)
                 size_bytes += log_size_bytes
                 size_count += 1
@@ -286,7 +286,7 @@ class DatadogScrubber(object):
                         )
                     )
             except Exception:
-                pass
+                raise Exception("could not compile rule with config: {}".format(config))
         self._rules = rules
 
     def scrub(self, payload):
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -150,7 +150,7 @@ class DatadogTCPClient(object):
     def _connect(self):
         sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         sock = ssl.wrap_socket(sock)
-        sock.create_connection((self.host, self.port), timeout=self._timeout)
+        sock.connect((self.host, self.port))
         self._sock = sock
 
     def _close(self):
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -11,6 +11,7 @@ import urllib
 import os
 import socket
 import requests
+import time
 import ssl
 import re
 from io import BytesIO, BufferedReader
@@ -145,39 +146,46 @@ class DatadogHTTPClient(object):
 
     def __init__(self, host, apiKey, max_retries=1, backoff=0.5):
         self._url = "https://{}/v1/input/{}".format(host, apiKey)
-        self._max_retries = max_retries
-        self._backoff = backoff
+        self._max_retries = max_retries if max_retries >= 0 else 1
+        self._backoff = backoff if backoff > 0 else 0.5
 
-    def send(self, content, metadata=None):
-        retries = 0
-        body = json.dumps(content)
-        while retries <= self._max_retries:
+    def send(self, batch, metadata=None):
+        """
+        Attemps to send a log with a linear retry strategy,
+        only retries on server and network errors.
+        """
+        body = json.dumps(batch)
+        for retry in range(1 + self._max_retries):
+            if retry > 0:
+                print("Retrying, retry: {}, max: {}".format(retry, self._max_retries))
+                time.sleep(self._backoff)
             try:
                 resp = requests.post(
                     self._url, headers=self._HEADERS, data=body, params=metadata
                 )
-                if resp.status_code >= 500:
-                    # server error
-                    print(
-                        "Server error, status: {}, reason {}".format(
-                            resp.status_code, resp.reason
-                        )
-                    )
-                elif resp.status_code >= 400:
-                    # client error
-                    print(
-                        "Client error, status: {}, reason {}".format(
-                            resp.status_code, resp.reason
-                        )
-                    )
-                    break
-                else:
-                    # success
-                    return
             except Exception as e:
+                # most likely a network error
                 print("Unexpected exception: {}".format(str(e)))
-            retries += 1
-        print("Could not send batch, dropping it")
+                continue
+            if resp.status_code >= 500:
+                # server error
+                print(
+                    "Server error, status {}, reason: {}".format(
+                        resp.status_code, resp.reason
+                    )
+                )
+                continue
+            elif resp.status_code >= 400:
+                # client error
+                raise Exception(
+                    "client error, status: {}, reason {}".format(
+                        resp.status_code, resp.reason
+                    )
+                )
+            else:
+                # success
+                return
+        raise Exception("max number of retries reached: {}".format(self._max_retries))
 
 
 class DatadogBatcher(object):
@@ -185,6 +193,10 @@ class DatadogBatcher(object):
         self._max_size = max_size
 
     def batch(self, logs):
+        """
+        Returns an array of batches,
+        each batch contains at most max_size logs.
+        """
         batches = []
         if len(logs) % self._max_size != 0:
             nb_batchs = int(len(logs) / self._max_size) + 1
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -124,7 +124,12 @@ class DatadogTCPClient(object):
     def send(self, logs, metadata):
         try:
             frame = "".join(
-                ["%s %s\n".format(self._api_key, json.dumps(log)) for log in logs]
+                [
+                    "%s %s\n".format(
+                        self._api_key, json.dumps(merge_dicts(log, metadata))
+                    )
+                    for log in logs
+                ]
             )
             self._sock.send(frame.encode("UTF-8"))
         except Exception as e:
@@ -207,10 +212,11 @@ def lambda_handler(event, context):
 
     if DD_USE_HTTP:
         cli = DatadogHTTPClient(DD_URL, DD_API_KEY)
+        batcher = DatadogBatcher(DD_BATCH_SIZE)
     else:
         cli = DatadogTCPClient(DD_URL, DD_PORT, DD_API_KEY)
+        batcher = DatadogBatcher(1)
 
-    batcher = DatadogBatcher(DD_BATCH_SIZE)
     with DatadogClient(cli) as client:
         for batch in batcher.batch(logs):
             try:
@@ -237,7 +243,7 @@ def generate_logs(event, context, metadata):
             str(e), event
         )
         logs = [err_message]
-    return logs
+    return normalize_logs(logs, metadata)
 
 
 def generate_metadata(context):
@@ -270,6 +276,18 @@ def generate_metadata(context):
 
 
 # Utility functions
+
+
+def normalize_logs(logs):
+    normalized = []
+    for log in logs:
+        if isinstance(log, dict):
+            batches.append(log)
+        elif isinstance(log, str):
+            batches.append({"message": log})
+        else:
+            batches.append({"message": str(log)})
+    return normalized
 
 
 def parse_event_type(event):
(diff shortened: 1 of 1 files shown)


aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +135,88 @@ def __exit__(self, ex_type, ex_value, traceback):
...
+                resp = requests.post(
+                    self._url, headers=self._HEADERS, data=body, params=metadata
+                )
+                if resp.status_code >= 500:
+                    # server error
+                    print(
+                        "Server error, status: {}, reason {}".format(
+                            resp.status_code, resp.reason
+                        )
+                    )
```
ajacquemot: continue

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +135,88 @@ def __exit__(self, ex_type, ex_value, traceback):
...
+                    break
+                else:
+                    # success
+                    return
+            except Exception as e:
+                print("Unexpected exception: {}".format(str(e)))
+            retries += 1
+        print("Could not send batch, dropping it")
```
ajacquemot: raise max number of retries

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +135,88 @@ def __exit__(self, ex_type, ex_value, traceback):
...
+                            resp.status_code, resp.reason
+                        )
+                    )
+                    break
+                else:
+                    # success
+                    return
+            except Exception as e:
+                print("Unexpected exception: {}".format(str(e)))
+            retries += 1
```
ajacquemot: sleep backoff

aws/logs_monitoring/lambda_function.py
```diff
@@ -10,13 +10,16 @@
 import urllib
 import os
 import socket
+import requests
```
ajacquemot: import time

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +135,88 @@ def __exit__(self, ex_type, ex_value, traceback):
...
+                        )
+                    )
+                elif resp.status_code >= 400:
+                    # client error
+                    print(
+                        "Client error, status: {}, reason {}".format(
+                            resp.status_code, resp.reason
+                        )
+                    )
+                    break
```
ajacquemot: raise client error

aws/logs_monitoring/lambda_function.py
ajacquemot: print retrying instead

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +135,88 @@ def __exit__(self, ex_type, ex_value, traceback):
...
-            logs = generate_logs(event, context, metadata)
-            for log in logs:
-                con = con.safe_submit_log(log, metadata)
-        except Exception as e:
-            print("Unexpected exception: {} for event {}".format(str(e), event))
+    if USE_HTTP_CLI:
+        batcher = DatadogBatcher(10)
+        client = DatadogHTTPClient(DD_URL, DD_API_KEY)
+        for batch in batcher.batch(logs):
+            client.send(batch, metadata)
```
ajacquemot: try catch

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +135,88 @@ def __exit__(self, ex_type, ex_value, traceback):
...
-        )
-
-        try:
-            logs = generate_logs(event, context, metadata)
-            for log in logs:
-                con = con.safe_submit_log(log, metadata)
-        except Exception as e:
-            print("Unexpected exception: {} for event {}".format(str(e), event))
+    if USE_HTTP_CLI:
+        batcher = DatadogBatcher(10)
```
ajacquemot: use a env var instead

aws/logs_monitoring/lambda_function.py
```diff
@@ -199,6 +240,33 @@ def generate_logs(event, context, metadata):
     return logs
 
 
+def generate_metadata(context):
```
ajacquemot, line 364: this may be broken, let's redo it

aws/logs_monitoring/lambda_function.py
```diff
@@ -58,7 +60,11 @@
 DD_CUSTOM_TAGS = "ddtags"
 DD_SERVICE = "service"
 DD_HOST = "host"
-DD_FORWARDER_VERSION = "1.2.1"
+DD_FORWARDER_VERSION = "1.3.0"
+
+# http specific
+DD_USE_HTTP = True
+DD_BATCH_SIZE = "10"
```
iksaif: This should probably not be a string.

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +138,112 @@ def __exit__(self, ex_type, ex_value, traceback):
...
+        self._max_retries = max_retries if max_retries >= 0 else 1
+        self._backoff = backoff if backoff > 0 else 0.5
+
+    def send(self, batch, metadata=None):
+        """
+        Attemps to send a log with a linear retry strategy,
+        only retries on server and network errors.
+        """
+        body = json.dumps(batch)
+        for retry in range(1 + self._max_retries):
```
iksaif: you should separate the retry logic from the request logic, can you add a method that only sends a request without the retry logic ?

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +138,112 @@ def __exit__(self, ex_type, ex_value, traceback):
...
+            if retry > 0:
+                print("Retrying, retry: {}, max: {}".format(retry, self._max_retries))
```
iksaif: is there a good reason to use print instead of logging ?

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +138,112 @@ def __exit__(self, ex_type, ex_value, traceback):
...
+                time.sleep(self._backoff)
+            try:
+                resp = self._session.post(
+                    self._url, data=body, params=metadata, timeout=self._timeout
+                )
+            except Exception as e:
+                # most likely a network error
+                print("Unexpected exception: {}".format(str(e)))
+                continue
+            if resp.status_code >= 500:
```
iksaif: `< 200 >= 300` ? (since the code under is the same)

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +138,112 @@ def __exit__(self, ex_type, ex_value, traceback):
...
+class DatadogHTTPClient(object):
+
+    _POST = "POST"
+    _HEADERS = {"Content-type": "application/json"}
+
+    def __init__(self, host, apiKey, timeout=1, max_retries=1, backoff=0.5):
+        self._url = "https://{}/v1/input/{}".format(host, apiKey)
+        self._session = requests.Session()
+        self._session.headers.update(self._HEADERS)
+        self._timeout = timeout if timeout > 0 else 1
```
iksaif: can you put the unit (timeout_second / backoff_second) ?

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +138,112 @@ def __exit__(self, ex_type, ex_value, traceback):
...
+        self._max_retries = max_retries if max_retries >= 0 else 1
+        self._backoff = backoff if backoff > 0 else 0.5
```
iksaif: optionally you could use what is proposed here https://www.peterbe.com/plog/best-practice-with-retries-with-requests

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +138,112 @@ def __exit__(self, ex_type, ex_value, traceback):
...
+        self._max_size = max_size
+
+    def batch(self, logs):
+        """
+        Returns an array of batches,
+        each batch contains at most max_size logs.
+        """
+        batches = []
+        if len(logs) % self._max_size != 0:
+            nb_batchs = int(len(logs) / self._max_size) + 1
```
iksaif: always do `+1`, worst case the last batch will be empty (`a = [], a[10:20]` works fine and return an empty list)

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +138,112 @@ def __exit__(self, ex_type, ex_value, traceback):
...
+        else:
+            nb_batchs = int(len(logs) / self._max_size)
+        for i in range(0, nb_batchs):
```
iksaif: an alternative is:

```
while i < len(logs):
      batches.append(logs[i:i + self._max_size]))
      i += self._max_size
```

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +138,112 @@ def __exit__(self, ex_type, ex_value, traceback):
...
-            for log in logs:
-                con = con.safe_submit_log(log, metadata)
-        except Exception as e:
-            print("Unexpected exception: {} for event {}".format(str(e), event))
+    metadata = generate_metadata(context)
+    logs = generate_logs(event, context, metadata)
+
+    if DD_USE_HTTP:
+        batcher = DatadogBatcher(DD_BATCH_SIZE)
+        with DatadogHTTPClient(DD_URL, DD_API_KEY) as client:
```
iksaif: I'd put this, and the other one, in two dedicated functions.

iksaif (commented):
As discussed offline, I think it would be worth adding a unit test for some of this, even if it has poor coverage and is not (yet) used by a ci.

aws/logs_monitoring/lambda_function.py
tmichelet: Also, I think I'd set a batch in size more than in number of messages

aws/logs_monitoring/lambda_function.py
tmichelet: I don't think we need to protect ourselves from a negative timeout.
Also, is a timeout of 1s a sane default?

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +138,112 @@ def __exit__(self, ex_type, ex_value, traceback):
             print("DatadogConnection exit: ", ex_type, ex_value, traceback)
 
 
+class DatadogHTTPClient(object):
+
+    _POST = "POST"
+    _HEADERS = {"Content-type": "application/json"}
+
+    def __init__(self, host, apiKey, timeout=1, max_retries=1, backoff=0.5):
```
tmichelet: Is a max_retries of 1 a sane default?

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +138,112 @@ def __exit__(self, ex_type, ex_value, traceback):
...
+        self._url = "https://{}/v1/input/{}".format(host, apiKey)
+        self._session = requests.Session()
+        self._session.headers.update(self._HEADERS)
+        self._timeout = timeout if timeout > 0 else 1
+        self._max_retries = max_retries if max_retries >= 0 else 1
```
tmichelet: same here

aws/logs_monitoring/lambda_function.py
tmichelet: Also, do we really want to log this? (maybe only when a debug flag is enabled?)

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +138,112 @@ def __exit__(self, ex_type, ex_value, traceback):
...
+            if retry > 0:
+                print("Retrying, retry: {}, max: {}".format(retry, self._max_retries))
+                time.sleep(self._backoff)
+            try:
+                resp = self._session.post(
+                    self._url, data=body, params=metadata, timeout=self._timeout
+                )
+            except Exception as e:
+                # most likely a network error
+                print("Unexpected exception: {}".format(str(e)))
```
tmichelet: Same here, do we want to print it if we were able to successfully retry later?

aws/logs_monitoring/lambda_function.py
```diff
@@ -132,50 +138,112 @@ def __exit__(self, ex_type, ex_value, traceback):
...
+
+
+class DatadogBatcher(object):
+    def __init__(self, max_size=25):
+        self._max_size = max_size
+
+    def batch(self, logs):
+        """
+        Returns an array of batches,
+        each batch contains at most max_size logs.
```
tmichelet: should we count the size of the log instead of the number of logs?

tmichelet:
thanks @ajacquemot for this first pass. Let's discuss this when you have some time

aws/logs_monitoring/lambda_function.py
ajacquemot: reused current logic, I think it's fine, WDYT ?

aws/logs_monitoring/lambda_function.py
ajacquemot: I can increase it to 10sec which is a bit long IMO

aws/logs_monitoring/lambda_function.py
ajacquemot: ?

aws/logs_monitoring/lambda_function.py
ajacquemot: I am not sure about this one, why would it be better ?

aws/logs_monitoring/lambda_function.py
ajacquemot: yes, that would work too.

aws/logs_monitoring/lambda_function.py
ajacquemot: I could but why ?

aws/logs_monitoring/lambda_function.py
tmichelet: Don't we have a limit in terms of size of the payload in the intake?

ajacquemot: we do have a limit on the total size of the request that we might want to use as well here

aws/logs_monitoring/lambda_function.py
```diff
@@ -61,78 +72,158 @@
...
+                        )
+                        for log in logs
+                    ]
+                )
+            )
+            self._sock.send(frame.encode("UTF-8"))
         except Exception as e:
-            # retry once
+            # most likely a network error, reset the connection
             if self._sock:
```
ajacquemot: not needed

aws/logs_monitoring/lambda_function.py
```diff
@@ -10,17 +10,26 @@
...
+
+# Client configuration
+DD_USE_HTTP = True
+
 # Proxy
 # Define the proxy endpoint to forward the logs to
 DD_SITE = os.getenv("DD_SITE", default="datadoghq.com")
-DD_URL = os.getenv("DD_URL", default="lambda-intake.logs." + DD_SITE)
+if DD_USE_HTTP:
+    DD_URL = os.getenv("DD_URL", default="lambda-http-intake.logs." + DD_SITE)
```
ajacquemot: DD_BATCH_SIZE

iksaif (approved):
Looks good :)

aws/logs_monitoring/lambda_function.py
```diff
@@ -30,6 +39,8 @@
         DD_PORT = int(os.environ.get("DD_PORT", 10516))
 except Exception:
     DD_PORT = 10516
+
```
tmichelet, line 48: I think we should use port 443 for the http lambda, what do you think?

aws/logs_monitoring/lambda_function.py
```diff
@@ -464,5 +630,12 @@ def parse_service_arn(source, key, bucket, context):
             filesplit = filename.split("_")
             if len(filesplit) == 6:
                 clustername = filesplit[3]
-                return "arn:aws:redshift:" + region + ":" + accountID + ":cluster:" + clustername
+                return (
```
tmichelet: prefer `"arn:aws:redshift:%s:%s:%s" % (region, accountId, clustername)`

aws/logs_monitoring/lambda_function.py
```diff
@@ -61,78 +72,198 @@
...
+        for log in logs:
+            log_size_bytes = self._sizeof_bytes(log)
+            if (
+                len(batch) >= self._max_size_count
+                or size_bytes + log_size_bytes > self._max_size_bytes
+            ) and len(batch) > 0:
+                batches.append(batch)
+                batch = []
+                size_bytes = 0
+            if log_size_bytes <= self._max_log_size_bytes:
```
tmichelet, line 273: prefer `else if`, it is less bugprone

aws/logs_monitoring/lambda_function.py
```diff
@@ -61,78 +72,198 @@
...
+        is not strictly greater than max_size_bytes.
+        All logs strictly greater than max_log_size_bytes are dropped.
+        """
+        batches = []
+        batch = []
+        size_bytes = 0
+        for log in logs:
+            log_size_bytes = self._sizeof_bytes(log)
+            if (
+                len(batch) >= self._max_size_count
```
tmichelet: why recomputing len(batch) for each log? We could keep a counter. I think it'd be clearer as we'd have the same logic for count and size with different increments and limits

aws/logs_monitoring/lambda_function.py
```diff
@@ -61,78 +72,198 @@
...
+                or size_bytes + log_size_bytes > self._max_size_bytes
+            ) and len(batch) > 0:
+                batches.append(batch)
+                batch = []
+                size_bytes = 0
+            if log_size_bytes <= self._max_log_size_bytes:
+                batch.append(log)
+                size_bytes += log_size_bytes
+        if len(batch) > 0:
```
tmichelet: here we are missing some edge cases, like what happens when a single log is bigger than the batch size

aws/logs_monitoring/lambda_function.py
```diff
@@ -61,78 +72,198 @@
...
-        self._sock = self._connect()
+        self._client.connect()
         return self
 
     def __exit__(self, ex_type, ex_value, traceback):
+        self._client.close()
+        return self
+
+
+class DatadogTCPClient(object):
```
tmichelet, line 135: I would avoid refactoring the TCP client, I'd keep the logic untouched if possible. 
Why? Because by refactoring it we risk introducing bugs, while we simply want to deprecate it. Also when we'll touch common logic we'll need to test twice, which we might not do so we might introduce regressions

aws/logs_monitoring/lambda_function.py
ajacquemot: `DD_PORT` is not used for HTTP, this part should be on the TCP section actually

aws/logs_monitoring/lambda_function.py
ajacquemot: Agreed but changes are safe, I am quite confident that nothing will break here and the goal is to fully switch to HTTP, TCP will only be used as a backup if needed.

aws/logs_monitoring/lambda_function.py
ajacquemot: the if is here on purpose, to tackle for example this https://github.com/DataDog/datadog-serverless-functions/pull/87/files#r258610050

aws/logs_monitoring/lambda_function.py
ajacquemot: this is caught here https://github.com/DataDog/datadog-serverless-functions/pull/87/files#diff-618731ddf1c446ef45d09613cb4b189bR248, the message is drop, it's also explained here https://github.com/DataDog/datadog-serverless-functions/pull/87/files#diff-618731ddf1c446ef45d09613cb4b189bR234 but I can add an additional comment

aws/logs_monitoring/lambda_function.py
ajacquemot: sounds good to me, I did not really touch this logic, it's just been reformatted by my linter

aws/logs_monitoring/lambda_function.py
tmichelet: Worth adding a comment then?

aws/logs_monitoring/lambda_function.py
tmichelet: ok then it's worth adding comments

aws/logs_monitoring/lambda_function.py
```diff
@@ -53,86 +62,220 @@
...
+        batches = []
+        batch = []
+        size_bytes = 0
+        size_count = 0
+        for log in logs:
+            log_size_bytes = self._sizeof_bytes(log)
+            if (
+                size_count >= self._max_size_count
+                or size_bytes + log_size_bytes > self._max_size_bytes
+            ) and size_count > 0:
```
achntrl: nits:

```python
size_count > 0 and 
(size_count >= self._max_size_count
or size_bytes + log_size_bytes > self._max_size_bytes)
```

achntrl (approved):

aws/logs_monitoring/lambda_function.py
```diff
@@ -53,86 +63,217 @@
...
         self._sock = None
 
     def _connect(self):
-        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-        s = ssl.wrap_socket(s)
-        s.connect((self.host, self.port))
-        return s
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock = ssl.wrap_socket(sock)
+        sock.connect((self.host, self.port))
```
joyzhang-dd, line 150: connect can fail with an exception, don't we want to retry also in case of connection failure? also it's be nice to specify a timeout

aws/logs_monitoring/lambda_function.py
```diff
@@ -53,86 +63,217 @@
...
+
+    def send(self, logs):
         try:
-            self.send_entry(log, metadata)
+            frame = self._scrubber.scrub(
+                "".join(
+                    ["{} {}\n".format(self._api_key, json.dumps(log)) for log in logs]
+                )
+            )
+            self._sock.send(frame.encode("UTF-8"))
```
joyzhang-dd: send can sometime return without sending all the bytes, you need to loop on the return value to ensure everything is sent. Or use sendall

aws/logs_monitoring/lambda_function.py
```diff
@@ -55,98 +74,231 @@
...
+    Client that sends a batch of logs over TCP.
+    """
+
+    def __init__(self, host, port, api_key, scrubber, timeout=5):
         self.host = host
         self.port = port
-        self.api_key = ddApiKey
+        self._api_key = api_key
+        self._scrubber = scrubber
+        self._timeout = timeout
```
joyzhang-dd: _timeout is not used anywhere

aws/logs_monitoring/lambda_function.py
```diff
@@ -55,98 +74,230 @@
...
+        for config in configs:
+            try:
+                if os.environ.get(config.name, False):
+                    rules.append(
+                        ScrubbingRule(
+                            re.compile(config.pattern, re.I), config.placeholder
+                        )
+                    )
+            except Exception:
+                pass
```
iksaif: log the exception here

ajacquemot: raise an exception instead, we should not forward the logs if we can't build the rule, the user should fix the pattern instead, thoughts ?

gaetan-deputier (approved):
We might want to add a comment about the retry logic (and how it's controlled by the lambda timeout)

achntrl (approved):

achntrl:
I find the Exception logic a bit hard to read but I don't have a better solution right now

PR #88, opened 2019-02-01
Add host for cloudfront and redshift logs

NilsBunge:
### What does this PR do?

Add the ARN as the hostname for cloudfront and redshift logs

### Motivation

As tags collection is soon to be added for redshift and cloudfront the host needs to be set.

### Additional Notes

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -191,7 +191,7 @@ def generate_logs(event, context, metadata):
         # Route to the corresponding parser
         event_type = parse_event_type(event)
         if event_type == "s3":
-            logs = s3_handler(event, metadata)
+            logs = s3_handler(event, context, metadata)
         elif event_type == "awslogs":
             logs = awslogs_handler(event, context, metadata)
         elif event_type == "events":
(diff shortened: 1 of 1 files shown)


aws/logs_monitoring/lambda_function.py
```diff
@@ -432,4 +432,36 @@ def parse_service_arn(source, key, bucket):
...
+        #1. We extract the cloudfront id  from the filename
+        #2. We extract the AWS account id from the lambda arn
+        #3. We build the arn
+        namesplit = key.split("/")
+        if len(namesplit) > 0:
+            filename = namesplit[len(namesplit)-1] 
+            #(distribution-ID.YYYY-MM-DD-HH.unique-ID.gz)
+            filenamesplit = filename.split(".")
+            if len(filenamesplit) > 3:
+                distributionID = filenamesplit[len(filenamesplit)-4].lower()
```
mnshdw, line 446: filenamesplit[0] ?

mnshdw (approved):

aws/logs_monitoring/lambda_function.py
NilsBunge: Yes.
The file path is `bucket-name.s3.amazonaws.com/optional-prefix/distribution-ID.YYYY-MM-DD-HH.unique-ID.gz`

So the first thing we do is to extract `distribution-ID.YYYY-MM-DD-HH.unique-ID.gz` and then we get the `distribution-ID` in the first element.

PR #90, opened 2019-02-05
Added kinesis stream event handler for kinesis streams containing logs.

Chukobyte:
### What does this PR do?

This PR is to add an additional event handler for kinesis streams

### Motivation

In my current organization, we needed a way to stream logs from a kinesis stream into datadog for log collection.

### Additional Notes

Please give feed back on how I should properly handle the `metadata` dictionary as tags aren't being set currently in kinesis_handler()

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -198,6 +198,8 @@ def generate_logs(event, context, metadata):
             logs = cwevent_handler(event, metadata)
         elif event_type == "sns":
             logs = sns_handler(event, metadata)
+        elif event_type == "kinesis":
+            logs = kinesis_handler(event)
     except Exception as e:
         # Logs through the socket the error
         err_message = "Error parsing the object. Exception: {} for event {}".format(
@@ -212,10 +214,13 @@ def generate_logs(event, context, metadata):
 
 def parse_event_type(event):
     if "Records" in event and len(event["Records"]) > 0:
-        if "s3" in event["Records"][0]:
+        event_records = event["Records"][0]
+        if "s3" in event_records:
             return "s3"
-        elif "Sns" in event["Records"][0]:
+        elif "Sns" in event_records:
             return "sns"
+        elif "kinesis" in event_records["eventName"]:
+            return "kinesis"
 
     elif "awslogs" in event:
         return "awslogs"
@@ -358,6 +363,20 @@ def sns_handler(event, metadata):
         structured_line = ev
         yield structured_line
 
+# Handle kinesis stream events
+def kinesis_handler(event):
+    kinesis = boto3.client("kinesis")
+    event_records = event["Records"][0]
+    kinesis_record = event_records["kinesis"]
+    kinesis_data = kinesis_record["data"]
+    data = ""
+    with gzip.GzipFile(
+        fileobj=BytesIO(base64.b64decode(kinesis_data))
+    ) as decompress_stream:
+        # Reading line by line avoid a bug where gzip would take a very long time (>5min) for
+        # file around 60MB gzipped
+        data = "".join(BufferedReader(decompress_stream))
+    yield {"message": data}
 
 def merge_dicts(a, b, path=None):
     if path is None:
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -366,7 +366,10 @@ def sns_handler(event, metadata):
 # Handle kinesis stream events
 def kinesis_handler(event):
     kinesis = boto3.client("kinesis")
+    metadata[DD_SOURCE] = "kinesis"
     event_records = event["Records"][0]
+    # Set service to kinesis stream name
+    metadata[DD_SERVICE] = event_records["eventSourceARN"].split(":stream/")[-1]
     kinesis_record = event_records["kinesis"]
     kinesis_data = kinesis_record["data"]
     data = ""


NilsBunge:
Hello @Chukobyte 

Thanks a lot for this and for helping us improve the Lambda function.
We will definitely review the PR and let you know our feedback.

Meanwhile, would you have some samples of logs coming from the Kinesis stream when they reach the lambda? (Just so we get an idea of the object format).
This would basically be the `event` object.

One thing that is important to keep in mind is that being able to detect where the logs are coming from (elb, s3 access logs, lambda, rds, ...) is very important in order to have those logs properly processed in Datadog. Which is why one of the key feature here would be to find a way to leverage the structure of the Kinesis object to identify those.

As you can see in [here](https://github.com/Chukobyte/datadog-serverless-functions/blob/master/aws/logs_monitoring/lambda_function.py#L405-L433) we tend to leverage either the event structure itself or the key (filepath or log group) of the log to identify it and it would be amazing if we could do the same from logs from Kinesis streams.

Chukobyte:
Hey @NBParis,

Thanks for the quick response!

> One thing that is important to keep in mind is that being able to detect where the logs are coming from (elb, s3 access logs, lambda, rds, ...) is very important in order to have those logs properly processed in Datadog. Which is why one of the key feature here would be to find a way to leverage the structure of the Kinesis object to identify those.

There doesn't appear to be a way to determine the source of where the logs are coming from when looking at the kinesis `event` object.  I can understand how this isn't ideal as we would only be able to determine that the logs came from a kinesis stream and not the source that placed the logs in the kinesis stream in the first place.

I've pasted both how the logs are appearing in JSON format and the contents of the event object.

**Sample Logs**
{
    "messageType": "DATA_MESSAGE",    
    "owner": "000000000000",
    "logGroup": "/aws/lambda/datadog-logs-cloudwatch-log-group-test",
    "logStream": "2019/02/06/[$LATEST]deff60c8da7a44e38e4e3c9ced765f20",
    "subscriptionFilters": [
        "datadog-logs-cloudwatch-log-group-test_filter"
    ],
    "logEvents": [
        {
            "id": "34554617487631864244591407702399056552361807747389784064",
            "timestamp": 1549482637464,
            "message": "log group 1 test...\n"
        },
        {
            "id": "34554617487631864244591407702399056552361807747389784065",
            "timestamp": 1549482637464,
            "message": "END RequestId: c516831c-6e9a-4ef2-ab5f-626ede183311\n"
        },
        {
            "id": "34554617487654164989789938325540592270634456108895764482",
            "timestamp": 1549482637465,
            "message": "REPORT RequestId: c516831c-6e9a-4ef2-ab5f-626ede183311\tDuration: 0.44 ms\tBilled Duration: 100 ms \tMemory Size: 128 MB\tMax Memory Used: 20 MB\t\n"
        }
    ]
}

**Event Object**
{u'Records': [{u'eventVersion': u'1.0', u'eventID': u'shardId-000000000000:49592624414945101230138183359391799201880779653933694978', u'kinesis': {u'approximateArrivalTimestamp': 1549383154.8, u'partitionKey': u'230d91b032cbca13cf2a8060289f82a2', u'data': u'H4sIAAAAAAAAAI1Qy07DMBD8lcriWCt+P3KLRKiQ4NJEXKCq7NgJkfIoiUOFqv47ToE72tPuzoxm5gJ6P8+m8eXXyYMU3GdldnzOiyLb5WALxvPgp3gmVEohGBIUiXjuxmY3jcspfhJznpPO9NaZxJlg3NjA+J5h1Y2LO5tQva87bFY8DH4OP/wiTN70qzTCOkEkQTx5vXvKyrwoD1Y6qTWy3PmaGVdb4Vzcha2IYljLKDEvdq6m9hTacXhou+CnGaSv4H8WjvWNAQ43K/mnH8LKvoDWRUeUcU6oVkpTjimlag3OKNWCKcyppJxyTeJdaYbXiVhCkIquQhvrDKaPzWDONFUUc0IE3f7VHOWLMtuXm73/WCL00aUbjBxW3ihIsDWQKUugqQmGskZe1qZinLDNSwwYo6ab347eBnA9XL8B2IzDUL8BAAA=', u'kinesisSchemaVersion': u'1.0', u'sequenceNumber': u'49592624414945101230138183359391799201880779653933694978'}, u'invokeIdentityArn': u'arn:aws:iam::000000000000:role/role-dev-cxp', u'eventName': u'aws:kinesis:record', u'eventSourceARN': u'arn:aws:kinesis:$REGION:000000000000:stream/logstreaming-dev-cxp', u'eventSource': u'aws:kinesis', u'awsRegion': u'$REGION'}]}

NilsBunge:
Hello @Chukobyte ,

Just wanted to let you know that we have not forgotten about this and that we are working on it. 
A big part of the work we are doing right now is to include as much of the functionalities we have for S3 buckets and CW log group for Kinesis stream as well which is why it is still not released.

The main idea is to benefit from the same functionalities in terms of source detection and metadata for Kinesis as well.

Chukobyte:
@NBParis Thanks for providing an update.  Please let me know if you need more information or changes from my end.

NilsBunge:
Hello @Chukobyte , so we realised that getting Kinesis plugged into the Lambda was properly not the best way to proceed and are working to have a better way to manage Kinesis integration for logs.

So why is this not the best way to proceed?

The issue with Kinesis is that you can have any type of Data coming in very different formats and it becomes very hard to get the same logic as we do currently to identify the log source and to properly tag the logs.
Currently we have a lot of logic in the code to detect the source and depending on this set specific metadata to the logs (like the hostname) and all this logic do not work anymore when using Kinesis stream (as we cannot rely on the log group or S3 file name anymore).
Which is why we are building a new integration dedicated to Kinesis streams.

Now i'll definitely leave your PR open until we release this new integration for other users that might have the same problem and are willing to use the workaround you found to make it work.

Once again, thanks a lot for submitting this PR, the idea was great and was definitely worth exploring and working in a lot of cases but unfortunately not all of those we would like to cover.

jmc-rival:
I opened a similar PR #108 but it assumes that CloudWatch is the stream producer which I guess is different here because CloudWatch events have fields like logGroup and such. Given that, we are safe to just unwrap the kinesis records and then pass the log processing on to the cloudwatch log handler.

Chukobyte:
> Hello @Chukobyte , so we realised that getting Kinesis plugged into the Lambda was properly not the best way to proceed and are working to have a better way to manage Kinesis integration for logs.
> 
> So why is this not the best way to proceed?
> 
> The issue with Kinesis is that you can have any type of Data coming in very different formats and it becomes very hard to get the same logic as we do currently to identify the log source and to properly tag the logs.
> Currently we have a lot of logic in the code to detect the source and depending on this set specific metadata to the logs (like the hostname) and all this logic do not work anymore when using Kinesis stream (as we cannot rely on the log group or S3 file name anymore).
> Which is why we are building a new integration dedicated to Kinesis streams.
> 
> Now i'll definitely leave your PR open until we release this new integration for other users that might have the same problem and are willing to use the workaround you found to make it work.
> 
> Once again, thanks a lot for submitting this PR, the idea was great and was definitely worth exploring and working in a lot of cases but unfortunately not all of those we would like to cover.

@NBParis Great thanks for taking the time out to review.  I'm glad to hear that you guys are building an integration for kinesis streams!

NilsBunge:
Hello @Chukobyte just wanted to let you know that support for the Cloudwatch logs that are coming from Kinesis has been added.

So you should now be able to use the current version of the Lambda.

We went ahead [with the changes](https://github.com/DataDog/datadog-serverless-functions/pull/108) suggested by @jmc-rival  but I really wanted to thank you for submitted this PR and helping us improve the Datadog Lambda function.

I'll go ahead and close this PR but once again, thanks for your contribution.

PR #91, opened 2019-02-26
Scrub email address if activated

dharada1:
### What does this PR do?

scrub email address in sql if activated.

inspired from https://github.com/DataDog/datadog-serverless-functions/pull/57

### Motivation

We wanted to mask email address in sql execution logs.

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -62,7 +62,7 @@ cloudtrail_regex = re.compile(
     "\d+_CloudTrail_\w{2}-\w{4,9}-\d_\d{8}T\d{4}Z.+.json.gz$", re.I
 )
 ip_regex = re.compile("\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}", re.I)
-email_regex = re.compile("(\"|\')[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+(\"|\')", re.I)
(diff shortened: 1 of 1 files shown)


NilsBunge:
Thanks @dharada1 for creating this PR and helping us improve the Datadog Lambda function.
We will make sure to review it and make any suggestion if we have any.

One thing that will probably come in the future is the ability to define custom pattern directly instead of having a list of fixed one as we are doing today.

aws/logs_monitoring/lambda_function.py
```diff
@@ -57,6 +62,8 @@
     "\d+_CloudTrail_\w{2}-\w{4,9}-\d_\d{8}T\d{4}Z.+.json.gz$", re.I
 )
 ip_regex = re.compile("\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}", re.I)
+email_regex = re.compile("(\"|\')[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+(\"|\')", re.I)
```
NilsBunge: Did some test and it seems that you would need to remove `(\"|\')` at the begining and at the end of the regex to have this working.

NilsBunge (changes requested):
See comment on the regex itself. Seems that it needs to be slightly changed to properly redact email addresses.

That said, otherwise it works perfectly, thanks for doing this.

NilsBunge (approved):

NilsBunge:
Hey there,

Took the liberty to do the change myself. We will take care of merging this.
Once again thanks for your contribution to improve this lambda function.

PR #93, opened 2019-03-06
allows for capitalized elbname for hostname matching

le-he-hoo:
### What does this PR do?

Removes `.lower()` from being applied on the elb name when building the ARN string.  

### Motivation

Support case where hostname was not being properly associated with the ARN when the elbname included capitalization.


NilsBunge (approved):
LGTM

PR #97, opened 2019-03-15
Extract RDS instance name and log name

sj26:
RDS databases can log to CloudWatch Logs now. The log group name includes the database instance name and the log name. Extract them and annotate into metadata during log submission. Populating `host` also adds extra meta-data to the logs once ingested.

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -85,6 +85,8 @@ if "DD_MULTILINE_LOG_REGEX_PATTERN" in os.environ:
         "^{}".format(DD_MULTILINE_LOG_REGEX_PATTERN)
     )
 
+rds_regex = re.compile("/aws/rds/instance/(?P<host>[^/]+)/(?P<name>[^/]+)")
+
 DD_SOURCE = "ddsource"
 DD_CUSTOM_TAGS = "ddtags"
 DD_SERVICE = "service"
@@ -508,6 +510,16 @@ def awslogs_handler(event, context, metadata):
     # Set host as log group where cloudwatch is source
     if metadata[DD_SOURCE] == "cloudwatch":
         metadata[DD_HOST] = aws_attributes["aws"]["awslogs"]["logGroup"]
+
+    # When parsing rds logs, use the cloudwatch log group name to derive the
+    # rds instance name, and add the log name of the stream ingested
+    if metadata[DD_SOURCE] == "rds":
+        match = rds_regex.match(logs["logGroup"])
+        if match is not None:
+            metadata[DD_HOST] = match.group("host")
+            metadata[DD_CUSTOM_TAGS] = (
+                    metadata[DD_CUSTOM_TAGS] + ",logname:" + match.group("name")
+            )
 
     # For Lambda logs we want to extract the function name,
     # then rebuild the arn of the monitored lambda using that name.
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -520,6 +520,9 @@ def awslogs_handler(event, context, metadata):
             metadata[DD_CUSTOM_TAGS] = (
                     metadata[DD_CUSTOM_TAGS] + ",logname:" + match.group("name")
             )
+            # We can intuit the sourcecategory in some cases
+            if match.group("name") == "postgresql":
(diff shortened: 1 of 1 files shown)


NilsBunge:
Hello @sj26 ,

Thanks a lot for your contribution and helping us improve the integration with AWS RDS logs.

Regarding the host value, I would like to triple check that this would match the value that are automatically set on metrics when those are pulled. 

Also regarding the `logname` custom tag, do you know if this would match the `dbclusteridentifier` or the `dbname` one that is also set on metrics?

The main idea behind is to make sure that you can easily link metrics and logs from the same service.
I'll make sure to test internally what you suggested to see if the tags and value are matching and provide my comments here.

Once again, thanks a lot for your contribution.

sj26:
Thanks!

My understanding is that this log group pattern is named after the database instance, not the cluster, and the "logname" portion is more like "server", "error", "query" etc. — in our case the logname is "postgres".

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_LogAccess.Concepts.PostgreSQL.html#USER_LogAccess.PostgreSQL.PublishtoCloudWatchLogs

Instead of host we could use dbinstance or something, if that's conventional.

There is also a different log group name pattern which starts `/aws/rds/cluster/...` which is for aurora database clusters:

https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/AuroraMySQL.Integrating.CloudWatch.html#AuroraMySQL.Integrating.CloudWatch.Monitor

NilsBunge:
Hello @sj26 ,

First of all sorry for the delay on this and thanks a lot for the details you provided.

## logname attribute

That's a bit of a tricky situation here as in the `postgresql` example it would be a great benefit to have it stored as the `source` of the log as the Postgresql integration would then automatically parse the logs.
Now the value in the log group can also have a totally different meaning (such as `slowquery ` for instance).

So a first step could be to set this as the `sourcecategory` value instead of `logname` if you don't mind changing this.

## hostname attribute

Regarding RDS clusters, the hostname we would be looking for would actually be `arn:aws:rds:<region>:<accountid>:db:<instance>`. This would enable the addition of tags to the logs if metrics are collected for that instance as well (thanks to the Datadog RDS metric integration).

Something that we do to get the region and account id is actually a trick of using the own lambda forwarder region and account id as shown in this [example](https://github.com/DataDog/datadog-serverless-functions/blob/359a45c96882361c375aab3f2d06433d8c2a62a7/aws/logs_monitoring/lambda_function.py#L311-L328).

So the idea would be to do the same for RDS to retrieve the full arn and set it as the hostname.

Would you mind updating the hostname value with those changes?

Once again, thanks a lot for your help and contribution and let us know if you want us to help you by doing some commits directly on your PR.

NilsBunge:
Hello @sj26 ,

Just wanted to check in to see if you wanted to do the suggested update yourself or if you wanted us to help on this one.

I would say that changing `logname` to `sourcecategory` and solving the conflicts should do the trick as we can keep the hostname as it is for now until we find a better way to rebuild the ARN for RDS logs.

What are your thoughts on this?

Once again, thanks for submitting this PR.

sj26:
Hi @NBParis! Sorry for the delay, this hasn't gotten to the top of my list for a while.

hostname — definitely understand where this is coming from, but I don't have enough context to derive the region and account id in here. The other examples all derive region and account from log name, but they are not present for the rds log group name. They might be present in the lambda context sometimes depending on the event being processed, but I'm unsure. Could you show me how to find these?

logname — I didn't use sourcecategory to begin with because this name is sometimes "postgresql" because that's the default postgres log file name, but it could also be "upgrade", or for mysql instances it could be "slowquery", "error", "audit", "general" etc. We could probably guess a sourcecategory from the name of the log file, and sometimes like in the "postgresql" case it does line up, but I don't think they're a straight 1:1 so shouldn't be directly mapped. We're better off relying on datadog getting that information elsewhere and correlating based on host (which our instance seems to do quite nicely, currently).

sj26:
I've added a little bit which will add `sourcecategory:postgresql` when `logname` is `postgresql`

NilsBunge:
@sj26 thanks for the changes. 
This looks good and I'll make sure it gets merged this week.

Once again, thanks for your contribution and helping us improve the Lambda forwarder.

NilsBunge (approved):

PR #98, opened 2019-03-15
Add Azure function that reads from Blob Storage

achntrl:
### What does this PR do?

Create an Azure function that sends events from Azure blob storage
As there was duplicated code, I created a shared folder with shared code

I would have prefered to split the shared code in multiple function, but that
would have required the clients to copy paste multiple files, which would be tedious and error prone. 

The function reads the blob file, splits lines on `\n` and sends each line as
a log entry to Datadog

### Motivation

What inspired you to submit this pull request?

### Additional Notes


azure/activity_logs_monitoring/README.md
```diff
@@ -19,9 +19,19 @@ The provided Node.js script must be deployed into your Azure Functions service.
...
 ### 2. Provide the code
 
-- Copy paste the code of the [Datadog-Azure function](./index.js)
+#### 2.1 Shared library
+
+- Select your Function App and go to `Function app settings` and copy paste the content of [`host.json`](../host.json).
+- Go to the `Platform features` and click on `Advanced tools (Kudu)` under the Development tools section
+- In the menu bar, click on Tools > Zip Push Deploy
+- Create a folder name `Shared` by clicking on the "+" next to `/wwwroot`
+- Click on Shared, add a file name `client.js`, and copy paste the content of [shared/client.js](../shared/client.js) then save.
```
euzuro:
```suggestion
- Click on Shared, add a file name `client.js`, copy paste the contents of [shared/client.js](../shared/client.js), and then save.
```

azure/activity_logs_monitoring/README.md
```diff
@@ -19,9 +19,19 @@ The provided Node.js script must be deployed into your Azure Functions service.
...
+
+#### 2.2 Azure function
+
+- Copy paste the code of the [Datadog-Azure function](./index.js).
 - In the `Integrate` part, `Event Hub Cardinality` must be set to `Many`.
-- In the `Integrate` part,  set the `Event Parameter Name` to `eventHubMessages`
+- In the `Integrate` part,  set the `Event Parameter Name` to `eventHubMessages`.
```
euzuro: opt: could use sub-items:
```
- In the `Integrate` part:
  - `Event Hub Cardinality` must be set to `Many`.
  - Set the `Event Parameter Name` to `eventHubMessages`
```

azure/shared/client.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+    var resourceId = record.resourceId.toLowerCase().split('/');
+    if (resourceId.length > 2) {
+        metadata.tags.push('subscription_id:' + resourceId[2]);
+    }
+    if (resourceId.length > 4) {
+        metadata.tags.push('resource_group:' + resourceId[4]);
+    }
+    if (resourceId.length > 6) {
+        metadata.source = resourceId[6].replace('microsoft.', 'azure.');
```
euzuro, line 126: not sure if you want/need a null/undefined check on `resourceId[6]`?

e.g.
```
const x = [,,,,,,,]; 
// x.length == 7
// x[6].replace() => exception thrown
```

azure/shared/client.js
```diff
@@ -0,0 +1,151 @@
...
+    record['ddsourcecategory'] = DD_SOURCE_CATEGORY;
+    record['service'] = DD_SERVICE;
+    record['ddtags'] = metadata.tags
+        .concat([
+            DD_TAGS,
+            'forwardername:' + context.executionContext.functionName
+        ])
+        .filter(Boolean)
+        .join(',');
+    return record;
```
euzuro, line 101: function that modifies first param & returns it... kind of a code smell.

azure/blobs_logs_monitoring/index.js
```diff
@@ -0,0 +1,55 @@
...
+            }). Please upgrade your ${
+                VERSION > client.VERSION ? 'client library' : 'function'
+            }.`
+        );
+    }
+
+    if (
+        client.DD_API_KEY === '<DATADOG_API_KEY>' ||
+        client.DD_API_KEY === '' ||
+        client.DD_API_KEY === undefined
```
euzuro: could be just `!client.DD_API_KEY`

azure/blobs_logs_monitoring/README.md
```diff
@@ -0,0 +1,51 @@
+# Datadog-Azure function
+
+The Datadog-Azure function is used to forward Azure logs to Datadog from new blob files added in
+a storage account. The function reads the file, split
```
euzuro: Looks like an unfinished sentence here

azure/activity_logs_monitoring/index.js
```diff
@@ -1,154 +1,51 @@
...
+            }.`
+        );
+    }
+
+    if (
+        client.DD_API_KEY === '<DATADOG_API_KEY>' ||
+        client.DD_API_KEY === '' ||
+        client.DD_API_KEY === undefined
+    ) {
+        context.log(
```
ajacquemot:
```suggestion
        context.log.error(
```

azure/activity_logs_monitoring/index.js
```diff
@@ -1,154 +1,51 @@
...
+            'You must configure your API key before starting this function (see ## Parameters section)'
+        );
         return;
     }
 
-    if (DD_TAGS === '<TAG_KEY>:<TAG_VALUE>') {
-        context.log.warn('You must configure your tags with a comma separated list of tags or an empty string');
+    if (client.DD_TAGS === '<TAG_KEY>:<TAG_VALUE>') {
```
ajacquemot: Why do we use as default values `<TAG_KEY>:<TAG_VALUE>`, does not look right to me.

azure/activity_logs_monitoring/index.js
```diff
@@ -1,154 +1,51 @@
...
+            'You must configure your tags with a comma separated list of tags or an empty string'
+        );
     }
 
-    var eventHubType = getEventHubMessagesFormat(eventHubMessages);
-
-    var socket = connectToDD(context);
-    var handleLogs = tagger => record => {
+    var socket = client.getSocket(context);
+    var handler = tagger => record => {
```
ajacquemot: could we rename `handler` into something more meaningful as `handlerLogs(handler, ...)` looks weird, maybe `sender` or `output`

azure/blobs_logs_monitoring/index.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    if (client.DD_TAGS === '<TAG_KEY>:<TAG_VALUE>') {
+        context.log.warn(
+            'You must configure your tags with a comma separated list of tags or an empty string'
+        );
+    }
+
+    var socket = client.getSocket(context);
+    var handler = tagger => record => {
```
ajacquemot: could we package this handler and import it so that we can easily hide its logic and reuse it for different azure functions ?

azure/shared/client.js
```diff
@@ -0,0 +1,151 @@
...
+    });
+
+    return socket;
+}
+
+function send(socket, record) {
+    return socket.write(DD_API_KEY + ' ' + JSON.stringify(record) + '\n');
+}
+
+function handleLogs(handler, logs, context) {
```
ajacquemot: I think you should create a new class `LogsHandler` that wraps the sending logic and tagging logic and just call `logsHandler.handler(logs, context)` from the outside world.
Would be nice maybe to inject a client for the sending logic and tagger for the tagging logic.

ajacquemot (commented):
LGTM too me but I would go further into refactoring logics, there is room to improve your client (handler) module and your different lambda function should much more simple.
I think you could also create a validator that you could call from your different functions.

azure/shared/client.js
achntrl: A priori those are Azure provided strings and shouldn't be too weird, then the array is generated by `String.split` so I don't expect null/undefined values here

azure/blobs_logs_monitoring/index.js
achntrl: 👍

azure/blobs_logs_monitoring/index.js
achntrl: Not sure, this is a closure to allow side effects on the socket object

azure/shared/client.js
achntrl: Yes, because this is one of the `tagger` function. If you call this one directly, you could reuse the 1st argument, but if call the one below `addTagsToStringLog`, you cannot because you pass a string and gets a new object. 

I'd be interested if you have a suggestion, ideally which doesn't allocate a new object if we call `addTagsToJsonLog` directly

azure/blobs_logs_monitoring/index.js
```diff
@@ -0,0 +1,45 @@
...
+    var sender = tagger => record => {
+        record = tagger(record, context);
+        if (!client.send(socket, record)) {
+            // Retry once
+            socket = client.getSocket(context);
+            client.send(socket, record);
+        }
+    };
+
+    var logs = blobContent.trim().split('\n');
```
ajacquemot, line 37: Could we stack overflow here ?

azure/blobs_logs_monitoring/index.js
```diff
@@ -0,0 +1,45 @@
...
+
+    logs.forEach(log => {
```
ajacquemot, line 39: Why do you need this for loop ? why not just just call `client.handleLogs(sender, logs, context);` ?

achntrl: In case you have several lines that aren't in the same format (we have no info on the format of the logs in the blob so this is more flexible)

azure/blobs_logs_monitoring/index.js
achntrl: Probably, not sure we can do much about it though because we receive the whole blob as an argument of the function.

azure/blobs_logs_monitoring/index.js
ajacquemot: can't we have the same issue with the other integration ?

ajacquemot (approved):

azure/blobs_logs_monitoring/index.js
achntrl: The other integration is from eventHub, so the messages have a more predictable format

PR #102, opened 2019-03-29
De-duplicate the functionname tag

mnshdw:
### What does this PR do?

The `functioname` tag is added to custom tags once _for each log_ being forwarded.
This PR fixes this behaviour by moving the dict attributes building outside of the for loop.


NilsBunge (approved):
LGTM - Tested on ELB, Lambdas and Cloudwatch logs and it seems to work as expected.

achntrl (approved):

PR #106, opened 2019-04-22
sets host as cloudwatch log group for non-lambda cloudwatch logs

adamlbailey:
### What does this PR do?

For cloudwatch logs that are not lambda created, set metadata[DD_HOST]
as the cloudwatch log group.

### Motivation

The current behavior in DataDog is to not surface a host for logs that
are not lambda generated. More specifically, where `metadata[DD_SOURCE] == "cloudwatch"`.

We use a lot of ECS Fargate tasks where the only DataDog logging option is to use the lambda handler to forward cloudwatch logs. We prefer to designate the "service" the logs are being generated from before passing the logs through a parser.

This is where things get opinionated:

In all of our ECS Fargate tasks, we create a log group that is named for the service. All logs for a given service, regardless of environment, are then stored under this namespace. Conveniently this attribute, `logGroup`, is already being stored in `aws_attributes`.

### Additional Notes

Noticed the requirement for tests in the contribution guidelines, but don't see where to implement. Please let me know and I'll be happy to add if need be.

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -316,6 +316,10 @@ def awslogs_handler(event, context, metadata):
         }
     }
 
+    # Set host as log group where cloudwatch is source
+    if metadata[DD_SOURCE] == "cloudwatch":
+        metadata[DD_HOST] = aws_attributes["aws"]["awslogs"]["logGroup"]
+
     # For Lambda logs we want to extract the function name,
     # then rebuild the arn of the monitored lambda using that name.
     # Start by splitting the log group to get the function name


NilsBunge:
Hello @adamlbailey,

Thanks a lot for submitting this PR and helping us improve the AWS Log collection through Lambda.
The host value is indeed very important in Datadog as it helps retrieving the right host tags associated to the logs.

The ultimate goal would be to find a way to set the same host value as the one you have on Fargate metrics. Would you have an example of log (+ Log group) coming from Fargate in Clouwatch as well as the associate hostname on related metrics? 

As making sure those two matches would guarantee the logs to share the same host tags as the metrics.

adamlbailey:
Hi @NBParis, 

Thank you for responding so quickly.

Fargate metrics are exclusively for ECS services -- many of the "services" we use are "tasks" in the ECS lexicon which are more like cron jobs that deploy, run, and then end.

In any case, here is how a log for one of these is coming through to DataDag with the suggested changes on this PR:

image

And here is that same log in Cloudwatch:

image

NilsBunge:
Thanks for sharing this information @adamlbailey .

## Feedback about the ECS use case

Just for my knowledge, may I ask why you are not using the Datadog Agent to collect metrics and logs directly from those ECS containers? ([More information on how to do so there](https://docs.datadoghq.com/integrations/amazon_ecs/?tab=python#log-collection) ).
I mainly asking because when I see the logs you have here in cloudwatch it seems that a lot of metadata is missing (such as container id, container name, task name, ....) which is automatically added on the Datadog Agent and shared across your metrics and logs collected for those containers.

We could definitely set the log group as the hostname by default but that might be a bit misleading as the expected hostname for those would be the instance id that is running the containers.

## General feedback

Now if we put the ECS use case aside, in general we try to put the arn of the service that generated the logs as host but for cloudwatch it might sometimes be tricky so we could have the log group by default.

Waiting for your feedback about the ECS use case but otherwise the code is good to go.
Once again thanks for doing this.

adamlbailey:
@NBParis, again thanks for all the detail and time here.

You are right to ask why I have not implemented the agent to collect logs for Fargate ECS tasks. I wish I could. According to a DataDog article (1) and my own failed efforts trying to mount volumes between a sidecar container running the DataDog Docker Agent and the application container (2), I think it is safe to say that ECS Fargate Task log collection is exclusively available through Cloudwatch at this point. 

(1)
> In Fargate, since you’re restricted to CloudWatch Logs, using this method is the only available option
-- [DataDog Article](https://www.datadoghq.com/blog/monitoring-ecs-with-datadog/)

(2)
The DataDog Docker Agent requires the following mount to pick up stderr/stdout from the docker socket: `-v /var/run/docker.sock:/var/run/docker.sock:ro`. However this type of mount is not supported in Fargate.
>Fargate tasks only support nonpersistent storage volumes, so the host and sourcePath fields are not supported.
--[AWS Docs](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/bind-mounts.html)

NilsBunge:
@adamlbailey Thanks for the clarification. For Fargate indeed it is not possible to use the agent as access to the Docker socket is not possible (as Fargate is multi tenant).

It would have been possible for regular ECS containers. 

So let's merge this and I'll try to see if we have a way to get the correct hostname value from the log group information (but it does not seems like its possible).

Once again, thanks for submitting this PR.

NilsBunge (approved):

PR #108, opened 2019-04-30
Added CloudWatch via Kinesis support

jmc-rival:
### What does this PR do?

This change allows the Datadog log forwarder to be run as a consumer to a Kinesis stream of log events.

### Motivation

We have our CloudWatch logs configured to ship to a Kinesis stream via a log subscription filter. This change allows the Datadog log forwarder to support being run as a consumer to this stream instead of directly from the log subscription. This model significantly reduces the amount of concurrent lambda executions one would spend on log shipping while also providing an additional layer of indirection for log processing and handling by other tools.

### Additional Notes

There are also some minor bug fixes included:
* API Gateway logs were being marked as source: cloudwatch
* Log parsing was crashing due to an index out of bounds exception on some entries which didn't follow the LogGroup format "A/lambda/B".

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -441,13 +441,14 @@ def parse_event_source(event, key):
             return source
     if "API-Gateway" in key or "ApiGateway" in key:
         return "apigateway"
-    if is_cloudtrail(str(key)):
+    if is_cloudtrail(str(key)) or ('logGroup' in event and event['logGroup'] == 'CloudTrail'):
         return "cloudtrail"
     if "awslogs" in event:
         return "cloudwatch"
     if "Records" in event and len(event["Records"]) > 0:
         if "s3" in event["Records"][0]:
             return "s3"
+
     return "aws"
 
 def parse_service_arn(source, key, bucket, context):
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -13,6 +13,7 @@ import re
 import socket
 import ssl
 import urllib
+import itertools
 from io import BytesIO, BufferedReader
 
 import boto3
@@ -212,6 +213,8 @@ def generate_logs(event, context, metadata):
             logs = cwevent_handler(event, metadata)
         elif event_type == "sns":
             logs = sns_handler(event, metadata)
+        elif event_type == "kinesis":
+            logs = kinesis_awslogs_handler(event, context, metadata)
     except Exception as e:
         # Logs through the socket the error
         err_message = "Error parsing the object. Exception: {} for event {}".format(
@@ -230,6 +233,8 @@ def parse_event_type(event):
             return "s3"
         elif "Sns" in event["Records"][0]:
             return "sns"
+        elif "eventSource" in event["Records"][0] and event["Records"][0]["eventSource"] == "aws:kinesis":
+            return "kinesis"
 
     elif "awslogs" in event:
         return "awslogs"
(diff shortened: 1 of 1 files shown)


NilsBunge:
Hello @jmc-rival ,

Thanks for submitting this PR and your contribution to improve the Datadog log forwarder.
I'll make sure to test this and give you my feedback. 

This could definitely be a solution for logs coming from Cloudwatch and sent trough a Kinesis stream. Once validated, we will make sure to update the read me to reflect this (and it's limitation).

jmc-rival:
Thanks! One note, we found that you need to run this with ~1.5GB of ram for a kinesis batch size of 1000 for it to work well. We had issues with the lambda timing out with the default 128mb and logs backing up.

NilsBunge:
Thanks a lot for the note @jmc-rival , we will definitely make sure that this is included in the documentation.

gaetan-deputier:
LGTM!

gaetan-deputier (dismissed):

aws/logs_monitoring/lambda_function.py
```diff
@@ -230,6 +233,8 @@ def parse_event_type(event):
             return "s3"
         elif "Sns" in event["Records"][0]:
             return "sns"
+        elif "eventSource" in event["Records"][0] and event["Records"][0]["eventSource"] == "aws:kinesis":
```
blemale, line 236: It could be also done in a more consistent way with the other types, ie.:
```python
elif "kinesis" in event["Records"][0]:
  return "kinesis"
```
See https://docs.aws.amazon.com/lambda/latest/dg/with-kinesis.html

WDYT @gaetan-deputier ?

aws/logs_monitoring/lambda_function.py
```diff
@@ -287,6 +292,18 @@ def s3_handler(event, context, metadata):
...
+# Handle CloudWatch logs from Kinesis
+def kinesis_awslogs_handler(event, context, metadata):
+    def reformat_record(record):
+        return {
+            "awslogs": {
+                "data": record["kinesis"]["data"]
+            }
+        }
+        
+    return itertools.chain.from_iterable(awslogs_handler(reformat_record(r), context, metadata) for r in event["Records"])
```
blemale, line 304: I'm little bit new at python but I'm not sure to understand the need of `itertools.chain.from_iterable`, which seems to chain several iterators, as we produce a single iterator with `awslogs_handler(reformat_record(r), context, metadata) for r in event["Records"]`. Am I missing something? WDYT @gaetan-deputier?

jmc-rival: In python when do you f(x) for x in A, it generates an iterator of values which are the result of each f(x) call (read more here: https://wiki.python.org/moin/Generators). Because the kinesis stream batch will contain multiple awslogs records, we need to iterate over the batch and process each one. Each awslogs_handler return produces an iterator, so we need to chain those together to get a single iterator of all of the log entries.

Doing it like this maintains the streaming nature of process. The output of this function is basically a flat iterator over all the awslog records across all of the kinesis records.

blemale: Thank you @jmc-rival for the explanation, I totally missed that the `awslogs_handler` is returning an iterator. 👌

aws/logs_monitoring/lambda_function.py
blemale: WDYT of this one @jmc-rival?

blemale (approved):
Tested on our staging environment, works fine!

LGTM!

Thank you @jmc-rival for the contribution!

pinkerton:
@blemale @NBParis What is the plan for documenting this?

Issue #110, opened 2019-05-01
Azure Blob triggered function fails blobContent.trim is not a function for WebApp Application and Server Diagnostic logs


KyleMartin901:
**Describe what happened:**
Once a WebApp application or server diagnostic log has been added to the storage account the function errors with Exception: TypeError: blobContent.trim is not a function

```bash
2019-05-01T02:44:48  Welcome, you are now connected to log-streaming service.
2019-05-01T02:45:31.401 [Information] Executing 'Functions.sitecore_cd_server_and_application_logs' (Reason='New blob detected: logs/PROD-CD-APP/2019/05/01/02/efbf18.log', Id=4f01a2c1-a616-4351-8a0b-6806fea15632)
2019-05-01T02:45:32.270 [Error] Executed 'Functions.sitecore_cd_server_and_application_logs' (Failed, Id=4f01a2c1-a616-4351-8a0b-6806fea15632)
Result: Failure
Exception: TypeError: blobContent.trim is not a function
Stack: TypeError: blobContent.trim is not a function
    at module.exports (D:\home\site\wwwroot\sitecore_cd_server_and_application_logs\index.js:44:28)
    at WorkerChannel.invocationRequest (D:\Program Files (x86)\SiteExtensions\Functions\2.0.12427\32bit\workers\node\worker-bundle.js:16077:26)
    at ClientDuplexStream.WorkerChannel.eventStream.on (D:\Program Files (x86)\SiteExtensions\Functions\2.0.12427\32bit\workers\node\worker-bundle.js:15967:30)
    at ClientDuplexStream.emit (events.js:182:13)
    at addChunk (_stream_readable.js:283:12)
    at readableAddChunk (_stream_readable.js:264:11)
    at ClientDuplexStream.Readable.push (_stream_readable.js:219:10)
    at Object.onReceiveMessage (D:\Program Files (x86)\SiteExtensions\Functions\2.0.12427\32bit\workers\node\worker-bundle.js:19692:19)
    at InterceptingListener.module.exports.InterceptingListener.recvMessageWithContext (D:\Program Files (x86)\SiteExtensions\Functions\2.0.12427\32bit\workers\node\worker-bundle.js:19007:19)
    at D:\Program Files (x86)\SiteExtensions\Functions\2.0.12427\32bit\workers\node\worker-bundle.js:19106:14
[...]

PR #112, opened 2019-05-07
Fix a trim issue in the azure function that triggers on blob storage

NilsBunge:
### What does this PR do?

Fix an issue when the blob content is not a string and therefore the trim function does trigger an exception.

### Motivation

https://github.com/DataDog/datadog-serverless-functions/issues/110

### Additional Notes

### DIFF
--- a/azure/blobs_logs_monitoring/index.js
+++ b/azure/blobs_logs_monitoring/index.js
@@ -41,7 +41,7 @@ module.exports = function(context, blobContent) {
         }
     };
 
-    var logs = blobContent.trim().split('\n');
+    var logs = blobContent.toString().trim().split('\n');
 
     logs.forEach(log => {
         handleLogs(sender, log, context);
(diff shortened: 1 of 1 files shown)


azure/blobs_logs_monitoring/index.js
```diff
@@ -41,7 +41,7 @@ module.exports = function(context, blobContent) {
         }
     };
 
-    var logs = blobContent.trim().split('\n');
+    var logs = blobContent.toString().trim().split('\n');
```
achntrl: I think what you want is because of the image below

```suggestion
    var logs;
    if (typeof blobContent === 'string') {
        logs = blobContent.trim().split('\n');
    } else {
        logs = JSON.stringify(blobContent).trim().split('\n');
    }
```

image

achntrl (changes requested):
See comment, 

I'd be curious to see the bug report that led to this change, to see the type and the content of `blobContent`

achntrl (approved):
Should we bump the function version ?

PR #117, opened 2019-05-27
Feature/aws reduce s3 handler mem usage

otherchirps:
### What does this PR do?

Reduce the memory consumption by the aws lambda function, mainly when processing gzipped logs.

### Motivation

When reading `.gz` logs, to send back to datadog, multiple in-memory copies of the decompressed data are being made.  This is ok in most cases, but when the compressed files are already large (eg. 100MB+), the lambda function tended to bomb out with Out of Memory exceptions.

Previously, the `s3_handler` would:

1. Read the entire contents of the gzipped file into memory from s3 (not too much choice here...),
2. Open a `gzip.GzipFile` iterator to walk each decompressed line of the file,
3. Immediately join all the lines ("all lines" list, unneeded copy 1) into one big `data` string (unneeded copy 2),
4. For the case of just "Send lines to Datadog", the `data` string is again split back in to a list of lines (unneeded copy 3).

This change tries to eliminate the copies created in steps 3 & 4.

In our local test case, memory usage against a 100MB gzip logfile dropped from approx 3GB (and crashing...) down to 300MB (and completing...).

For the other cases catered for in this handler (`is_cloudtrail` and has a `DD_MULTILINE_LOG_REGEX_PATTERN`), the behaviour remains the same as before.

For the `is_cloudtrail` case, it expects `data` to be a big blob of json to be parsed.
Whereas the multi-line regex check expects many lines in `data`.  So both of these have been left alone for now.

### Additional Notes

`normalize_logs` and `DatadogBatcher.batch` were consuming the log generators they were given and returning full lists (extra copies of all the log data).  Also altered these to be generator functions.

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -74,8 +74,8 @@ SCRUBBING_RULE_CONFIGS = [
         "xxxxx@xxxxx.com",
     ),
     ScrubbingRuleConfig(
-        "DD_SCRUBBING_RULE", 
-        os.getenv("DD_SCRUBBING_RULE", default=None), 
+        "DD_SCRUBBING_RULE",
+        os.getenv("DD_SCRUBBING_RULE", default=None),
         os.getenv("DD_SCRUBBING_RULE_REPLACEMENT", default="xxxxx")
     )
 ]
@@ -92,7 +92,7 @@ def compileRegex(rule, pattern):
             raise Exception("could not compile {} regex with pattern: {}".format(rule, pattern))
 
 
-# Filtering logs 
+# Filtering logs
 # Option to include or exclude logs based on a pattern match
 INCLUDE_AT_MATCH = os.getenv("INCLUDE_AT_MATCH", default=None)
 include_regex = compileRegex("INCLUDE_AT_MATCH", INCLUDE_AT_MATCH)
@@ -357,7 +357,7 @@ class DatadogScrubber(object):
             if config.name in os.environ:
                 rules.append(
                     ScrubbingRule(
-                        compileRegex(config.name, config.pattern), 
+                        compileRegex(config.name, config.pattern),
                         config.placeholder
                     )
                 )
@@ -494,7 +494,7 @@ def filter_logs(logs):
     """
     if INCLUDE_AT_MATCH is None and EXCLUDE_AT_MATCH is None:
         # convert to strings
-        return logs 
+        return logs
     # Add logs that should be sent to logs_to_send
     logs_to_send = []
     # Test each log for exclusion and inclusion, if the criteria exist
@@ -505,7 +505,7 @@ def filter_logs(logs):
                 if re.search(exclude_regex, log):
                     continue
             if INCLUDE_AT_MATCH is not None:
-                # if no include match is found, do not add log to logs_to_send 
+                # if no include match is found, do not add log to logs_to_send
                 if not re.search(include_regex, log):
                     continue
             logs_to_send.append(log)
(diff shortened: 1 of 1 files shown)


otherchirps:
👋 Hi folks, just checking if anyone has any problems or questions with these changes?  
Understand everyone is likely busy, so mightn't have had a chance to look yet.  But putting it out there that I'm happy to talk through the changes / adjust the PR if necessary.

fwiw, have been using these changes in production here at my workplace for a while now, and has seen a significant reduction in memory usage by the lambda function.  I didn't attempt to make changes to the azure routine, since we're an aws-only shop at the moment.

NilsBunge:
Hello @otherchirps ,

Thanks a lot for submitting this PR with all this context and details. 
I'll make sure to forward this to team so they can review it and make any comment if any.

otherchirps:
Thanks @NBParis. I notice that there's now a conflict with master though.  I'll try to submit an update to resolve that in the next day or two.

otherchirps:
> I notice that there's now a conflict with master though. I'll try to submit an update to resolve that in the next day or two.

Well, that was a long "day or two". My apologies.

I believe this change is ready again, and it's working for us in our environments.  Let me know if anything needs clarifying/changing.

aws/logs_monitoring/lambda_function.py
```diff
@@ -591,7 +584,23 @@ def s3_handler(event, context, metadata):
...
+            # Only make another copy of our log data if we're performing some operation
+            # against the entire block of data. Need to conserve memory in lambda land.
+            if DD_MULTILINE_LOG_REGEX_PATTERN or is_cloudtrail(str(key)):
+                data = "".join(BufferedReader(decompress_stream))
+            else:
+                # Send decompressed lines to Datadog. Stream iter is line-by-line already.
+
+                # NOTE: Duplicates the non-compressed case below, but we can't allow falling through to that
+                #   since we'll lose the decompress_stream context.
+                for line in decompress_stream:
```
ajacquemot, line 613: this block is duplicate from https://github.com/DataDog/datadog-serverless-functions/pull/117/files#diff-618731ddf1c446ef45d09613cb4b189bR622, do you think we can factorise this?

ajacquemot (approved):
Hi @otherchirps,

Thanks a lot for working on this, I left one comment but I don't think this should be a blocker.
Cheers

aws/logs_monitoring/lambda_function.py
otherchirps: Good call.  Yes, I'll push through an update shortly.

aws/logs_monitoring/lambda_function.py
```diff
@@ -572,6 +565,25 @@ def s3_handler(event, context, metadata):
...
+        "aws": {
+            "s3": {
+                "bucket": bucket, 
+                "key": key
+            }
+        }
+    }
+
+    # Build a structured message out of the s3 event data.
+    def reformat_record(record):
```
ajacquemot: nits
```suggestion
    def build_message(record):
```
or
```suggestion
    def to_message(record):
```

otherchirps: No problem.  Renamed to `build_message`.

otherchirps:
Rebased on current master to resolve merge conflict.

tianchu:
Opened a new PR with @otherchirps commits rebased against the latest master https://github.com/DataDog/datadog-serverless-functions/pull/210

tianchu:
Please try out the new approach to ingest logs from AWS to Datadog using the [Datadog Kinesis firehose destination](https://docs.datadoghq.com/logs/guide/send-aws-services-logs-with-the-datadog-kinesis-firehose-destination/).

PR #118, opened 2019-05-28
Update link for adding custom tags to lambda logs

gzysk8:
### What does this PR do?

Updates readme to point to correct lines of code in lambda function for adding custom tags.

### Motivation

Previous link was broken.


NilsBunge (approved):
Thanks for this. Definitely needed an update.

PR #120, opened 2019-06-05
Forward metrics from Lambda logs

tianchu:
### What does this PR do?

This PR enables the log forwarder to detect metrics in Lambda CloudWatch log groups and forward the metrics through a background thread using Datadog Lambda Layer.

This is the related PR (https://github.com/DataDog/datadog-lambda-layer-python/pull/4) that enables Lambda layer to write metrics to logs.

### Motivation

Forwarding monitoring data from CloudWatch logs to Datadog is the only viable and the AWS-recommended way to collect monitoring data without a performance penalty on customers' Lambda functions.

Need to add this functionality to the existing log forwarder, instead of creating a separate Lambda function, because each CloudWatch log group only support one Lambda subscription and customers often need to forward both logs and metrics.

### Additional Notes

This change is backward-compatible, and existing users can safely upgrade by copy & paste.

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -21,19 +21,18 @@ from io import BytesIO, BufferedReader
 import boto3
 
 try:
+    # Datadog Lambda layer is required to forward metrics
     from datadog_lambda.wrapper import datadog_lambda_wrapper
-    from datadog_lambda.metric import lambda_metric
+    from datadog_lambda.metric import lambda_stats
     DD_FORWARD_METRIC = True
 except ImportError:
-    # Datadog Lambda layer is required to forward metrics
+    # For backward-compatibility
     DD_FORWARD_METRIC = False
 
 
 # Set this variable to `False` to disable log forwarding.
-# E.g., you only want to forward metrics from logs, but not logs.
+# E.g., when you only want to forward metrics from logs.
 DD_FORWARD_LOG = os.getenv("DD_FORWARD_LOG", default="true").lower() == "true"
-if not DD_FORWARD_LOG:
-    print("Log forwarding is disabled")
 
 
 # Change this value to change the underlying network client (HTTP or TCP),
@@ -89,6 +88,26 @@ elif "DD_API_KEY" in os.environ:
 
 # Strip any trailing and leading whitespace from the API key
 DD_API_KEY = DD_API_KEY.strip()
+
+# DD_API_KEY must be set
+if DD_API_KEY == "<your_api_key>" or DD_API_KEY == "":
+    raise Exception(
+        "You must configure your Datadog API key using "
+        "DD_KMS_API_KEY or DD_API_KEY"
+    )
+# Check if the API key is the correct number of characters
+if len(DD_API_KEY) != 32:
+    raise Exception(
+        "The API key is not the expected length. "
+        "Please confirm that your API key is correct"
+    )
+# Validate the API key
+validation_res = requests.get(
+    "https://api.{}/api/v1/validate?api_key={}".format(DD_SITE, DD_API_KEY)
+)
+if not validation_res.ok:
+    raise Exception("The API key is not valid.")
+
 
 # DD_MULTILINE_REGEX: Datadog Multiline Log Regular Expression Pattern
 DD_MULTILINE_LOG_REGEX_PATTERN = None
@@ -327,12 +346,12 @@ class DatadogScrubber(object):
 
 def datadog_forwarder(event, context):
     """The actual lambda function entry point"""
-    check_api_key()
     logs = generate_logs(event, context)
+    metrics, logs = extract_metrics(logs)
     if DD_FORWARD_LOG:
         forward_logs(logs)
     if DD_FORWARD_METRIC:
-        forward_metrics(logs)
+        forward_metrics(metrics)
 
 
 if DD_FORWARD_METRIC:
@@ -340,22 +359,6 @@ if DD_FORWARD_METRIC:
     lambda_handler = datadog_lambda_wrapper(datadog_forwarder)
 else:
     lambda_handler = datadog_forwarder
-
-
-def check_api_key():
-    """Check Datadog API Key"""
-    # Check prerequisites
-    if DD_API_KEY == "<your_api_key>" or DD_API_KEY == "":
-        raise Exception(
-            "You must configure your Datadog API key using "
-            "DD_KMS_API_KEY or DD_API_KEY"
-        )
-    # Check if the API key is the correct number of characters
-    if len(DD_API_KEY) != 32:
-        raise Exception(
-            "The API key is not the expected length. "
-            "Please confirm that your API key is correct"
-        )
 
 
 def forward_logs(logs):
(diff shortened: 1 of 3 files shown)


aws/logs_monitoring/lambda_function.py
```diff
@@ -390,6 +425,29 @@ def generate_metadata(context):
...
+        return
+
+    for log in logs:
+        try:
+            if log.get(DD_SOURCE) != 'lambda':
+                continue
+            metric = json.loads(log['message'])
+            required_attrs = ('metric_name', 'value', 'timestamp', 'tags')
+            if all(attr in metric for attr in required_attrs):
+                lambda_metric(**metric)
```
tianchu: `lambda_metric` queues the metrics and flush them in another python thread periodically.

aws/logs_monitoring/log-sam-template.yaml
```diff
@@ -1,13 +1,15 @@
...
+Description: Pushes logs and metrics from AWS to Datadog.
 Resources:
   loglambdaddfunction:
     Type: 'AWS::Serverless::Function'
     Properties:
-      Description: Pushes logs from S3 and CloudWatch Logs to Datadog.
+      Description: Pushes logs and metrics from AWS to Datadog.
       Handler: lambda_function.lambda_handler
-      MemorySize: 128
+      MemorySize: 1024
```
tianchu, line 10: Updated the parameters to match the recommended values in the public datadog document.

aws/logs_monitoring/lambda_function.py
```diff
@@ -19,6 +19,19 @@
 
 import boto3
 
+try:
```
tianchu, line 23: Existing users may not have lambda layer added to their existing log forwarder installations, the new installations of log forwarder are expected to have the layer added.

aws/logs_monitoring/lambda_function.py
```diff
@@ -89,7 +102,7 @@ def __init__(self, name, pattern, placeholder):
 DD_CUSTOM_TAGS = "ddtags"
 DD_SERVICE = "service"
 DD_HOST = "host"
-DD_FORWARDER_VERSION = "1.3.1"
+DD_FORWARDER_VERSION = "1.3.2"
```
ajacquemot: I think this is a minor change

aws/logs_monitoring/lambda_function.py
ajacquemot: can we use a more explicit name or use your message as comment please ? What about `lambda_metric.push(**metric)` or `lambda_metric.add(**metric)` or `forward_to_lambda_metric(**metric)`

aws/logs_monitoring/lambda_function.py
```diff
@@ -309,19 +322,41 @@ def scrub(self, payload):
...
-            "The API key is not the expected length. Please confirm that your API key is correct"
+            "The API key is not the expected length. "
+            "Please confirm that your API key is correct"
         )
 
-    logs = generate_logs(event, context)
+
+def forward_logs(logs):
+    """Forward logs to Datadog"""
+    if not DD_FORWARD_LOG:
```
ajacquemot: we could probably move this check to the caller ?

aws/logs_monitoring/lambda_function.py
```diff
@@ -390,6 +425,29 @@ def generate_metadata(context):
     return metadata
 
 
+def forward_lambda_metrics(logs):
+    """
+    Process custom metrics submitted via lambda logs, and send them to Datadog
+    in a background thread using `lambda_metric` that is provided by the
+    Datadog Python Lambda Layer.
+    """
+    if not DD_FORWARD_METRIC:
```
ajacquemot: same ?

aws/logs_monitoring/lambda_function.py
```diff
@@ -309,19 +322,41 @@ def scrub(self, payload):
         return payload
 
 
-def lambda_handler(event, context):
+def lambda_handler_wrapped(event, context):
```
ajacquemot: Can we use a more explicit name here please ? I don't think `wrapped` is very explicit

ajacquemot (commented):
Few nits, overall LGTM, it's worth having @NBParis in the loop.

aws/logs_monitoring/lambda_function.py
tianchu: Yes agreed. I will add a comment. `lambda_metric` is provided by lambda layer which is used by customers too.

aws/logs_monitoring/lambda_function.py
tianchu: Renamed it to `datadog_log_forwarder`.

aws/logs_monitoring/lambda_function.py
```diff
@@ -322,18 +324,21 @@ def scrub(self, payload):
         return payload
 
 
-def lambda_handler_wrapped(event, context):
+def datadog_log_forwarder(event, context):
```
ajacquemot: nits: let's remove log as it can forward metrics as well

aws/logs_monitoring/lambda_function.py
```diff
@@ -322,18 +324,21 @@ def scrub(self, payload):
...
+    """The actual lambda function entry point"""
     check_api_key()
     logs = generate_logs(event, context)
-    forward_logs(logs)
-    forward_lambda_metrics(logs)
+    if DD_FORWARD_LOG:
+        forward_logs(logs)
+    if DD_FORWARD_METRIC:
+        forward_lambda_metrics(logs)
```
ajacquemot: nits: forward_metrics

aws/logs_monitoring/lambda_function.py
```diff
@@ -19,6 +19,21 @@
...
+except ImportError:
+    # Datadog Lambda layer is required to forward metrics
+    DD_FORWARD_METRIC = False
+
+
+# Set this variable to `False` to disable log forwarding.
+# E.g., you only want to forward metrics from logs, but not logs.
+DD_FORWARD_LOG = os.getenv("DD_FORWARD_LOG", default="true").lower() == "true"
+if not DD_FORWARD_LOG:
+    print("Log forwarding is disabled")
```
tmichelet: Do we want to print this for every single lambda function run?

tianchu: I was printing it for every single invocation, but I realized that it would be way too verbose for metric-only customers (the primary use case). It would also contribute 33% (function start, this line, and function stop) of the cost of CW logs for the forwarder lambda itself. I also felt that it's quite easy to notice this environment variable in AWS console, if a customer is wondering why logs are not being forwarded.

Do you think it should be logged for every run?

tmichelet: I think we should completely remove this line

aws/logs_monitoring/lambda_function.py
```diff
@@ -310,20 +344,25 @@ def scrub(self, payload):
...
-    # Check if the API key is the correct number of characters
-    if len(DD_API_KEY) != 32:
-        raise Exception(
-            "The API key is not the expected length. Please confirm that your API key is correct"
-        )
-
+def datadog_forwarder(event, context):
+    """The actual lambda function entry point"""
     logs = generate_logs(event, context)
+    metrics, logs = extract_metrics(logs)
```
ajacquemot: nits, this method does not really extract logs, it splits logs from metrics

aws/logs_monitoring/lambda_function.py
```diff
@@ -310,20 +344,25 @@ def scrub(self, payload):
...
-        )
-    # Check if the API key is the correct number of characters
-    if len(DD_API_KEY) != 32:
-        raise Exception(
-            "The API key is not the expected length. Please confirm that your API key is correct"
-        )
-
+def datadog_forwarder(event, context):
+    """The actual lambda function entry point"""
     logs = generate_logs(event, context)
```
ajacquemot: nits: this does not only generate logs, it can generate metrics as well.

ajacquemot (commented):
Left two nits but looks good overall

aws/logs_monitoring/lambda_function.py
```diff
@@ -393,16 +433,58 @@ def generate_metadata(context):
     return metadata
 
 
+def extract_metric(event):
+    """Extract metric from an event if possible"""
+    try:
+        metric = json.loads(event['message'])
+        required_attrs = {'m', 'v', 'e', 't'}
```
achntrl, line 440: Where can I see the code related to that ?

achntrl (approved):
LGTM on the condition that metric format is properly identified. Is the format documented somewhere ?

aws/logs_monitoring/lambda_function.py
achntrl: Nevermind found it: https://github.com/DataDog/datadog-lambda-layer-python/pull/5/files
I didn't see it because it isn't merged yet 👍

tianchu:
Got the green light from @NBParis in slack.

PR #124, opened 2019-06-24
raises an exception if multiline regex is invalid

le-he-hoo:
### What does this PR do?

Raises an exception if the regex passed in with the DD_MULTILINE_LOG_REGEX_PATTERN env var is invalid

### Motivation

Add helpful feedback for users trying to configure multiline aggregation that helps identify what the problem is more than the previous message `module initialization error: unexpected end of pattern`

### Additional Notes


ajacquemot (approved):
Thanks for the PR 🚀

PR #126, opened 2019-06-28
Adds include and exclude at match to Lambda log forwarder

le-he-hoo:
### What does this PR do?

Adds `INCLUDE_AT_MATCH` and `EXCLUDE_AT_MATCH` to Lambda log forwarder
### Motivation

Customer  request

### Additional Notes

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -272,7 +272,7 @@ class DatadogHTTPClient(object):
         try:
             resp = self._session.post(
                 self._url,
-                data=self._scrubber.scrub("[{}]".format(", ".join(map(json.dumps, logs)))),
+                data=self._scrubber.scrub("[{}]".format(",".join(logs))),
                 timeout=self._timeout,
             )
         except ScrubbingException:
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -74,9 +74,9 @@ SCRUBBING_RULE_CONFIGS = [
         "xxxxx@xxxxx.com",
     ),
     ScrubbingRuleConfig(
-        "CUSTOM_SCRUBBING_RULE", 
-        os.getenv("CUSTOM_SCRUBBING_RULE", default=None), 
-        os.getenv("CUSTOM_SCRUBBING_RULE_REPLACEMENT", default="xxxxx")
+        "DD_SCRUBBING_RULE", 
+        os.getenv("DD_SCRUBBING_RULE", default=None), 
+        os.getenv("DD_SCRUBBING_RULE_REPLACEMENT", default="xxxxx")
     )
 ]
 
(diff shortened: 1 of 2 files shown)


le-he-hoo:
TODO: update README.md

aws/logs_monitoring/lambda_function.py
```diff
@@ -465,6 +486,48 @@ def split(events):
     return metrics, logs
 
 
+# should only be called when INCLUDE_AT_MATCH and/or EXCLUDE_AT_MATCH exist
+def filter_logs(logs):
```
ajacquemot, line 488: I think you could simplify this method by iterating over all logs and for each log, try include, try exclude, this would limit the cases to two.

ajacquemot: The question now is which rule wins where the log should both included and excluded, on the agent we decided to exclude it https://github.com/DataDog/datadog-agent/blob/master/pkg/logs/processor/processor.go#L78

aws/logs_monitoring/lambda_function.py
```diff
@@ -80,6 +80,27 @@ def __init__(self, name, pattern, placeholder):
     )
 ]
 
+
+# Filtering logs 
+# Option to include or exclude logs based on a pattern match
+INCLUDE_AT_MATCH = os.getenv("INCLUDE_AT_MATCH", default=None)
+if INCLUDE_AT_MATCH is not None:
```
ajacquemot: We can see code duplication here, the exact same logic is also used inside the scrubber, I think we should isolate this logic inside a method, and use from different places.
The interface would look like:
```
def compileRegex(pattern):
  # TODO: implement me
  pass
```

ajacquemot (commented):
Thanks for working on this, the logic is here, let's factorise and simplify a bit what we have 👍

aws/logs_monitoring/lambda_function.py
le-he-hoo: Exclude 🛑 wins here too 👍

Simplified it a bit.

aws/logs_monitoring/lambda_function.py
```diff
@@ -465,6 +486,34 @@ def split(events):
...
+    If no filtering rules exist, return all the logs.
+    """
+    if INCLUDE_AT_MATCH is None and EXCLUDE_AT_MATCH is None:
+        return logs 
+    # Add logs that should be sent to logs_to_send
+    logs_to_send = []
+    # Test each log for exclusion and inclusion, if the criteria exist
+    for log in logs:
+        try:
+            string_log = json.dumps(log)
```
ajacquemot: Here we are transforming logs to json-string and storing the logs at the end, could we store a json-string instead and change the logic downstream to manipulation string instead of objects so that we don't do the object-to-string operation twice ?

aws/logs_monitoring/lambda_function.py
```diff
@@ -74,12 +74,33 @@ def __init__(self, name, pattern, placeholder):
         "xxxxx@xxxxx.com",
     ),
     ScrubbingRuleConfig(
-        "CUSTOM_SCRUBBING_RULE", 
-        os.getenv("CUSTOM_SCRUBBING_RULE", default=None), 
-        os.getenv("CUSTOM_SCRUBBING_RULE_REPLACEMENT", default="xxxxx")
+        "DD_SCRUBBING_RULE",
```
ajacquemot, line 77: The change on scrubbing rules could have been done in another PR to ease review

aws/logs_monitoring/lambda_function.py
ajacquemot: I can help with this one if you want, feel free to ping me IRC.

aws/logs_monitoring/lambda_function.py
le-he-hoo: Good point!

aws/logs_monitoring/lambda_function.py
le-he-hoo: I went for it and made a few downstream changes.

aws/logs_monitoring/lambda_function.py
le-he-hoo: I added a `compileRegex` and used it for the include, exclude, and scrubbing rules.

aws/logs_monitoring/lambda_function.py
```diff
@@ -465,6 +484,35 @@ def split(events):
     return metrics, logs
 
 
+# should only be called when INCLUDE_AT_MATCH and/or EXCLUDE_AT_MATCH exist
+def filter_logs(logs):
```
ajacquemot, line 488: would be nice that this method does not have to transform the logs to json, can we do it upstream please ?

aws/logs_monitoring/lambda_function.py
```diff
@@ -251,7 +272,7 @@ def send(self, logs):
         try:
             resp = self._session.post(
                 self._url,
-                data=self._scrubber.scrub(json.dumps(logs)),
+                data=self._scrubber.scrub("[{}]".format(", ".join(map(json.dumps, logs)))),
```
ajacquemot: Why do we need to `json.dumps` again ? We already do it upstream right ? I think we just need to join the jsonString with `,` and wrap this under `[]`

le-he-hoo: You're right.  Good catch.

aws/logs_monitoring/lambda_function.py
le-he-hoo: @ajacquemot Sure! Where do you think is a good place upstream to do it?

le-he-hoo: Moved it to the `split` method

ajacquemot (approved):
LGTM, I think we should test thoroughly this branch before merging as we touched the frame for TCP and the payload for HTTP as well as the regexes. Once you validated that nothing broke, let's merge it 👍

PR #129, opened 2019-07-07
Adds CloudWatch multiline log support

ericmustin:
### What does this PR do?

Adds `DD_MULTILINE_CLOUDWATCH_LOG_REGEX_PATTERNS` environment variable to Lambda log forwarder to allow multiline log support for cloudwatch logs.

### Motivation

We currently support handling multiline logs for s3, this extends that support to cloudwatch logs. Multiple customer requests.

### Additional Notes

As cloudwatch log groups can each contain their own unique log formatting, this PR requires a multiline pattern for each log group. the environment variable is intended to be JSON, and parsing JSON requires common regex characters like `\` to be themselves be escaped, so double escaping is required (`\d{2}` becomes `\\d{2}` ). Tried to document this in the readme as it's unintuitive, but any guidance on how to appropriately document and demonstrate this is much appreciated, or if there's a better approach to environment variable configuration here I would be open to it.

Additionally the working assumption here for what constitutes a group of logs that should be concatenated into one multiline log is that:

1. the logs are all part of the same log stream and log group
2. the log group contains a log which matches the appropriate regex start pattern
3. the logs following the log that matches the regex do not themselves match the regex start pattern, are in consecutive order from the log that matched the regex start pattern, and contain the same timestamp as the log that matched the regex start pattern.

Lastly, I'd like to include some basic unit tests here eventually, but prefer to get some eyes on this early and get feedback. Could make that a separate PR if there's interest.


aws/logs_monitoring/README.md
```diff
@@ -115,7 +115,16 @@ If there are multiline logs in s3, set `DD_MULTILINE_LOG_REGEX_PATTERN` environm
 
 - Example: for multiline logs beginning with pattern `11/10/2014`: `DD_MULTILINE_LOG_REGEX_PATTERN="\d{2}\/\d{2}\/\d{4}"`
 
-## 8. (optional) Forward Metrics from Logs
+## 8. (optional) Multiline Log support for CloudWatch
+
+If there are multiline logs in CloudWatch, set `DD_MULTILINE_CLOUDWATCH_LOG_REGEX_PATTERNS` environment variable to the specified regex pattern to detect for a new log line per log group. This environment variable must be set to JSON, with each log group it's own key, with each value the regex start pattern to detect for a new log line in that log group.
```
ajacquemot, line 132: I think `re.escape` could help here, the user should not have to worry about escaping the characters, agreed.

aws/logs_monitoring/lambda_function.py
```diff
@@ -122,11 +122,34 @@ def __init__(self, name, pattern, placeholder):
...
     except Exception:
-        raise Exception("could not compile multiline regex with pattern: {}".format(DD_MULTILINE_LOG_REGEX_PATTERN))
+        raise Exception("could not compile s3 multiline regex with pattern: {}".format(DD_MULTILINE_LOG_REGEX_PATTERN))
     multiline_regex_start_pattern = re.compile(
         "^{}".format(DD_MULTILINE_LOG_REGEX_PATTERN)
     )
 
+
+# DD_MULTILINE_CLOUDWATCH_REGEX: Datadog Multiline Cloudwatch Logs Regular Expression Patterns for Log Groups
+DD_MULTILINE_CLOUDWATCH_LOG_REGEX_PATTERNS = None
```
ajacquemot, line 156: Do you think we might expect other AWS products to require multiline regexes ? If so, do we want to create a new environment variable for each product ?

aws/logs_monitoring/lambda_function.py
```diff
@@ -122,11 +122,34 @@ def __init__(self, name, pattern, placeholder):
...
+if "DD_MULTILINE_CLOUDWATCH_LOG_REGEX_PATTERNS" in os.environ:
+    multiline_regexs = os.environ["DD_MULTILINE_CLOUDWATCH_LOG_REGEX_PATTERNS"]
```
ajacquemot: nits: 
```suggestion
    multiline_regexes = os.environ["DD_MULTILINE_CLOUDWATCH_LOG_REGEX_PATTERNS"]
```

aws/logs_monitoring/lambda_function.py
```diff
@@ -122,11 +122,34 @@ def __init__(self, name, pattern, placeholder):
...
+    try:
+        DD_MULTILINE_CLOUDWATCH_LOG_REGEX_PATTERNS = {}
+        cloudwatch_regex_dict = json.loads(multiline_regexs)
+
+        # compile regex start paterns for each dictionary entry
+        for log_group_name in cloudwatch_regex_dict:
```
ajacquemot: do we want to use `items()` do have both the `group_name` and the `rule` ?

aws/logs_monitoring/lambda_function.py
```diff
@@ -122,11 +122,34 @@ def __init__(self, name, pattern, placeholder):
...
+            if log_group_name not in DD_MULTILINE_CLOUDWATCH_LOG_REGEX_PATTERNS:
+                log_group_rule = cloudwatch_regex_dict[log_group_name]
+
+                # ensure pattern only matchs at start of line
+                DD_MULTILINE_CLOUDWATCH_LOG_REGEX_PATTERNS[log_group_name] = re.compile(
+                    "^{}".format(log_group_rule)
```
ajacquemot: this code is duplicated from https://github.com/DataDog/datadog-serverless-functions/pull/129/files#diff-618731ddf1c446ef45d09613cb4b189bR127, do we want to factorise ?

aws/logs_monitoring/lambda_function.py
```diff
@@ -652,7 +675,35 @@ def awslogs_handler(event, context, metadata):
                 metadata[DD_HOST] = arn
 
     # Create and send structured logs to Datadog
-    for log in logs["logEvents"]:
+    # if multiline regex, iterate over log events and concatenate multiline logs into parent log
+    # child multiline logs must have matching timestamp and not match regex start pattern
```
ajacquemot, line 733: `must have matching` is it a hard guarantee on AWS side ? If yes, can you link the information in the comment please. If not I think it's a strong assumption we are making and I am worried that there could be edge cases where multiline logs have 1 digit different and more.

aws/logs_monitoring/lambda_function.py
```diff
@@ -652,7 +675,35 @@ def awslogs_handler(event, context, metadata):
...
+                multiline_log_timestamp = log["timestamp"]
+                multiline_log_flag = True
+                log_events.append(log)
+            elif log["timestamp"] != multiline_log_timestamp:
+                multiline_log_timestamp = None
+                multiline_log_flag = False
+                log_events.append(log)
+            elif multiline_log_flag:
+                # valid child multiline log, matching timestamp and multiline log flag has been toggled
+                log_events[-1]["message"] += log["message"]
```
ajacquemot: could we use an intermediate buffer instead of appending the content directly to the last log please ?

aws/logs_monitoring/lambda_function.py
```diff
@@ -652,7 +675,35 @@ def awslogs_handler(event, context, metadata):
...
+            else:
+                # edge case, multiple logs without starting pattern regex
```
ajacquemot: I feel like we can't fall into this case, can you describe me a use case please ?

ajacquemot (commented):
Good work. left a few questions, thanks for implementing this feature.

ericmustin:
@ajacquemot thanks for giving this a look, will try to answer all the followups as soon as i have some free cycles here

aws/logs_monitoring/README.md
ericmustin: had some trouble here but was able to get this working without having to force the user to escape char's by using  `ast.literal_eval` as a workaround to `json.loads()` . Seems to work for me but lmk if this approach is alright

aws/logs_monitoring/lambda_function.py
ericmustin: I believe the other options here that we differentiate on would be `aws sns` or `aws kenesis`  logs, while i'm not familiar with them I assume they could also contain multiline logs. My thinking was that if we implement handling multiline logging for either of those we would add their own environment variable at that time.

aws/logs_monitoring/lambda_function.py
ericmustin: 👍 updated

aws/logs_monitoring/lambda_function.py
ericmustin: 👍 makes sense, abstracted this to it's own function [`create_multiline_start_regex`](https://github.com/DataDog/datadog-serverless-functions/blob/c1f2d4dceae006868c6b9697fe7b5919b0da83c6/aws/logs_monitoring/lambda_function.py#L117)

aws/logs_monitoring/lambda_function.py
ericmustin: 👍  agree, I updated this to use an empty dictionary `{}` as intermediate storage here of the multiline log, which let me clean up `if/elif/else` code here.  Apologies as I wasn't completely sure what you meant by buffer so if there's a better or more pythonic way of doing this please lmk, can update

aws/logs_monitoring/lambda_function.py
ericmustin: So i've been testing this with the cloudwatch logs from a couple very basic ruby/js lambdas, which can be found here: [`example lambdas and related logs`](https://gist.github.com/ericmustin/adb15cef1576baf5df7697db1f5ee7b7)

The use case here would be if multiple logs in a row did not contain the multiline start pattern. An example is aws lambdas automatically insert two logs to denote the `END` and `REPORT` of the lambda, as seen here: [`example of consecutive logs without start pattern`](https://gist.github.com/ericmustin/adb15cef1576baf5df7697db1f5ee7b7#file-example_output_node-log-L13-L14) , but there are probably a number of ways to hit this case

aws/logs_monitoring/lambda_function.py
ericmustin: 👍 ty, updated to use `iteritems` (for 2.7)

aws/logs_monitoring/lambda_function.py
ericmustin: that's a good question and feedback. 

This is just what I saw in sample logs I generated with a lambda. Spent some time investigating but haven't been able to confirm. I don't think this is documented as a hard guarantee anywhere. 

The matching timestamps requirement was meant to be a conservative approach so that any unrelated logs without the start pattern would not accidentally get appended to the previous log. My thinking here is that the new behavior, in the case of a multiline log with different timestamps on "child" lines, would be that the multiline log would not include the full log. These log lines of the multiline log wouldn't get lost, but just sent as their own individual lines, which is the same as the current behavior.

It's hard for me to be super confident here about these sort of edge cases without more thorough logs to test against from different aws services. do you have any advice here? I can try to spend some time collecting different logs from the different aws services if we feel like it will add confidence to this PR.

aws/logs_monitoring/README.md
```diff
@@ -126,6 +126,16 @@ If there are multiline logs in s3, set `DD_MULTILINE_LOG_REGEX_PATTERN` environm
 
 - Example: for multiline logs beginning with pattern `11/10/2014`: `DD_MULTILINE_LOG_REGEX_PATTERN="\d{2}\/\d{2}\/\d{4}"`
 
+<<<<<<< HEAD
+## 8. (optional) Multiline Log support for CloudWatch
+
+If there are multiline logs in CloudWatch, set `DD_MULTILINE_CLOUDWATCH_LOG_REGEX_PATTERNS` environment variable to the specified regex pattern to detect for a new log line per log group. This environment variable must be set to JSON, with each log group it's own key, with each value the regex start pattern to detect for a new log line in that log group.
+
+- Example: for multiline logs from CloudWatch log group `aws/lambda/exampleOne` beginning with pattern `11/10/2014` and log group `aws/lambda/exampleTwo` beginning with pattern: `2014-11-10`: `DD_MULTILINE_CLOUDWATCH_LOG_REGEX_PATTERNS=
+{"/aws/lambda/exampleOne":"\d{2}/\d{2}/\d{4}", "/aws/lambda/exampleTwo":"\d{4}-\d{2}-\d{2}"}`
```
achntrl, line 135: Is this a valid env variable or do you need to escape the `"` ?

achntrl (commented):
Seems ok, I'd like to see some unit tests to see it in action though

ericmustin:
@achntrl thanks for taking a look, makes sense re:tests and i agree, will get some unit tests added for this

ericmustin:
still working on this, should have some unit tests to add shortly, has required a bit of cleanup to get this to run within a testing framework

ericmustin:
closing this as it's fallen quite a bit out of step with master at this point, seemed to work just fine but didn't have the bandwidth to gain confidence testing against the various aws sources formatting besides the ones mentioned previously, or add relevant unit tests / test framework. might be something @NoisomePossum or @le-he-hoo can pick up in the future?

PR #132, opened 2019-07-15
checks if blobContent isBuffer and if so, converts it to string

le-he-hoo:
### What does this PR do?

Checks if blobContent isBuffer and converts to string if so

### Motivation

Support Issue.  Logs were coming in as blocks of ascii character codes instead of text because `JSON.stringify(blobContent)` does not properly convert to string if blobContent `isBuffer`. 

### Additional Notes


ajacquemot (approved):
LGTM thanks for fixing this issue, I let you test it.

PR #134, opened 2019-07-29
Parse ddtags from message json and add to metadata, if possible

jeffgran:
### What does this PR do?

This PR allows you to put datadog tags inside your log message. It requires that your message is valid JSON, and it looks for tags under the `"ddtags"` key. The tags should be in a single string, separated by commas, for example:

```
{
   http_response_code: 200,
   my_other_data: {...},
   ddtags: "zach:morris,ac:slater",
}
```

If the message is not JSON or if the `"ddtags"` key is not present, the function continues to work as it did before this PR.

### Motivation

When running an application in AWS Fargate, it seems currently the only option for logging to datadog is to log to STDOUT, set it up to be picked up by Cloudwatch, and then use this lambda to push the logs to datadog. However, this lambda only allows hard-coded tags. I needed a way to pass dynamically generated tags from my application, and the only place I can do that is inside the log message itself. 

### Additional Notes

It has been a long time since I wrote python! Please let me know if I need to fix anything. But this is working for me in production.

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -460,6 +460,16 @@ def generate_metadata(context):
 
     return metadata
 
+def parse_tags_from_message(event):
+    """extract tags from message json if possible"""
+    try:
+        data = json.loads(event['message'])
+        if 'ddtags' in data:
+            return data['ddtags']
+        else:
+            return None
+    except Exception:
+        return None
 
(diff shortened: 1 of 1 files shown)


tianchu:
@jeffgran I know this is a very old thread, but the feature you suggested has been added in https://github.com/DataDog/datadog-serverless-functions/pull/326!

PR #135, opened 2019-07-30
changes multi-line aggregation regex to fix bug

le-he-hoo:
### What does this PR do?

Changes the multi-line aggregation rule regular expression to fix an aggregation bug where logs could be split incorrectly.  The fix forces the split to happen only when the matched pattern occurs after a newline. 

### Motivation

_Bug Fix_: 
Noticed a situation when the multi-line aggregation rule could fail

### Additional Notes

_Bug Description_:
If you have multi-line regex pattern `my_pattern`, and a log like the following:

```
my_pattern [INFO] [98632] my_pattern random text
1234 - 200 /api/healthcheck
```

Then, the following two logs is created

```
my_pattern [INFO] [98632] 
```

```
my_pattern random text
1234 - 200 /api/healthcheck
```

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -140,7 +140,7 @@ if "DD_MULTILINE_LOG_REGEX_PATTERN" in os.environ:
     DD_MULTILINE_LOG_REGEX_PATTERN = os.environ["DD_MULTILINE_LOG_REGEX_PATTERN"]
     try:
         multiline_regex = re.compile(
-            "\n+(?={})".format(DD_MULTILINE_LOG_REGEX_PATTERN)
+            "[\n\r\f]+(?={})".format(DD_MULTILINE_LOG_REGEX_PATTERN)
         )
     except Exception:
         raise Exception("could not compile multiline regex with pattern: {}".format(DD_MULTILINE_LOG_REGEX_PATTERN))
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -140,7 +140,7 @@ if "DD_MULTILINE_LOG_REGEX_PATTERN" in os.environ:
     DD_MULTILINE_LOG_REGEX_PATTERN = os.environ["DD_MULTILINE_LOG_REGEX_PATTERN"]
     try:
         multiline_regex = re.compile(
-            "(?<!^)\s+(?={})(?!.\s)".format(DD_MULTILINE_LOG_REGEX_PATTERN)
+            "\n+(?={})".format(DD_MULTILINE_LOG_REGEX_PATTERN)
         )
     except Exception:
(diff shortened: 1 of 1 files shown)


aws/logs_monitoring/lambda_function.py
```diff
@@ -140,7 +140,7 @@ def compileRegex(rule, pattern):
     DD_MULTILINE_LOG_REGEX_PATTERN = os.environ["DD_MULTILINE_LOG_REGEX_PATTERN"]
     try:
         multiline_regex = re.compile(
-            "(?<!^)\s+(?={})(?!.\s)".format(DD_MULTILINE_LOG_REGEX_PATTERN)
+            "\n+(?={})".format(DD_MULTILINE_LOG_REGEX_PATTERN)
```
ericmustin: Thanks for looking into this.  Generally i think this is good to make this a bit simpler and more strict but have some questions.

I believe the reasoning here for `\s+` is that we can't assume `\n` will always be the new line separator, and `\s+` encompasses `\n` , `\r`, `\r\n`, tabs, etc. This does introduce that edge case you're describing where someone is using a non-unique multi-line start pattern throughout their logs. Is it documented anywhere that `\n` will always be the newline separator in s3 logs, if so can you point to that doc, as it would give me more confidence here about this change? My concern is that we solve one case and introduce others with this.

for the `(?<!^)` group since we're passing this regex into `.split`, this was meant to ensure we don't split before the first multiline match (creating an empty `''` log in the array that `.split()` returns), only the second, third, etc. I think this is now redundant by adding [`multiline_regex_start_pattern.match(data)`](https://github.com/DataDog/datadog-serverless-functions/blob/66354f2c29826de3cc8997ef79618812e720c4dd/aws/logs_monitoring/lambda_function.py#L607) later on in this file, probably good to remove from the regex

for the `(?!.\s)` group the intention here is to ensure that if the line starts with something like `my_patternAndSomeOtherStuffWithoutASpace` that we don't match on that as the start of a new log.

le-he-hoo: @ericmustin Thanks! I can update from `\n+` to `[\r\n]+` and that way the log always comes after either a newline `\n` or a carriage return `\r`, but not after spaces and tabs.

Making the split occur only when there is [\r\n]+ will ensure we don't split before the first multiline match, and I think we can remove the "not followed by a single non-newline character and a space" part `(?!.\s)`.

ericmustin: alright, i think that makes sense. one small nit would be to also include the `\f` form feed character, as it appears that is a valid line break in some formats as well? 

lgtm assuming changes mentioned in this thread.

fwiw, not tied to this PR, but l i think it would make sense to work on unit tests for these lambdas as we change behavior in the future. it's quite hard to have confidence in these code changes, especially given the scope of possible inputs to this lambda from s3, without a spec to test against.

ericmustin (approved):
👍

PR #144, opened 2019-08-20
Read tags from log groups

pparkkinen:
### What does this PR do?

Add functionality in log forwarding lambda to read additional custom tags from log group when forwarding logs from CloudWatch Logs events.

### Motivation

To be able to use the same log forwarding lambda to forward logs with different custom tags.

### Additional Notes

The PR also adds a couple unit tests for the log forwarding lambda along with a way to import the file without a Datadog API key.


NilsBunge:
Hello @pparkkinen ,

First of all thanks for submitting this PR and helping us improve the Log Forwarder and sorry for the delay.
Would you mind letting us know if you are targeting a specific service when trying to get the tags?

As I would expect that the tags you want to get would be on the service directly (lambda, ...) and not on the log group itself.

pparkkinen:
What we use this for is logs from Lambdas. I wasn't sure whether the log forwarder lambda would really be able to know which lambda (or service) the logs were coming from, and whether it would have access to the tags for the lambdas, so the way I did it is that it gets the tags from the log groups. Also that way the log forwarder wouldn't have to care where the logs are coming from.

And personally I'm not really sure the log forwarder lambda should really care about what service the logs are coming from. All it's doing is forwarding the logs. And if the logs should be tagged in some way, it would make sense to me to have the tags on the logs (the log groups).

dalen:
This seems like a great addition. We have also modified the log forwarder to be able to get tags service and environment tags on our logs from Fargate services. Getting some ability to do that upstream would be nice.

pparkkinen:
Looks like there is already a configuration option to get tags from Lambdas, which solves our use case. I'm closing this PR.

PR #146, opened 2019-08-29
Tag lambda logs with env and service

tianchu:
### What does this PR do?

Tag lambda logs with `env:none` and `service:<function_name>` (the defaults).

### Motivation

To enable log and apm correlation.

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -154,7 +154,7 @@ DD_SOURCE = "ddsource"
 DD_CUSTOM_TAGS = "ddtags"
 DD_SERVICE = "service"
 DD_HOST = "host"
-DD_FORWARDER_VERSION = "1.5.0"
+DD_FORWARDER_VERSION = "1.5.1"
 
 # Pass custom tags as environment variable, ensure comma separated, no trailing comma in envvar!
 DD_TAGS = os.environ.get("DD_TAGS", "")
@@ -698,6 +698,10 @@ def awslogs_handler(event, context, metadata):
                 metadata[DD_CUSTOM_TAGS] += ",functionname:" + function_name
                 # Set the arn as the hostname
                 metadata[DD_HOST] = arn
+                # Default `env` to `none` and `service` to the function name,
+                # for correlation with the APM env and service.
+                metadata[DD_SERVICE] = function_name
+                metadata[DD_CUSTOM_TAGS] += ",env:none"
 
     # Create and send structured logs to Datadog
     for log in logs["logEvents"]:


aws/logs_monitoring/lambda_function.py
```diff
@@ -698,6 +699,11 @@ def awslogs_handler(event, context, metadata):
                 metadata[DD_CUSTOM_TAGS] += ",functionname:" + function_name
                 # Set the arn as the hostname
                 metadata[DD_HOST] = arn
+                # Default `env` to `none` and `service` to the function name,
+                # for correlation with the APM env and service.
+                metadata[DD_ENV] = "none"
+                metadata[DD_SERVICE] = function_name
+                metadata[DD_CUSTOM_TAGS] += ",env:none,service:" + function_name
```
blemale: Why are you also adding them to the custom tags?

tianchu: It appears that when navigating from a log to APM service (by clicking the service name from the sliding panel with log entry's details), the `env` value from tags (rather than the one from attributes) is used to populate the navigation URL.

Technically speaking, I only need to set the service attribute and env tag. But I felt that it might be better to set both of service and env both as an attribute and tag, just for consistency, unless this creates some drawbacks?

blemale: I lack of knowledge on that part, WDYT @NBParis ?

NilsBunge: There is no need to add the `service` as tag as we rely on the Reserved attribute on logs (which here is extracted from the `service` attribute that is set thanks to `DD_SERVICE`).

Now it's indeed a great idea to set the service to the function name in addition of having it as `functionname` tag (which is used to correlate with metrics).

Regarding the env environment, you are indeed correct. If no environment has been specified on traces, they will have `env:none` as tag so adding it by default would be a good idea. But I believe it would be rather more interesting to have it set through environment variable and if not to default it to `none`. What do you think about this?

So we could:

- Check if there is a `env` tag in the `DD_TAGS` environment variable
- If there is one we don't do anything
- If there is none we add `env:none` to the logs. 

Do you think that would work?

tianchu: Thanks for your quick feedback @NBParis!

I will set only set `service` as attribute and `env` as tag (done).

I don't think we want users to set `env` via `DD_TAGS` though, because it likely varies function by function (some smaller companies use one aws account and region for all environments, but likely using only one datadog forwarder). 

On the APM side, we already finished the work to allow users tag lambda functions with `env` and `service` in order to customize the `env` and `service` for the corresponding apm service. I believe you guys are still pushing for the work to apply lambda tags to logs on the backend, once that's done, the `env` tag from the individual lambda function should automatically override the default value `none` set by the forwarder here.

NilsBunge: That's perfect then. Let's proceed like this.

blemale (approved):

PR #147, opened 2019-09-03
use SSLcontext.wrap_socket() instead of ssl.wrap_socket()

chengpanmlp:
### What does this PR do?

use SSLContext.wrap_socket() instead of ssl.wrap_socket  Since Python 3.2 and 2.7.9, it is recommended to use the SSLContext.wrap_socket() instead of wrap_socket(). The top-level function is limited and creates an insecure  client socket without server name indication or hostname matching.

### Motivation

For users that use the Datadog Lambda function behind a transparent proxy that uses SNI, it is recommended to use the newer library. 

### Additional Notes

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -211,7 +211,7 @@ class DatadogTCPClient(object):
 
     def _connect(self):
         sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-        sock = ssl.wrap_socket(sock)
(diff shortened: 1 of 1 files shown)


gaetan-deputier:
Hey! Thanks for your contribution!
I agree with using SSLContext.wrap_socket() but I'd have the hostname validation behind a flag disabled by default.

gaetan-deputier:
Here is the reference to the master branch: https://github.com/DataDog/datadog-serverless-functions/pull/208

We don't need a flag for that, apologies.

I'll close this branch and ensure the other one gets in. Thanks again for your contrib.

PR #149, opened 2019-09-05
[Lambda]: Default the transport protocol to HTTP

gaetan-deputier:
[General]: Introduce github action to lint the lambda code

### What does this PR do?
[Lambda]: Default the transport protocol to HTTP & update the default batcher variables to higher values

[General]: Introduce github action to lint the lambda code

### Motivation

The HTTP protocol is more adapted for the lambda traffic patterns. Defaulting it is the natural next step now.

### DIFF
--- a/.github/workflows/lambdachecks.yml
+++ b/.github/workflows/lambdachecks.yml
@@ -1,0 +1,25 @@
+name: Lambda checks
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        python-version: [2.7]
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install pip
+      run: |
+        python -m pip install --upgrade pip
+    - name: Lint with flake8
+      run: |
+        pip install flake8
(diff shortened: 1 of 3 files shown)


.github/workflows/lambdachecks.yml
```diff
@@ -0,0 +1,25 @@
...
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        python-version: [2.7]
```
achntrl: Should we also add 3.7 now that the lambda is py3 compatible ?

achntrl (approved):

aws/logs_monitoring/README.md
```diff
@@ -75,6 +75,12 @@ You have two options to add custom tags to your logs:
 
 ## 3. (optional) Send logs to EU or to a proxy
 
+### Send logs through TCP
+
+By default, the forwarder sends logs using HTTPS through the port 443. 
+
+If you would rather use raw TCP, set the environment variable `DD_USE_TCP` to `true`.
```
NilsBunge: I would replace this by: `To send logs over a SSL encrypted TCP connection, set ....`

.github/workflows/lambdachecks.yml
gaetan-deputier: Added but restricted the checks to logs-monitoring for now

ajacquemot (approved):

PR #150, opened 2019-09-05
Trace Forwarding

DarcyRaynerDD:
### What does this PR do?
Adds experimental trace forwarding support to the log forwarder. More info can be found in the [RFC here](https://github.com/DataDog/architecture/pull/462). Also has an optional dependency on a [new lambda layer](https://github.com/DataDog/datadog-trace-forwarder) which supports trace forwarding + trace obfuscation.

### Motivation

Supporting DataDog tracing in Lambda environments.

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -145,7 +145,6 @@ if not validation_res.ok:
 trace_connection = None
 if DD_FORWARD_TRACES:
     trace_connection = TraceConnection("https://trace.agent.{}".format(DD_SITE), DD_API_KEY)
-trace_regex = re.compile(r"^\s*{\s*\"traces\"\s*\:\s*\[")
 
 # DD_MULTILINE_LOG_REGEX_PATTERN: Datadog Multiline Log Regular Expression Pattern
 DD_MULTILINE_LOG_REGEX_PATTERN = None
@@ -387,7 +386,6 @@ class DatadogScrubber(object):
 
 def datadog_forwarder(event, context):
     """The actual lambda function entry point"""
-    global trace_connection
     events = parse(event, context)
     metrics, logs, traces = split(events)
     if DD_FORWARD_LOG:
@@ -480,9 +478,11 @@ def generate_metadata(context):
 def extract_trace(event):
     """Extract traces from an event if possible"""
     try:
+        print(event)
         message = event["message"]
-        if not trace_regex.match(message):
-            return
+        obj = json.loads(event['message'])
+        if not "traces" in obj or not isinstance(obj["traces"], list):
+            return None
         return message
     except Exception:
         return None
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -24,11 +24,14 @@ try:
     # Datadog Lambda layer is required to forward metrics
     from datadog_lambda.wrapper import datadog_lambda_wrapper
     from datadog_lambda.metric import lambda_stats
+    from datadog_lambda.trace_intake import forward_traces
+
     DD_FORWARD_METRIC = True
+    DD_FORWARD_TRACES = True
 except ImportError:
     # For backward-compatibility
     DD_FORWARD_METRIC = False
-
+    DD_FORWARD_TRACES = False
 
 # Set this variable to `False` to disable log forwarding.
 # E.g., when you only want to forward metrics from logs.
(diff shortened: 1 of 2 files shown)


aws/logs_monitoring/lambda_function.py
```diff
@@ -374,15 +387,22 @@ def scrub(self, payload):
...
     """The actual lambda function entry point"""
+    global trace_connection
     events = parse(event, context)
-    metrics, logs = split(events)
+    metrics, logs, traces = split(events)
     if DD_FORWARD_LOG:
         forward_logs(filter_logs(logs))
     if DD_FORWARD_METRIC:
         forward_metrics(metrics)
+    if DD_FORWARD_TRACES and len(traces) > 0:
```
tianchu, line 395: Can you refactor this into a similar `forward_traces`?

```python
if DD_FORWARD_TRACES:
    forward_traces(traces)
```

aws/logs_monitoring/lambda_function.py
```diff
@@ -476,14 +507,17 @@ def extract_metric(event):
 
 def split(events):
     """Split events to metrics and logs"""
-    metrics, logs = [], []
+    metrics, logs, traces = [], [], []
     for event in events:
         metric = extract_metric(event)
+        new_trace = extract_traces(event)
```
tianchu: Why it's named `new`?

aws/logs_monitoring/lambda_function.py
```diff
@@ -476,14 +507,17 @@ def extract_metric(event):
...
         if metric:
             metrics.append(metric)
+        elif new_trace:
+            traces = traces + [new_trace]
```
tianchu: Why not `traces.append`? Maybe i missed something.

aws/logs_monitoring/lambda_function.py
```diff
@@ -461,6 +481,17 @@ def generate_metadata(context):
     return metadata
 
 
+def extract_traces(event):
```
tianchu: Is it `trace` or `traces`? I believe one message can only contain spans from the same trace?

aws/logs_monitoring/lambda_function.py
```diff
@@ -374,15 +387,22 @@ def scrub(self, payload):
...
-    metrics, logs = split(events)
+    metrics, logs, traces = split(events)
     if DD_FORWARD_LOG:
         forward_logs(filter_logs(logs))
     if DD_FORWARD_METRIC:
         forward_metrics(metrics)
+    if DD_FORWARD_TRACES and len(traces) > 0:
+        try:
+            for trace in traces:
+                trace_connection.send_trace(trace)
```
tianchu: Is `send_trace` blocking or non-blocking?

DarcyRaynerDD: Blocking

tianchu (approved):

tt810 (approved):

gaetan-deputier (dismissed):

aws/logs_monitoring/lambda_function.py
```diff
@@ -374,15 +387,18 @@ def scrub(self, payload):
 
 def datadog_forwarder(event, context):
     """The actual lambda function entry point"""
+    global trace_connection
```
gaetan-deputier: The global comes from the layer? Why having a global instead of creating the connection here?

aws/logs_monitoring/lambda_function.py
```diff
@@ -461,6 +477,17 @@ def generate_metadata(context):
     return metadata
 
 
+def extract_trace(event):
+    """Extract traces from an event if possible"""
+    try:
+        message = event["message"]
+        if not trace_regex.match(message):
```
gaetan-deputier: Can't we use something else than a regex to match? Aka reuse the event field maybe ?

aws/logs_monitoring/lambda_function.py
ajacquemot: We could probably move this down to `forward_traces` ?

aws/logs_monitoring/lambda_function.py
ajacquemot: if the schema is static, I think it's probably better to check the existence of the attribute in the map, WDYT ?

ajacquemot (commented):
Thanks for the PR, left two comments

aws/logs_monitoring/lambda_function.py
DarcyRaynerDD: Thanks. I've removed the regex, and parsed the line as json instead to check for the field.

aws/logs_monitoring/lambda_function.py
DarcyRaynerDD: I've removed this.

PR #152, opened 2019-09-12
Decode DD_API_KEY into UTF-8 

c-py:
### What does this PR do?

Seems like python3 fails when using DD_KMS_API_KEY as DD_API_KEY is type bytes. This decodes this into utf-8 if the key is of type bytes.


tianchu (approved):

tianchu:
Actually `type(DD_API_KEY) is bytes` evaluates to `True` in python 2.7 as well, since `bytes` is simply an alias for `str` prior to python 3. But I think what you have is actually better (more explicit and defensive). 

Thanks for the PR! Really appreciate!

PR #154, opened 2019-09-20
[Lambda]: Add a NO SSL option to disable SSL for TCP and support proxy for HTTP

gaetan-deputier:
### What does this PR do?

- Add a NO_SSL option to the lambda forwarder for TCP proxying
- Support custom host/port for proxy

### Motivation

Enhance proxy support for the logs lambda forwarder

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -9,16 +9,16 @@ import base64
 import gzip
 import json
 import os
-import re
-import socket
-from botocore.vendored import requests
-import time
-import ssl
-import six.moves.urllib as urllib  # for for Python 2.7 urllib.unquote_plus
-import itertools
-from io import BytesIO, BufferedReader
 
 import boto3
+import itertools
+import re
+import six.moves.urllib as urllib  # for for Python 2.7 urllib.unquote_plus
+import socket
+import ssl
+import time
+from botocore.vendored import requests
+from io import BytesIO, BufferedReader
 
 try:
     # Datadog Lambda layer is required to forward metrics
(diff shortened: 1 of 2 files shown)


NilsBunge (approved):
Looks good on documentation side. Thanks for this

aws/logs_monitoring/lambda_function.py
```diff
@@ -266,8 +271,8 @@ class DatadogHTTPClient(object):
     _POST = "POST"
     _HEADERS = {"Content-type": "application/json"}
 
-    def __init__(self, host, api_key, scrubber, timeout=10):
-        self._url = "https://{}/v1/input/{}".format(host, api_key)
+    def __init__(self, host, port, api_key, scrubber, timeout=10):
+        self._url = "https://{}:{}/v1/input/{}".format(host, port, api_key)
```
ajacquemot: Do we want to send traffic in clear over `http` ? If so, I think we probably want to use `http` as scheme instead of `https`

aws/logs_monitoring/lambda_function.py
```diff
@@ -48,21 +48,24 @@
 # by default, use the TCP client.
 DD_USE_TCP = os.getenv("DD_USE_TCP", default="false").lower() == "true"
 
+# Set this value to disable SSL over our TCP client.
+# Useful when you are forwarding your logs via TCP to a proxy.
+DD_NO_SSL = os.getenv("DD_NO_SSL", default="false").lower() == "true"
```
ajacquemot: this lines is very similar to line 49, we could probably create a helper to collect a boolean from environment variables.

ajacquemot (approved):

aws/logs_monitoring/lambda_function.py
gaetan-deputier: Updated

aws/logs_monitoring/lambda_function.py
gaetan-deputier: Updated

PR #156, opened 2019-09-23
Readme-organisation

NilsBunge:
### What does this PR do?

Reorganise the read me so that the AWS main part is just a link to the other functions

### Motivation

Avoid confusion in the setup

### Additional Notes


cswatt (approved):
approved from docs

PR #157, opened 2019-09-23
Adds ability to set DD_API_KEY in AWS secrets manager

mlehner616:
### What does this PR do?

This allows a new method to set the datadog API key. This would allow you to place the DD_API_KEY as a string within AWS secrets manager and avoid the need to set an environment variable directly and also avoid the need for KMS keys directly. Secrets manager secrets allow more granular permissions control than these other options.


vhiairrassary:
Our team store most of our secrets in AWS Secrets Manager. It would be a great addition if possible.

tianchu:
I'm going to `Output` the function arn in PR: https://github.com/DataDog/datadog-serverless-functions/pull/171/files#diff-3f76f158c2bc9ff25737be780caeb370R61

tianchu:
Thanks for your PR! I'm going to include it in https://github.com/DataDog/datadog-serverless-functions/pull/171

PR #159, opened 2019-10-01
Parse REPORT log to generate and submit memory, duration, and estimated cost metrics

swfunc:
### What does this PR do?

Parses the Lambda REPORT log line to submit these metrics to Datadog from within the log forwarder Lambda function:

- aws.lambda.enhanced.estimated_cost
- aws.lambda.enhanced.max_memory_used
- aws.lambda.enhanced.duration
- aws.lambda.enhanced.billed_duration

If the env var DD_FETCH_CUSTOM_TAGS is set to "true", the log forwarder will fetch and cache custom tags of Lambdas in the account in order to apply the custom tags to the enhanced metrics. The log forwarder will increment the metric aws.lambda.enhanced.get_resources_api_calls on each request.

I've updated the demo account to use this version for a few functions' log groups. You can see the enhanced metrics reporting to the Datadog internal account.


NilsBunge:
Hello @sfirrin ,

So this should be done automatically on Datadog side.
All the processing is happening on Datadog to avoid doing any of those in the collectors.

So if the logs are properly tagged with the source `lambda` (which should be automatically if `lambda` is [contained in the log group](https://github.com/DataDog/datadog-serverless-functions/blob/master/aws/logs_monitoring/lambda_function.py#L844)) those should be parsed.

Did you get unparsed logs for the Lambda logs?

swfunc:
Hi Nils, this is a little different from using logs analytics to parse the REPORT log since this is submitting the data as normal infrastructure metrics that don't require Datadog to ingest logs. I.e. customers can get these metrics even if they can't forward logs to Datadog. The RFC here has more details about the motivation https://github.com/DataDog/architecture/blob/d9c25772552f3ee93b00a81542dbff8080babaaf/rfcs/serverless/lambda-enhanced-integration-metrics/rfc.md

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+        log.setLevel(logging.WARN)
+
+
+set_log_level_to_env_var()
+
+
+def should_fetch_custom_tags():
+    """Checks the env var to determine if the customer has opted-in to fetching custom tags
+    """
+    env_var_value = os.environ.get(FETCH_CUSTOM_TAGS_ENV_VAR_NAME)
```
tianchu: Should we update https://github.com/DataDog/datadog-serverless-functions/blob/master/aws/logs_monitoring/README.md, or you are trying to keep it as an experimental (undocumented) feature for now?

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
+import logging
```
tianchu: How about `enhanced_lambda_metrics.py` for additional explicity?

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+import os
+import re
+
+from collections import defaultdict
+from time import time
+
+import boto3
+from botocore.exceptions import ClientError
+from datadog_lambda.metric import lambda_metric
```
tianchu: Until we (semi) automate the upgrade process, I suspect existing users will keep upgrading by copy & pasting source code, without realizing that the lambda layer is now required. This will break the forwarder.

Basically, the forwarder should work for log forwarding without the lambda layer and this added new file, while the enhanced metrics only work when the layer and new file are added to the forwarder.

aws/logs_monitoring/lambda_function.py
```diff
@@ -16,8 +16,11 @@
 import ssl
 import six.moves.urllib as urllib  # for for Python 2.7 urllib.unquote_plus
 import itertools
+
 from io import BytesIO, BufferedReader
 
+from enhanced_metrics import parse_and_submit_enhanced_metrics
```
tianchu: Until we (semi) automate the upgrade process, I suspect existing users may not notice the new file for copy & paste, and the forwarder will break on this line?

How about following the pattern from line 27 ~ 43?

```python
try:
    # Datadog Lambda layer is required to submit enhanced lambda metrics
    import datadog_lambda
    DD_ENHANCED_LAMBDA_METRICS = True
except:
    DD_ENHANCED_LAMBDA_METRICS = False
```

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+
+
+set_log_level_to_env_var()
+
+
+def should_fetch_custom_tags():
+    """Checks the env var to determine if the customer has opted-in to fetching custom tags
+    """
+    env_var_value = os.environ.get(FETCH_CUSTOM_TAGS_ENV_VAR_NAME)
+    return env_var_value and env_var_value.lower() == "true"
```
tianchu: I think a oneliner actually is simpler to read and gives you a better sense of the default? 

```python
return os.environ.get(FETCH_CUSTOM_TAGS_ENV_VAR_NAME, "false").lower() == "true"
```

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+
+# Latest Lambda pricing per https://aws.amazon.com/lambda/pricing/
+BASE_LAMBDA_INVOCATION_PRICE = 0.0000002
+LAMBDA_PRICE_PER_GB_SECOND = 0.0000166667
+
+ESTIMATED_COST_METRIC_NAME = "estimated_cost"
+
+GET_RESOURCES_LAMBDA_FILTER = "lambda"
+
+FETCH_CUSTOM_TAGS_ENV_VAR_NAME = "DD_FETCH_CUSTOM_TAGS"
```
tianchu: How about `DD_FETCH_LAMBDA_TAGS`? Probably reserve `DD_FETCH_CUSTOM_TAGS` for the future, when the forwarder fetch tags for other log sources too?

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+
+from collections import defaultdict
+from time import time
+
+import boto3
+from botocore.exceptions import ClientError
+from datadog_lambda.metric import lambda_metric
+
+
+ILLEGAL_TAG_CHARACTERS = {":", "/", "-"}
```
tianchu: https://docs.datadoghq.com/developers/guide/what-best-practices-are-recommended-for-naming-metrics-and-tags/

do we need to worry about other special characters? say `@` or `$`? How are we dealing with this in the aws tag crawlers?

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+
+        enhanced_metrics += parsed_metrics
+
+    return enhanced_metrics
+
+
+# Names to use for metrics and for the named regex groups
+REQUEST_ID_FIELD_NAME = "request_id"
+DURATION_METRIC_NAME = "duration"
+BILLED_DURATION_METRIC_NAME = "billed_duration"
```
tianchu: Why not define all the constants at the top?

tianchu: Nits, I personally prefer defining related constants in class, so they can be used like `ENHANCED_METRICS.DURATION` and etc.

```python
class ENHANCED_METRICS:
    DURATION = "duration"
    ....
```

aws/logs_monitoring/tests/test_enhanced_metrics.py
```diff
@@ -0,0 +1,433 @@
+import unittest
```
tianchu: 👍

aws/logs_monitoring/lambda_function.py
```diff
@@ -407,12 +416,16 @@ def datadog_forwarder(event, context):
     events = parse(event, context)
     metrics, logs, traces = split(events)
     if DD_FORWARD_LOG:
-        forward_logs(filter_logs(logs))
+        log_event_strings = map(json.dumps, logs)
```
tianchu: Nits, I personally feel the original pipeline-like onliner is easier to read -- how about `forward_logs(filter_logs(map(json.dumps, logs)))`?

aws/logs_monitoring/lambda_function.py
```diff
@@ -407,12 +416,16 @@ def datadog_forwarder(event, context):
...
+        filtered_logs = filter_logs(log_event_strings)
+        forward_logs(filtered_logs)
     if DD_FORWARD_METRIC:
         forward_metrics(metrics)
     if DD_FORWARD_TRACES and len(traces) > 0:
         forward_traces(traces)
 
+    parse_and_submit_enhanced_metrics(logs)
```
tianchu: I would avoid passing the entire set of `logs` to `parse_and_submit_enhanced_metrics`, though it's the last line of code here, it still means anything can happen to any log entry within `parse_and_submit_enhanced_metrics`. If I'm new to the code, I would have to drill into enhanced metrics code and realize that not all the logs are used by that function for instance. 

In general, I would prefer the "main" function to make the actual data flow very explicit. 

I think it's also better to front-load the filtering logic (only Lambda REPORT logs) here.

```python
if DD_ENHANCED_LAMBDA_METRICS:
    lambda_report_logs = extract_lambda_report_logs(logs)
    generate_enhanced_lambda_metrics(lambda_report_logs)
```

or 
```python
metrics, traces, logs, lambda_report_logs = process(events)
......
if DD_ENHANCED_LAMBDA_METRICS:
    generate_enhanced_lambda_metrics(lambda_report_logs)
```

aws/logs_monitoring/lambda_function.py
```diff
@@ -110,20 +113,26 @@ def __init__(self, name, pattern, placeholder):
...
 ]
 
 # Use for include, exclude, and scrubbing rules
 def compileRegex(rule, pattern):
     if pattern is not None:
         if pattern == "":
             # If pattern is an empty string, raise exception
-            raise Exception("No pattern provided:\nAdd pattern or remove {} environment variable".format(rule))
+            raise Exception(
+                "No pattern provided:\nAdd pattern or remove {} environment variable".format(
```
tianchu, line 155: 👍 Thanks for the reformatting!

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+    Args:
+        logs (dict<str, multiple types>[]): the logs parsed from the event in the split method
+            See docstring below for an example.
+    """
+    # Wrap everything in try/catch to prevent failing the Lambda on enhanced metrics
+    try:
+        enhanced_metrics = generate_enhanced_lambda_metrics(
+            logs, account_lambda_tags_cache
+        )
+        for enhanced_metric in enhanced_metrics:
```
tianchu: Does this mean if one log entry causes an exception in `generate_enhanced_lambda_metrics` none of the already generated metrics will be submitted?

How about something like this?
```python
for log in logs:
    try:
        # generate metrics from `log`
        # submit metrics from `log`
    except Exception:
        log.exception("Error processing lambda report log: %s", log)
```

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+# Latest Lambda pricing per https://aws.amazon.com/lambda/pricing/
+BASE_LAMBDA_INVOCATION_PRICE = 0.0000002
+LAMBDA_PRICE_PER_GB_SECOND = 0.0000166667
+
+ESTIMATED_COST_METRIC_NAME = "estimated_cost"
+
+GET_RESOURCES_LAMBDA_FILTER = "lambda"
+
+FETCH_CUSTOM_TAGS_ENV_VAR_NAME = "DD_FETCH_CUSTOM_TAGS"
+LOG_LEVEL_ENV_VAR_NAME = "DD_LOG_LEVEL"
```
tianchu: Just personal preference, I don't think it's worth abstracting the environment variable name to a constant -- it's usually read in only one place of the code and it does not change often.

tianchu: Same for `FETCH_CUSTOM_TAGS_ENV_VAR_NAME`.

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+    env_var_log_level = os.environ.get(LOG_LEVEL_ENV_VAR_NAME, "").lower()
+    if env_var_log_level == "debug":
+        log.setLevel(logging.DEBUG)
+    if env_var_log_level == "error":
+        log.setLevel(logging.ERROR)
+    if env_var_log_level == "warn" or env_var_log_level == "warning":
+        log.setLevel(logging.WARN)
+
+
+set_log_level_to_env_var()
```
tianchu: How about this one-liner?

```python
log.setLevel(logging.getLevelName(os.environ.get("DD_LOG_LEVEL", "INFO").upper()))
```

Also, you probably should do this in the `lambda_function.py` since you are setting log level on the `root` logger, which means it affects the log level in other files. You can limit the impact by modifying the log level only on the local logger `logging.getLogger(__name__)`. However, i believe `DD_LOG_LEVEL` should affect the entire forwarder.

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+
+
+def should_fetch_custom_tags():
+    """Checks the env var to determine if the customer has opted-in to fetching custom tags
+    """
+    env_var_value = os.environ.get(FETCH_CUSTOM_TAGS_ENV_VAR_NAME)
+    return env_var_value and env_var_value.lower() == "true"
+
+
+def sanitize_aws_tag_string(raw_string):
```
tianchu: Not 100%, but `re.sub()` could do this in a oneliner?

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+        aws_key_value_tag_dict (dict): the dict the GetResources endpoint returns for a tag
+            ex: { "Key": "creator", "Value": "swf"}
+
+    Returns:
+        key:value colon-separated string built from the dict
+            ex: "creator:swf"
+    """
+    key = sanitize_aws_tag_string(aws_key_value_tag_dict["Key"])
+    value = sanitize_aws_tag_string(aws_key_value_tag_dict["Value"])
+    return "{}:{}".format(key, value)
```
tianchu: I believe `value` is both optional in AWS and Datadog?

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+
+    def _are_tags_out_of_date(self):
+        """Returns bool for whether the tag fetch TTL has expired
+        """
+        earliest_time_to_refetch_tags = (
+            self.last_tags_fetch_time + self.tags_ttl_seconds
+        )
+        return time() > earliest_time_to_refetch_tags
+
+    def get_lambda_tags(self, resource_arn):
```
tianchu: nits, the class name already implied its purpose, it's unnecessary to repeat `lambda` and `tags` in its method names. Also it's a typical cache class, it would be nice to follow the convention cache api.

```python
class LambdaTagsCache(object):
    def set():  # perhaps not needed in your use case
        ...
    def get():
        ...
    def is_expired():
        ...
    def refresh():
        ...
```

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+    Args:
+        report_log_line (str): The REPORT log generated by Lambda
+        EX: "REPORT RequestId: 814ba7cb-071e-4181-9a09-fa41db5bccad	Duration: 1711.87 ms	\
+            Billed Duration: 1800 ms	Memory Size: 128 MB	Max Memory Used: 98 MB	\
+            XRAY TraceId: 1-5d83c0ad-b8eb33a0b1de97d804fac890	SegmentId: 31255c3b19bd3637	Sampled: true"
+
+    Returns:
+        metrics - DatadogMetricPoint[]
+    """
+    regex_search = re.compile(REPORT_LOG_REGEX_SEARCH)
```
tianchu: There is no need to compile a regex multiple times. The convention is to make a global.

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+    regex_match = regex_search.search(report_log_line)
+
+    if not regex_match:
+        return []
+
+    metric_tags = []
+
+    for tag_key in TAGS_TO_PARSE_FROM_REPORT:
```
tianchu: I think it's way more useful to submit `aws.lambda.memorysize` as a metric rather than a tag, it doesn't change very often, and user would like to graph it too. I would drop this one and log a ticket in backlog for memorysize metric.

aws/logs_monitoring/enhanced_metrics.py
tianchu: You should use `from datadog_lambda.metric import lambda_stats` instead (see https://github.com/DataDog/datadog-serverless-functions/blob/master/aws/logs_monitoring/lambda_function.py#L26), so it doesn't automatically tag the metric with `dd_lambda_layer`. The value for this tag will be the lambda layer attached to the forwarder, rather than the layer attached to the customer's lambda function, it's pretty much useless.

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+    # Cap the number of times to split
+    split_arn = arn.split(":")
+
+    # If ARN includes version / alias at the end, drop it
+    if len(split_arn) > 7:
+        split_arn = split_arn[:7]
+
+    _, _, _, region, account_id, _, function_name = split_arn
+
+    return [
```
tianchu: Just flagging as a reminder, without summaries enabled for distribution metrics, only 4 tags are allowed.

aws/logs_monitoring/enhanced_metrics.py
```diff
@@ -0,0 +1,445 @@
...
+
+        parsed_metrics = parse_metrics_from_report_log(log_message)
+        if not parsed_metrics:
+            continue
+
+        # Add the tags from ARN, custom tags cache, and env var
+        tags_from_arn = parse_lambda_tags_from_arn(log_function_arn)
+        lambda_custom_tags = tags_cache.get_lambda_tags(log_function_arn)
+        # "ddtags" is the name of the key in the log event dict
+        tags_from_env_var = log.get("ddtags").split(",") if log.get("ddtags") else []
```
tianchu: I would not apply these tags for now, as the existing tags are mostly meant for logs, and I do not want user to lose tags in the future if we change our minds and generate these metrics from somewhere else. 

Also most of users can only have 4 tags, and submitting > 4 tags will completely ruin the metric. `lambda_custom_tags` is fine, since it's opt-in and we can make sure customers have distribution summaries enabled before enabling it.

aws/logs_monitoring/lambda_function.py
```diff
@@ -16,8 +16,11 @@
 import ssl
 import six.moves.urllib as urllib  # for for Python 2.7 urllib.unquote_plus
 import itertools
+
```
tianchu: Don't forget to bump up the version: https://github.com/DataDog/datadog-serverless-functions/pull/159/files#diff-618731ddf1c446ef45d09613cb4b189bR203

aws/logs_monitoring/enhanced_metrics.py
swfunc: Good point, I thought it was only this three characters allowed in AWS tags but there are more https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions (confirmed by trying to add tags with and without those chars in my test lambda).

In the AWS tag crawlers I know we replace the illegal chars with underscores also, I just spent some time trying to find where it happens in the code but I gave up looking after a while.

I'm going to update this set to include all the AWS chars in that doc.

aws/logs_monitoring/enhanced_metrics.py
swfunc: I'll add info to the readme

aws/logs_monitoring/enhanced_metrics.py
swfunc: indeed, `re.sub(r'[\+\-=\.:/@]', '_', raw_string)`

aws/logs_monitoring/enhanced_metrics.py
swfunc: good point, fixed

aws/logs_monitoring/lambda_function.py
```diff
@@ -188,7 +217,8 @@ def compileRegex(rule, pattern):
 DD_CUSTOM_TAGS = "ddtags"
 DD_SERVICE = "service"
 DD_HOST = "host"
-DD_FORWARDER_VERSION = "2.0.0"
+DD_FORWARDER_VERSION = "2.1.0"
```
tianchu: Not sure, but 2.1.0 could have already existed. Perhaps it's better update version in a separate PR before publishing to SAM repo?

tianchu (approved):

aws/logs_monitoring/lambda_function.py
swfunc: ah yes you're right, it's currently 2.1.0 https://github.com/DataDog/datadog-serverless-functions/blame/1a7cda80b13496f7fa0192475861f6f535bdf49f/aws/logs_monitoring/lambda_function.py#L204

PR #160, opened 2019-10-03
[Don't merge] Add default concurrency limit to SAM template

pinkerton:
### What does this PR do?

Adds a default concurrency limit to the SAM configuration used for doing one-click setup of the Lambda log forwarder.

### Motivation

We recommend this in our public docs and is supported in SAM https://github.com/awslabs/serverless-application-model/releases/tag/1.4.0

This prevents the log forwarder from consuming too much account-wide concurrency.

### Additional Notes

I have not tested this yet.


tianchu:

PR #164, opened 2019-10-11
Expose serverless application parameters (DD_API_KEY, DD_SITE)

philhoff-edeka:
### What does this PR do?

Exposes the following parameters for the serverless application:
* `DatadogApiKey` (for environment variable: `DD_API_KEY`)
* `DatadogSite` (for environment variable: `DD_SITE`)

### Motivation

Exposing the API key and site as parameters for the serverless application allows others to use the app without modifications to the Python code.

### Additional Notes

If you like the idea, additional parameters (for example custom tags) could be exposed similarly.


tianchu:

PR #166, opened 2019-10-11
made possible custom retention for lambda's log group

vagharshakus:
### What does this PR do?

It will be possible to define custom retention parameter for the log group, which is attached to the lambda function. If parameter's value is null, no new log group resource will be created.

### Motivation

It would be nice to have the ability to customize the retention.

### Additional Notes

No :-)


tianchu:

PR #163, opened 2019-10-11
Add possibility to customize function name

gruebel:
### What does this PR do?

Adds the possibility to give the function a custom name.

### Motivation

To be able to reuse the function easily cross mutliple CloudFormation templates, it is easier to refernce a human readable name, than the auto-generated one.

### Additional Notes

:)

PR #162, opened 2019-10-11
Remove duplicate key 'Type'

philhoff-edeka:
### What does this PR do?

Removes the duplicate key `Type` from the SAM template.

### Motivation

Fixes the following error in the IDE: `Key 'Type' is duplicated`

PR #171, opened 2019-10-25
Forwarder 3.0.0

tianchu:
### What does this PR do?

Easy install and upgrade:
- Provide CloudFormation quick launch URL for "one-click" install ([try me out](https://console.aws.amazon.com/cloudformation/home#/stacks/new?stackName=datadog-serverless&templateURL=https://dd-log-sam.s3.amazonaws.com/templates/3.0.0.yaml))
- Host versioned CloudFormation templates in S3 for upgrade (rollback) to a specific version
- Add required IAM permissions to the execution role automatically
- Keep forwarder existing settings, such as environment variables through upgrade
- Group parameters into multiple groups for an organized user interface

Support infrastructure as code:
- Expose all the forwarder settings, such as memory size, timeout and `DD_` environment variables as CloudFormation parameters
- Export the forwarder lambda ARN as a CloudFormation output

General improvements:
- Upgrade the Lambda runtime from Python 2.7 to Python 3.7
- Store Datadog API Key in AWS Secrets Manager by default
- Reduce forwarder's own log retention default from forever to 90 days
- Add a release script for release automation
- Use github releases to notify users of new versions
- Attach function.zip to github releases for manual install/upgrade if needed
- Add additional debug logs

Related PRs from the community:
https://github.com/DataDog/datadog-serverless-functions/pull/157
https://github.com/DataDog/datadog-serverless-functions/pull/160
https://github.com/DataDog/datadog-serverless-functions/pull/162
https://github.com/DataDog/datadog-serverless-functions/pull/163
https://github.com/DataDog/datadog-serverless-functions/pull/164
https://github.com/DataDog/datadog-serverless-functions/pull/166
https://github.com/DataDog/datadog-serverless-functions/pull/176
https://github.com/DataDog/datadog-serverless-functions/pull/183


aws/logs_monitoring/lambda_function.py
```diff
@@ -255,7 +268,7 @@ def compileRegex(rule, pattern):
 DD_CUSTOM_TAGS = "ddtags"
 DD_SERVICE = "service"
 DD_HOST = "host"
-DD_FORWARDER_VERSION = "2.3.4"
+DD_FORWARDER_VERSION = get_env_var("DD_FORWARDER_VERSION", "unknown")
```
gaetan-deputier: Why using unknown here? If the forwarder version is not set, we could still default on something known aka 3.0.0 here.

aws/logs_monitoring/README.md
```diff
@@ -1,188 +1,41 @@
...
-
-For example:
-
-```
-arn:aws:lambda:us-east-1:464622532012:layer:Datadog-Python37:8
-```
-
-
-### 4. Set your Parameters
-
```
gaetan-deputier, line 54: Why removing the parameters doc?

tianchu: I'm planning to copy over the doc to the parameters' description: https://github.com/DataDog/datadog-serverless-functions/pull/171/files#diff-7463e3e489db7e82f7ac591389236943R53, so it's only documented in one place (the cloudformation template and thus reflected in the stack creation/update user interface).

aws/logs_monitoring/lambda_function.py
tianchu: I was merely trying to avoid setting it in two places -- the template and source code, which could potentially result in unmatched versions set. Especially given that this fallback logic rarely kicks in, developers could easily forget to update the version number in source code (here).

The reason I extracted version to the template and set it through an environment variable was to make it available to the existing lambda metadata crawler, so we can track the usage internally and present users a nudge to upgrade based on the version. Given this plan, I thought "unknown" is a better fallback, as it also gives us indicator that the forwarder did install through the official approach. 

Let me know what do you think? I can definitely change this back to the version number and document one extra step in our release process to update version in both places. Don't think this is a big deal.

aws/logs_monitoring/template.yaml
```diff
@@ -0,0 +1,457 @@
...
+    Type: String
+    Default: ''
+    Description: Use the supplied regular pattern to detect for a new log line for multiline logs from S3
+  DdSkipSslValidation:
+    Type: String
+    Default: false
+    AllowedValues:
+      - true
+      - false
+    Description: Send logs over HTTPS, while NOT validate the certificate provided by the endpoint
```
DarcyRaynerDD: typo: while NOT validating the certificate

aws/logs_monitoring/template.yaml
```diff
@@ -0,0 +1,457 @@
...
+    Default: ''
+    Description: DO NOT send logs matching the supplied regular expression
+  IncludeAtMatch:
+    Type: String
+    Default: ''
+    Description: Only send logs matching the supplied regular expression and not excluded by EXCLUDE_AT_MATCH
+  DdMultilineLogRegexPattern:
+    Type: String
+    Default: ''
+    Description: Use the supplied regular pattern to detect for a new log line for multiline logs from S3
```
DarcyRaynerDD: Maybe use 'expression' instead of 'pattern' to keep the descriptions consistent.

aws/logs_monitoring/template.yaml
```diff
@@ -0,0 +1,457 @@
...
+    Type: Custom::ForwarderZip
+    Properties:
+      ServiceToken: !GetAtt 'ForwarderZipCopier.Arn'
+      DestZipsBucket: !Ref 'ForwarderZipsBucket'
+      SourceZipURL:
+        Fn::Sub:
+          - 'https://github.com/DataDog/datadog-serverless-functions/releases/download/aws-dd-forwarder-${DdForwarderVersion}/aws-dd-forwarder-${DdForwarderVersion}.zip'
+          - { DdForwarderVersion: !FindInMap [Constants, DdForwarder, Version] }
+  ForwarderZipCopier:
+    Type: AWS::Serverless::Function
```
DarcyRaynerDD, line 395: With the ZipCopier, this resource is still in the customers account after they install the forwarder right? Also, just want to check it isn't publicly callable.

DarcyRaynerDD (approved):

aws/logs_monitoring/template.yaml
tianchu: Correct. The function is not publicly callable, I believe CF invokes the function on behalf of the user creates/updates the CF stack.

tianchu: The solution was adapted based on the aws blog post https://aws.amazon.com/blogs/infrastructure-and-automation/deploying-aws-lambda-functions-using-aws-cloudformation-the-portable-way/

tianchu: `aws lambda get-policy` returns nothing on my testing function, so there is not resource-based policy making the function available to everyone.

PR #174, opened 2019-11-05
[Lambda]: Add DD_SKIP_SSL_VALIDATION option to disable hostname validation for HTTP forwarding

gaetan-deputier:
### What does this PR do?

Provide the DD_SKIP_SSL_VALIDATION option to skip hostname validation for the lambda forwarder.

Requests documentation: https://requests.kennethreitz.org//en/latest/user/advanced/#ssl-cert-verification

### Motivation

This option is provided for specific scenarios such as private link where we can't yet guarantee to provide an endpoint under the logs.datadoghq.com domain

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -255,7 +255,7 @@ DD_SOURCE = "ddsource"
 DD_CUSTOM_TAGS = "ddtags"
 DD_SERVICE = "service"
 DD_HOST = "host"
-DD_FORWARDER_VERSION = "2.3.1"
+DD_FORWARDER_VERSION = "2.3.2"
 
 class RetriableException(Exception):
     pass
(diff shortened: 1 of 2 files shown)


tianchu (approved):

NilsBunge (approved):
Looks good to me.

Nit pick: I used to be told that method signature should be changed by adding new elements at the end. I guess that's not a problem but wanted to double check

aws/logs_monitoring/lambda_function.py
```diff
@@ -250,7 +255,7 @@ def compileRegex(rule, pattern):
 DD_CUSTOM_TAGS = "ddtags"
 DD_SERVICE = "service"
 DD_HOST = "host"
-DD_FORWARDER_VERSION = "2.3.0"
+DD_FORWARDER_VERSION = "2.3.1"
```
swfunc: sorry I took 2.3.1 in https://github.com/DataDog/datadog-serverless-functions/pull/175

```suggestion
DD_FORWARDER_VERSION = "2.3.2"
```

gaetan-deputier:
> Looks good to me.
> 
> Nit pick: I used to be told that method signature should be changed by adding new elements at the end. I guess that's not a problem but wanted to double check

Totally true when the method is called multiple times. In this case, I grouped the ssl values together

PR #175, opened 2019-11-06
Only add env:none tag to logs if no env tag is set

swfunc:
### What does this PR do?

We had a customer reach out that the env:none tag was being applied to their logs even when they supplied their own env: in the `DD_TAGS` environment variable. This PR checks whether an env: tag exists in the custom tags and only adds env:none if there is no env tag already set.

I tested this using demo Lambdas and it's stopped adding the env:none tag when log forwarders have an env set in DD_TAGS.


tianchu (approved):
🎉

PR #176, opened 2019-11-12
Export lambda function ARN

billyshambrook:
### What does this PR do?

Export the Lambda Function ARN so it can be referenced in other Cloudformation Stacks. For example, you could reference this function in a serverless framework service using the [serverless-plugin-log-subscription](https://serverless.com/plugins/serverless-plugin-log-subscription/)

### Motivation

Reference the function that has been deployed into each AWS account, using the same name.


tianchu:
@billyshambrook I have incorporated your PR into mine, since I have to rename a few things: https://github.com/DataDog/datadog-serverless-functions/pull/171/files#diff-3f76f158c2bc9ff25737be780caeb370R61

PR #179, opened 2019-12-02
Updating RDS Enhanced Monitoring to use Lambda layer for sending metrics

jvanbrunschot:
### What does this PR do?

Updating the RDS Enhanced monitoring lambda to make use of the Datadog lambda layer and updating it to at least Python3

Not sure if this is the way to go, especially because of the new distribution metric

If you agree with this change, I'll update the docs as well.

Btw; t he current implementation  required read & write permissions to the datadog account using the app & api key, I don't think that was intentional.

### DIFF
--- a/aws/rds_enhanced_monitoring/lambda_function.py
+++ b/aws/rds_enhanced_monitoring/lambda_function.py
@@ -1,29 +1,21 @@
 # Unless explicitly stated otherwise all files in this repository are licensed
 # under the Apache License Version 2.0.
 # This product includes software developed at Datadog (https://www.datadoghq.com/).
-# Copyright 2018 Datadog, Inc.
-
+# Copyright 2019 Datadog, Inc.
+from datadog_lambda.wrapper import datadog_lambda_wrapper
+from datadog_lambda.metric import lambda_stats
 import gzip
 import json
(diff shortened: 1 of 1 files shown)


aws/rds_enhanced_monitoring/lambda_function.py
```diff
@@ -1,29 +1,21 @@
 # Unless explicitly stated otherwise all files in this repository are licensed
 # under the Apache License Version 2.0.
 # This product includes software developed at Datadog (https://www.datadoghq.com/).
-# Copyright 2018 Datadog, Inc.
-
+# Copyright 2019 Datadog, Inc.
+from datadog_lambda.wrapper import datadog_lambda_wrapper
+from datadog_lambda.metric import lambda_stats
```
jvanbrie, line 6: Do we want to add a check to ensure the lambda layer exists like we do in [here](https://github.com/DataDog/datadog-serverless-functions/blob/master/aws/logs_monitoring/lambda_function.py#L56)?

jvanbrunschot: The whole function relies on the layer being there. We can check and early exit.

PR #182, opened 2019-12-13
Don't duplicate functionname tag when triggered by Kinesis

swfunc:
### What does this PR do?

Fixes a bug that would happen when the log forwarder was triggered by Kinesis with CloudWatch logs from multiple Lambda functions and all of the logs would get a `functionname:` tag for each of the functions in the Kinesis event batch. 

This PR makes it so that the logs are tagged with only the functionname of the specific function that created the log line.

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -473,8 +473,7 @@ class DatadogScrubber(object):
 
 def datadog_forwarder(event, context):
     """The actual lambda function entry point"""
-    events = parse(event, context)
-    metrics, logs, traces = split(events)
+    metrics, logs, traces = split(enrich(parse(event, context)))
 
     if DD_FORWARD_LOG:
         forward_logs(filter_logs(map(json.dumps, logs)))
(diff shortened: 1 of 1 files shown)


aws/logs_monitoring/lambda_function.py
```diff
@@ -607,10 +629,12 @@ def split(events):
         elif trace and DD_FORWARD_TRACES:
             traces.append(trace)
         else:
+            add_metadata_to_lambda_log(event)
```
tianchu: I would leave `split` to ONLY split events, and fix the issue in `awslogs_handler`. See my other comment.

aws/logs_monitoring/lambda_function.py
```diff
@@ -822,13 +846,6 @@ def awslogs_handler(event, context, metadata):
                 # Add the arn as a log attribute
                 arn_attributes = {"lambda": {"arn": arn}}
                 aws_attributes = merge_dicts(aws_attributes, arn_attributes)
-                # Add the function name as tag
```
tianchu, line 825: If I understand correctly, `metadata` is supposed to only carry metadata that applies to **all** the log events, and the root issue is using `metadata` to carry event-specific tags, such as `functionname`?  

If so, I would apply the functionname, host and service to the log event right here, instead of adding `add_metadata_to_lambda_log` to the `split`. We can either loop through `logs["logEvents"]` right here and set these attributes, or cache the attributes in something and apply it later in the existing for loop, like `aws_attributes`.

tianchu (approved):
👍

PR #188, opened 2020-01-08
added another check for gzip files

ghost:
### What does this PR do?

We have some log files in our S3 bucket that are gzipped but do not contain the gzip extension. This is also true of files that are sent from CloudWatch to Kinesis to S3. I would appreciate you considering to merge this for these type of files.

### Motivation

We want to invoke the datadog_lambda_collection function solely from an S3 bucket that collects all of our AWS logs.

### Additional Notes

Please ask if you have questions about how we have AWS & Datadog setup.

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -754,7 +754,7 @@ def s3_handler(event, context, metadata):
     data = body.read()
 
     # If the name has a .gz extension, then decompress the data
-    if key[-3:] == ".gz":
+    if key[-3:] == ".gz" or data[:2] == b"\x1f\x8b":
         with gzip.GzipFile(fileobj=BytesIO(data)) as decompress_stream:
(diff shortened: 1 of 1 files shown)


tianchu:
Hi @shicholas, thanks for your PR! Looks good to me! Do you mind adding a comment with a link to the reference of the magic number? Also I assume you have tested this change and it worked in your setup, but can you confirm? Thank you!

ghost:
Here is a link with the `magic header`, the two binary numbers referenced in this PR - http://www.onicos.com/staff/iz/formats/gzip.html

And yes, I can confirm this works in my setup.

ghost:
hi @tianchu, is there anything else you'd like to see from me before deciding to accept/reject this PR? Thank you!

tianchu:
@shicholas sorry the confusion and the late reply, my original ask was to add the link in the code as a comment :) I made the change directly, and thanks for your contribution!

PR #190, opened 2020-01-14
Add support for HTTPS compression

NoisomePossum:
### What does this PR do?

Adds support for compression when sending logs over HTTPS.

### Motivation

Now that the function sends logs over HTTPS by default we should allow users to compress the logs.

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -100,10 +100,24 @@ DD_API_KEY = "<YOUR_DATADOG_API_KEY>"
 DD_FORWARD_LOG = get_bool_env_var("DD_FORWARD_LOG", "true")
 
 ## @param DD_USE_TCP - boolean - optional -default: false
-## Change this value to `true` to send your logs and metrics using the HTTP network client
+## Change this value to `true` to send your logs and metrics using the TCP network client
 ## By default, it uses the HTTP client.
 #
 DD_USE_TCP = get_bool_env_var("DD_USE_TCP", "false")
+
+## @param DD_USE_COMPRESSION - boolean - optional -default: true
+## Only valid when sending logs over HTTP
+## Change this value to `false` to send your logs without any compression applied
+## By default, compression is enabled.
+#
+DD_USE_COMPRESSION = get_bool_env_var("DD_USE_COMPRESSION", "true")
+
+## @param DD_USE_COMPRESSION - integer - optional -default: 6
+## Change this value to set the compression level.
+## Values range from 0 (no compression) to 9 (best compression).
+## By default, compression is set to level 6.
+#
+DD_COMPRESSION_LEVEL = int(os.getenv("DD_COMPRESSION_LEVEL", 6))
 
 ## @param DD_USE_SSL - boolean - optional -default: false
 ## Change this value to `true` to disable SSL
@@ -349,7 +363,10 @@ class DatadogHTTPClient(object):
     """
 
     _POST = "POST"
-    _HEADERS = {"Content-type": "application/json"}
+    if DD_USE_COMPRESSION:
+        _HEADERS = {"Content-type": "application/json", "Content-Encoding": "gzip"}
+    else:
+        _HEADERS = {"Content-type": "application/json"}
 
     def __init__(self, host, port, no_ssl, skip_ssl_validation, api_key, scrubber, timeout=10):
         protocol = "http" if no_ssl else "https"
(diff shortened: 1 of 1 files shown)


aws/logs_monitoring/lambda_function.py
```diff
@@ -370,15 +386,22 @@ def send(self, logs):
         """
         Sends a batch of log, only retry on server and network errors.
         """
+        try:
+            data=self._scrubber.scrub("[{}]".format(",".join(logs)))
+        except ScrubbingException:
+            raise Exception("could not scrub the payload")
+        if DD_USE_COMPRESSION:
+            try:
+                data = compress_logs(str(data), DD_COMPRESSION_LEVEL)
```
ajacquemot: Why do we need to convert the data to a string here ?

ajacquemot: I think we'll recompress the same batch over and over on retries, we could probably do the compression outside of this method to avoid extra computations on network errors.

aws/logs_monitoring/lambda_function.py
```diff
@@ -444,6 +466,18 @@ def batch(self, logs):
         return batches
 
 
+def compress_logs(batch, level):
+    raw_logs = batch
```
gaetan-deputier: You could remove this assignment

aws/logs_monitoring/lambda_function.py
```diff
@@ -444,6 +466,18 @@ def batch(self, logs):
...
+
+    if level < 0:
+        compression_level = 0
+    elif level > 9:
+        compression_level = 9
+    else:
+        compression_level = level
+    compressed_logs = gzip.compress(bytes(raw_logs, 'utf-8'), compression_level)
```
gaetan-deputier: You can return directly (return gzip.compress ....)

aws/logs_monitoring/lambda_function.py
```diff
@@ -100,11 +100,24 @@ def get_bool_env_var(envvar, default):
 DD_FORWARD_LOG = get_bool_env_var("DD_FORWARD_LOG", "true")
 
 ## @param DD_USE_TCP - boolean - optional -default: false
-## Change this value to `true` to send your logs and metrics using the HTTP network client
+## Change this value to `true` to send your logs and metrics using the TCP network client
 ## By default, it uses the HTTP client.
 #
 DD_USE_TCP = get_bool_env_var("DD_USE_TCP", "false")
 
+## @param DD_USE_COMPRESSION - boolean - optional -default: true
```
gaetan-deputier, line 108: Worth mentioning that's only valid for HTTP

aws/logs_monitoring/lambda_function.py
NoisomePossum: I went ahead and removed the try block.

For the casting to string, I was getting some type errors for a while from the `gzip.compress` function when trying to convert `data` to bytes and casting it to a string first seemed to fix it. I just tried again though without it and everything works ok. Must have been some other error I fixed inadvertently. 🤷‍♂

aws/logs_monitoring/lambda_function.py
NoisomePossum: Added here: https://github.com/DataDog/datadog-serverless-functions/pull/190/files#diff-618731ddf1c446ef45d09613cb4b189bR109

aws/logs_monitoring/lambda_function.py
```diff
@@ -444,6 +464,16 @@ def batch(self, logs):
...
 
+def compress_logs(batch, level):
+    if level < 0:
+        compression_level = 0
+    elif level > 9:
+        compression_level = 9
+    else:
+        compression_level = level
+
+    return gzip.compress(bytes(batch, 'utf-8'), compression_level)
```
ajacquemot, line 475: Can this method throw an error ? If so, we probably want to catch it here https://github.com/DataDog/datadog-serverless-functions/pull/190/files#diff-618731ddf1c446ef45d09613cb4b189bR339 to avoid retrying indefinitely

ajacquemot (approved):
PR looks nice, thanks for working on this, I think we could probably have moved the encoding logic in `DatadogClient` to avoid re-encoding the payload for each retry.

gaetan-deputier (approved):

NoisomePossum:
I took a stab at putting the compression in `DatadogClient` but when I do that I get the encoding error I mentioned above : https://a.cl.ly/NQuvpL0k

I think that error gets solved when the compression happens in `DatadogHTTPClient` because the scrubbing function just before it converts the array of logs into a single string with `join`. I’m also wondering if the scrubbing would still work if you do the compression in `DatadogClient` because wouldn't that mean you're trying to scrub text from a string of bytes once you get to that level?

Issue #195, opened 2020-01-26
Fetch API Key from AWS SSM


demolitionmode:
**Describe what happened:**
I have stored my Datadog API key as a SecureString in AWS SSM and would like to have the Datadog Forwarder Lambda function fetch from there.

**Describe what you expected:**
An environment variable (similar to the secrets manager var) that tells the function to fetch the key, and where from.

PR #196, opened 2020-01-26
Add option to fetch API key from AWS SSM

demolitionmode:
### What does this PR do?

- Adds option to fetch API Key from AWS SSM.
- Updates Parameters description to show all the options

### Motivation

I use the parameter store for all other things, I like to be consistent where possible.

### Additional Notes

Related issue: https://github.com/DataDog/datadog-serverless-functions/issues/195

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -89,14 +89,31 @@ def get_env_var(envvar, default, boolean=False):
 ############# PARAMETERS ############
 #####################################
 
-## @param DD_API_KEY - String - required - default: none
+## @param DD_API_KEY - String - conditional - default: none
 ## The Datadog API key associated with your Datadog Account
 ## It can be found here:
 ##
 ##   * Datadog US Site: https://app.datadoghq.com/account/settings#api
 ##   * Datadog EU Site: https://app.datadoghq.eu/account/settings#api
+##
+## Must be set if one of the following is not set: DD_API_KEY_SECRET_ARN, DD_API_KEY_SSM_NAME, DD_KMS_API_KEY
 #
 DD_API_KEY = "<YOUR_DATADOG_API_KEY>"
+
+## @param DD_API_KEY_SECRET_ARN - String - optional - default: none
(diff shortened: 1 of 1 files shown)


aws/logs_monitoring/lambda_function.py
```diff
@@ -220,6 +237,12 @@ def compileRegex(rule, pattern):
     DD_API_KEY = boto3.client("secretsmanager").get_secret_value(
         SecretId=SECRET_ARN
     )["SecretString"]
+elif "DD_API_KEY_SSM_NAME" in os.environ:
```
DarcyRaynerDD, line 240: Could we rename this it DD_SSM_API_KEY, just to keep it consistent with DD_KMS_API_KEY

DarcyRaynerDD:
Hi marshall0705 👋 
Thanks for the PR, this looks super useful. I left one small nitpick, but otherwise we should be good to add this to the next release.

demolitionmode:
Hi @DarcyRaynerDD 
That's good to hear. I welcome all comments, even nitpicking 😄, however I don't seem to be able to see your comment.

DarcyRaynerDD (changes requested):

aws/logs_monitoring/lambda_function.py
DarcyRaynerDD: @marshall0705, I forgot to submit the review, doh

demolitionmode: I had considered naming it that. It was a choice between the convention used for the secrets manager var, and the convention used for the KMS var.
I chose this name due to DD_KMS_API_KEY and DD_API_KEY both actually containing the key (albeit one of them encrypted), whereas DD_API_KEY_SECRET_ARN is a reference to a remote location, as is DD_API_KEY_SSM_NAME.
If you would still like the name changing, I'm fine to do so.

DarcyRaynerDD: That makes sense, I think we can keep your naming convention then. I'll merge this and roll it out into the next release.

DarcyRaynerDD (approved):

PR #205, opened 2020-02-11
Add Lambda tags to logs, traces and all metrics

DarcyRaynerDD:
### What does this PR do?

When DD_FETCH_LAMBDA_TAGS is true and the forwarder has tags:GetResource permissions, enrich logs/traces/metrics with those tags. Previously only enhanced metrics were enriched.

### DIFF
--- a/aws/logs_monitoring/enhanced_lambda_metrics.py
+++ b/aws/logs_monitoring/enhanced_lambda_metrics.py
@@ -420,3 +420,14 @@ def calculate_estimated_cost(billed_duration_ms, memory_allocated):
     gb_seconds = (billed_duration_ms / 1000.0) * (memory_allocated / 1024.0)
 
     return BASE_LAMBDA_INVOCATION_PRICE + gb_seconds * LAMBDA_PRICE_PER_GB_SECOND
+
+def get_enriched_lambda_log_tags(log):
+    '''
+    Retrieves extra tags from lambda, either read from the function arn, or by fetching lambda tags from the function itself.
+    '''
+    log_function_arn = log.get("lambda", {}).get("arn")
+    if not log_function_arn:
+        return []
+    tags_from_arn = parse_lambda_tags_from_arn(log_function_arn)
+    lambda_custom_tags = account_lambda_tags_cache.get(log_function_arn)
+    return tags_from_arn + lambda_custom_tags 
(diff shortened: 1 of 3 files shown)


aws/logs_monitoring/enhanced_lambda_metrics.py
```diff
@@ -420,3 +420,14 @@ def calculate_estimated_cost(billed_duration_ms, memory_allocated):
     gb_seconds = (billed_duration_ms / 1000.0) * (memory_allocated / 1024.0)
 
     return BASE_LAMBDA_INVOCATION_PRICE + gb_seconds * LAMBDA_PRICE_PER_GB_SECOND
+
+def get_enriched_lambda_log_tags(log):
+    '''
+    Retrieves extra tags from lambda, either read from the function arn, or by fetching lambda tags from the function itself.
+    '''
```
swfunc: nit, the other docstrings in this file use double quotes `"""`

swfunc: I wonder if we should just start black formatting this (not in this PR)

aws/logs_monitoring/lambda_function.py
```diff
@@ -650,9 +650,21 @@ def add_metadata_to_lambda_log(event):
...
 
     event[DD_HOST] = lambda_log_arn
-    event[DD_CUSTOM_TAGS] = "{},functionname:{}".format(event[DD_CUSTOM_TAGS], function_name)
     event[DD_SERVICE] = function_name
 
+    tags = ['functionname:{}'.format(function_name)]
+
+
+    # Add any enhanced tags from metadata
+    if DD_ENHANCED_LAMBDA_METRICS:
```
swfunc: nit: thinking about it more I think the name `DD_ENHANCED_LAMBDA_METRICS` doesn't make as much sense as I originally thought, since the variable is really representing whether the enhanced metrics file is present. I think something like `IS_ENHANCED_METRICS_FILE_PRESENT` would be more descriptive but we don't necessarily need to change it in this PR

swfunc (approved):
:+1: looks good, I left a couple nits

aws/logs_monitoring/enhanced_lambda_metrics.py
DarcyRaynerDD: Good idea, I'll add a tech debt ticket for it.

aws/logs_monitoring/lambda_function.py
DarcyRaynerDD: It's a small enough change, I'll add it here.

swfunc: thanks

swfunc (approved):

PR #209, opened 2020-02-20
allow permissions boundary to be set

Klintrup:
### What does this PR do?

Enable permissions boundary to be set when creating policies

### Motivation

A network boundaries is a requirement for some of our AWS accounts

### DIFF
--- a/aws/logs_monitoring/template.yaml
+++ b/aws/logs_monitoring/template.yaml
@@ -135,6 +135,10 @@ Parameters:
     Default: 6
     AllowedValues: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
     Description: Set the compression level from 0 (no compression) to 9 (best compression).
+  PermissionBoundaryArn:
+    Type: String
+    Default: ''
+    Description: ARN for the Permission Boundary Policy
 Conditions:
   SetFunctionName:
     Fn::Not:
@@ -344,6 +348,8 @@ Resources:
               - Ref: AWS::NoValue
       ReservedConcurrentExecutions:
         Ref: ReservedConcurrency
+      PermissionsBoundary:
+        Ref: PermissionBoundaryArn
       Policies:
       - Version: '2012-10-17'
         Statement:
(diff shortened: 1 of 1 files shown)


aws/logs_monitoring/template.yaml
```diff
@@ -344,6 +348,8 @@ Resources:
               - Ref: AWS::NoValue
       ReservedConcurrentExecutions:
         Ref: ReservedConcurrency
+      PermissionsBoundary:
+        Ref: PermissionBoundaryArn
```
tianchu: I think you have to create a condition `SetPermissionBoundaryArn` and use `AWS::NoValue` when user does not set `PermissionBoundaryArn` like [this](https://github.com/DataDog/datadog-serverless-functions/pull/209/files#diff-7463e3e489db7e82f7ac591389236943R333), otherwise launching the stack will fail.

aws/logs_monitoring/template.yaml
```diff
@@ -450,6 +456,8 @@ Resources:
                   send_cfn_resp(event, context, 'SUCCESS')
               finally:
                   timer.cancel()
+      PermissionsBoundary:
+        Ref: PermissionBoundaryArn
```
tianchu: Same here.

tianchu:
@Klintrup thanks for your contribution! I left a small comment. Everything else looked good to me!

aws/logs_monitoring/template.yaml
Klintrup: Thanks for the feedback @tianchu - I have implemented the proposed changes. I also changed parameter name from PermissionBoundary to PermissionsBoundary to be in sync with AWS Naming of the parameter.

aws/logs_monitoring/template.yaml
Klintrup: same here

tianchu (approved):
LGTM 👍 Thanks for your contribution! @Klintrup Will include in the next release!

PR #210, opened 2020-02-20
Reduce s3 handler mem usage

tianchu:
### What does this PR do?

Somehow the original PR was never merged and released: https://github.com/DataDog/datadog-serverless-functions/pull/117. This PR rebased it against the latest master.

### Motivation

Reduce memory usage when streaming large log objects from s3.


tianchu:
@otherchirps @ajacquemot I'm going to put this to a test. Meanwhile, do you mind giving this PR another review? I don't think that much has changed since then, but just in case. Thanks!

tianchu:
I noticed a few issues while testing, likely due to features/code added recently. I have opened a ticket internally for tracking, and will try to make some progress on this later.

Issue #114, opened 2019-05-15
Function does not handle s3 events delivered via SNS.


hauboldj:
**Describe what happened:**
If you need s3 events to flow to multiple lambda functions you need to send the s3 events to an SNS topic and then subscribe each lambda to the SNS topic.

This function just handles the SNS payload as a structured data, not as an s3 event.
**Describe what you expected:**
This function should detect the s3 event inside the SNS payload and handle the object.

**Steps to reproduce the issue:**
Send s3 event to SNS topic. Subscribe lambda to SNS topic.

NilsBunge:
Hello @hauboldj ,

Thanks for submitting this request.

That's indeed something that should be improved to better support data sent through SNS topic.
We are currently improving the function to better support logs sent through Kinesis and we will make sure to do the same for SNS topics.

Do not hesitate to submit a PR if you have ideas how this could be done.

Once again, thanks for helping improving this Lambda function.

tianchu:
Closing the issue, feel free to track the progress at https://github.com/DataDog/datadog-serverless-functions/pull/351

PR #211, opened 2020-02-24
update sns handler to handle s3 events when s3 events arrive vs sns

hauboldj:
### What does this PR do?

This small update allows the function to handle S3 events when they come via SNS. E.g. you have S3 events going to SNS instead of Lambda directly because you need to have multiple Lambda subscribers for the S3 event.

### Motivation

https://github.com/DataDog/datadog-serverless-functions/issues/114

### Additional Notes

This works for me, it could fail is someone was sending an event through SNS that looked a lot like an S3 object or if they wanted to just log the S3 Event itself.

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -984,6 +984,7 @@ def cwevent_handler(event, metadata):
 
 # Handle Sns events
 def sns_handler(event, context, metadata):
+    data = event
     # test if sns payload contains an s3 event
     try:
         event = json.loads(event['Records'][0]['Sns']['Message'])
@@ -994,7 +995,6 @@ def sns_handler(event, context, metadata):
     except json.decoder.JSONDecodeError:
         pass
 
-    data = event
     # Set the source on the log
     metadata[DD_SOURCE] = parse_event_source(event, "sns")
 
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -606,7 +606,7 @@ def parse(event, context):
         elif event_type == "events":
             events = cwevent_handler(event, metadata)
         elif event_type == "sns":
-            events = sns_handler(event, metadata)
+            events = sns_handler(event, context, metadata)
         elif event_type == "kinesis":
             events = kinesis_awslogs_handler(event, context, metadata)
     except Exception as e:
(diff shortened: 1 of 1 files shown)


aws/logs_monitoring/lambda_function.py
```diff
@@ -983,7 +983,17 @@ def cwevent_handler(event, metadata):
 
 
 # Handle Sns events
-def sns_handler(event, metadata):
+def sns_handler(event, context, metadata):
+    # test if sns payload contains an s3 event
+    try:
+        event = json.loads(event['Records'][0]['Sns']['Message'])
```
DarcyRaynerDD, line 990: I think there is a bug here. If the event successfully parses as json, but fails any of the below checks , it will overwrite the original event object, and the rest of the sns_handler function will  be operating on the wrong object. We are currently working on building an integration testing suite to catch regressions for things like this.

hauboldj: Fixed. I moved the `data = event` back to the top of the function, and use `data` in the rest of the function.

DarcyRaynerDD: My other suggestions would be to move this logic into the `for ev in data["Records"]:`, because right now only the first event in the list is being processed. Also, it might make sense to reuse the parse method so any other event types submitted via SNS can be processed. eg

```python
for ev in data["Records"]:
   try:
      nested_event = json.loads(ev['Sns']['Message'])
      if isinstance(event, dict) and "Records" in event and hasattr(event["Records"], '__len__'):
         nested_evs = parse(event, context)
         for ne in nested_evs:
             yield ne
         return
   except json.decoder.JSONDecodeError:
      pass
   yield ev
```
I haven't tested this code

DarcyRaynerDD (commented):
:wave: Hey @hauboldj, thanks for the PR. I left some comments. I'm going to have a discussion with our team here to decide if this is how we want to support this use case.

mpratsch:
Any news on that?

laurrentt:
I was thinking of opening a PR for the same reason. Since there was no activity for last 3-4 months, I would gladly lend a hand. @DarcyRaynerDD what would you recommend doing? Should I open a new PR addressing the suggestions that were not addressed?

DarcyRaynerDD:
@laurrentt That would be great. I'd recommend creating a new PR. We also added our integration testing suite, so adding a sample event payload to the tests/snapshots folder will help prevent any regressions in the future.

tianchu:
Closing this one in favor of https://github.com/DataDog/datadog-serverless-functions/pull/351

PR #216, opened 2020-03-03
[AWS Log Forwarder] Adds script to test installation process 

DarcyRaynerDD:
### What does this PR do?

Tests deployment/installation process of cloud formation template
There were also some formatting changes to the file. It looks like there were some inconsistent whitespaces, which is the reason fo the large diff.

Also adds the ability to set an s3 source for SourceZipUrl.


aws/logs_monitoring/release.sh
```diff
@@ -26,60 +22,101 @@ if [ -z "$2" ]; then
...
+
+if [ -z "$AWS_ACCESS_KEY_ID" ]; then
+    echo 'AWS_ACCESS_KEY_ID not set. Are you using aws-vault?'
+    exit 1
+fi
+
+if [ -z "$AWS_SECRET_ACCESS_KEY" ]; then
+    echo 'AWS_SECRET_ACCESS_KEY not set. Are you using aws-vault?'
+    exit 1
+fi
```
tianchu: Just an idea, can we replace all these lines with `aws sts get-caller-identity`, which should fail and give user an error message if no AWS credentials provided? 

Nits, shall we make the error message more generic about missing aws credentials rather than `aws-vault`?

aws/logs_monitoring/release.sh
```diff
@@ -26,60 +22,101 @@ if [ -z "$2" ]; then
...
+    zip -r aws-dd-forwarder-${VERSION}.zip .
+    hub release create -a aws-dd-forwarder-${VERSION}.zip -m "aws-dd-forwarder-${VERSION}" aws-dd-forwarder-${VERSION}
+    TEMPLATE_URL="https://${BUCKET}.s3.amazonaws.com/aws/forwarder/latest.yaml"
+    FORWARDER_SOURCE_URL="https://github.com/DataDog/datadog-serverless-functions/releases/download/aws-dd-forwarder-${VERSION}/aws-dd-forwarder-${VERSION}.zip'"
+else
+    echo "About to release non-public staging version of forwarder, upload aws-dd-forwarder-${VERSION} to s3, and upload the template.yaml to s3://${BUCKET}/aws/forwarder-staging/${VERSION}.yaml"
+    # Upload to s3 instead of github
+    rm -f aws-dd-forwarder-*.zip
+    zip -r aws-dd-forwarder-${VERSION}.zip .
+    aws s3 cp aws-dd-forwarder-${VERSION}.zip s3://${BUCKET}/aws/forwarder-staging-zip/aws-dd-forwarder-${VERSION}.zip
```
tianchu, line 89: Just want to make sure we update the internal wiki with the new staging testing steps.

tianchu (approved):
👍

aws/logs_monitoring/release.sh
DarcyRaynerDD: Will do when this and the other PR is merged

PR #217, opened 2020-03-05
[AWS Log Forwarder] Add local integration tests with snapshotting

DarcyRaynerDD:
### What does this PR do?

Adds tools for local snapshot integration testing for event payloads.

```bash
./tools/integration_test.sh
```
Or to update event snapshots:
```bash
UPDATE_SNAPSHOTS=true ./tools/integration_test.sh
```


agocs (approved):
Works on my machine

PR #228, opened 2020-03-17
Add aws_account tag to enhanced metrics and Lambda logs

swfunc:
### What does this PR do?
Adds the aws_account tag to enhanced metrics and Lambda logs to match how we tag CloudWatch aws.lambda metrics.

The logs and metrics are currently being tagged with account_id only, which causes issues when trying to pivot from the graphs at the top of the serverless detail page to the Lambda's logs - because the metrics are tagged with aws_account and the logs are not, the `View related logs` link returns no results.

### Checklist

### DIFF
--- a/aws/logs_monitoring/enhanced_lambda_metrics.py
+++ b/aws/logs_monitoring/enhanced_lambda_metrics.py
@@ -341,7 +341,7 @@ def generate_enhanced_lambda_metrics(log, tags_cache):
     return parsed_metrics
 
 
-def parse_lambda_tags_from_arn(arn, aws_account_tag=False):
+def parse_lambda_tags_from_arn(arn):
     """Generate the list of lambda tags based on the data in the arn
 
     Args:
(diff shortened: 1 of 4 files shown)


aws/logs_monitoring/enhanced_lambda_metrics.py
```diff
@@ -341,7 +341,7 @@ def generate_enhanced_lambda_metrics(log, tags_cache):
     return parsed_metrics
 
 
-def parse_lambda_tags_from_arn(arn):
+def parse_lambda_tags_from_arn(arn, aws_account_tag=False):
```
agocs: Are you using `aws_account_tag` for something?

agocs (approved):
😉

aws/logs_monitoring/enhanced_lambda_metrics.py
swfunc: oops no, I meant to remove that, thanks :)

tianchu (approved):

PR #232, opened 2020-03-18
Update "extractResourceId" function

anshumgargdd:
### What does this PR do?
This extends "extractResourceId" function to extract "source" tag from applications that are deployed at Tenant level.

The existing behavior is to discard all resourceId that don't begin with "/subscriptions/" thus it doesn't handle Tenant level resourceId that begin with "/tenants/".

Adds a else if block to handle resourceIds with Tenant level deployemnts(Like Active Directory)

I have kept the source and tag extraction logic similar to what currently exists.

A brief description of the change being made with this pull request.

### Motivation

Currently we don't get the "source" tag for Azure Active Directory thus making it not possible to ship logs integration pipeline and security rules for the integration. 

### Additional Notes

@steveharrington1 for additional context on the Azure integrations side.

### Checklist

### DIFF
--- a/azure/activity_logs_monitoring/index.js
+++ b/azure/activity_logs_monitoring/index.js
@@ -135,22 +135,31 @@ function extractResourceId(record) {
     metadata = { tags: [], source: '' };
     if (
         record.resourceId === undefined ||
-        typeof record.resourceId !== 'string' ||
-        !record.resourceId.toLowerCase().startsWith('/subscriptions/')
+        typeof record.resourceId !== 'string'
     ) {
         return metadata;
     }
-    var resourceId = record.resourceId.toLowerCase().split('/');
-    if (resourceId.length > 2) {
-        metadata.tags.push('subscription_id:' + resourceId[2]);
+    else if(record.resourceId.toLowerCase().startsWith('/subscriptions/')){
+      var resourceId = record.resourceId.toLowerCase().split('/');
+      if (resourceId.length > 2) {
+          metadata.tags.push('subscription_id:' + resourceId[2]);
+      }
+      if (resourceId.length > 4) {
+          metadata.tags.push('resource_group:' + resourceId[4]);
+      }
+      if (resourceId.length > 6 && resourceId[6]) {
+          metadata.source = resourceId[6].replace('microsoft.', 'azure.');
+      }
+      return metadata;
     }
-    if (resourceId.length > 4) {
-        metadata.tags.push('resource_group:' + resourceId[4]);
+    else if(record.resourceId.toLowerCase().startsWith('/tenants/')){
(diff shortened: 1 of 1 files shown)


azure/activity_logs_monitoring/index.js
```diff
@@ -135,22 +135,31 @@ function extractResourceId(record) {
...
+      }
+      return metadata;
     }
-    if (resourceId.length > 4) {
-        metadata.tags.push('resource_group:' + resourceId[4]);
+    else if(record.resourceId.toLowerCase().startsWith('/tenants/')){
+      var resourceId = record.resourceId.toLowerCase().split('/');
+      if (resourceId.length > 2 && resourceId[4]) {
+          metadata.tags.push('tenant:' + resourceId[2]);
+          metadata.source = resourceId[4].replace('microsoft.', 'azure.').replace('aadiam', 'active_directory');
```
steveharrington1: @anshumgargdd we'll want to update this to remove the '_' like we talked about. This will make it match the format of all the other azure source.

Updatted format should be: 'source:azure.activedirectory'

anshumgargdd: Done.

azure/activity_logs_monitoring/index.js
```diff
@@ -135,22 +135,31 @@ function extractResourceId(record) {
...
+      if (resourceId.length > 6 && resourceId[6]) {
+          metadata.source = resourceId[6].replace('microsoft.', 'azure.');
+      }
+      return metadata;
     }
-    if (resourceId.length > 4) {
-        metadata.tags.push('resource_group:' + resourceId[4]);
+    else if(record.resourceId.toLowerCase().startsWith('/tenants/')){
+      var resourceId = record.resourceId.toLowerCase().split('/');
+      if (resourceId.length > 2 && resourceId[4]) {
```
graybjones: Is there any chance the resourceId.length could be > 2 but less than 5?

anshumgargdd: Unfortunately we don't have an exhaustive list of resourceIds so I am unsure. However in that case we just won't process the resourceId to extract source and assign 'azure' as the default source which is the behavior right now.

graybjones: I think line 157 should read:

`if (resourceId.length > 4 && resourceId[4]) {`

That would cover the situation where the resourceId is greater than 2, but less than 5.

graybjones (changes requested):
A small change at line 157 mentioned in the existing conversation. Thanks.

azure/activity_logs_monitoring/index.js
anshumgargdd: Fixed.

graybjones (approved):
Thanks!

PR #234, opened 2020-03-19
Exclude unuseful files from zip

tianchu:
### What does this PR do?

Exclude unuseful files from forwarder release zip.

### Checklist

### DIFF
--- a/aws/logs_monitoring/release.sh
+++ b/aws/logs_monitoring/release.sh
@@ -77,7 +77,11 @@ if [ "$PROD_RELEASE" = true ] ; then
     echo "Release aws-dd-forwarder-${VERSION} to github"
     go get github.com/github/hub
     rm -f aws-dd-forwarder-*.zip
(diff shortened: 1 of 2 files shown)


aws/logs_monitoring/release.sh
```diff
@@ -77,15 +77,23 @@ if [ "$PROD_RELEASE" = true ] ; then
     echo "Release aws-dd-forwarder-${VERSION} to github"
     go get github.com/github/hub
     rm -f aws-dd-forwarder-*.zip
-    zip -r aws-dd-forwarder-${VERSION}.zip .
+    zip -r aws-dd-forwarder-${VERSION}.zip . \
+        --exclude=*tools* \
```
DarcyRaynerDD, line 81: This is will work well enough, but would it make sense to whitelist files instead?

DarcyRaynerDD (approved):

PR #236, opened 2020-03-25
fix sanitize tags rule in order to accept hyphens characters

asaghri:
### What does this PR do?

This PR will just accept tags with hyphen instead of replacing them

### Motivation

Currently, the hyphen character in tags are replaced by an underscore. Which cause many issues on our side, because the ec2 hosts accept the hyphen so there is no more correlation between our lambda logs and our ec2 logs.

The datadog support gave me the following answer 

> Essentially, in the 
enhanced_lambda_metrics.py script we have some logic to convert some illegal characters to underscores and hyphens have been incorrectly added to this rule. Our legal characters are listed in our documentation[1] and hold true for tags from Lambda also with the exception of colons. I have since flagged this with our Product & Engineering team and they are working on fixing this rule as we speak. 

### Checklist

### DIFF
--- a/aws/logs_monitoring/enhanced_lambda_metrics.py
+++ b/aws/logs_monitoring/enhanced_lambda_metrics.py
@@ -174,7 +174,7 @@ def should_fetch_custom_tags():
 def sanitize_aws_tag_string(raw_string):
     """Convert characters banned from DD but allowed in AWS tags to underscores
     """
-    return re.sub(r"[\+\-=\.:/@]", "_", raw_string)
+    return re.sub(r"[\+=\.:/@]", "_", raw_string)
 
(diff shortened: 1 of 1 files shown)


DarcyRaynerDD:
Hi @asaghri, thanks for the PR.
We are going to hold off merging for now. We're aware of this issue and actually have an internal ticket for this queued up for work. There are a few more edge cases around the tag parsing we need to address and we would rather apply all those fixes/breaking changes in a single release. We will update here with any progress.

tianchu:
I believe https://github.com/DataDog/datadog-serverless-functions/pull/240 fixed the issue. But feel free to reopen if that's not the case.

PR #240, opened 2020-03-31
Fix enhanced metric tag conversion

DarcyRaynerDD:
### What does this PR do?

Modify tag parsing rules to match Datadog's internal integrations
1. Convert camel case to underscores
2. Replaces colons with _
3. Convert to all lowercase unicode string
4. Convert bad characters to underscores
5. Dedupe contiguous underscores
6. Remove initial underscores/digits such that the string
7. Truncate to 200 characters
8. Strip trailing underscores

### Motivation

What inspired you to submit this pull request?

### Additional Notes

### Checklist

### DIFF
--- a/aws/logs_monitoring/enhanced_lambda_metrics.py
+++ b/aws/logs_monitoring/enhanced_lambda_metrics.py
@@ -188,7 +188,7 @@ def convert_camel_case_to_underscore(string):
     return AllCapRe.sub(r"\1_\2", string).lower()
 
 
-def sanitize_aws_tag_string(tag):
+def sanitize_aws_tag_string(tag, remove_colons=False):
     """Convert characters banned from DD but allowed in AWS tags to underscores
     """
     global Sanitize, Dedupe, FixInit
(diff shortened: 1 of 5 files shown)


aws/logs_monitoring/enhanced_lambda_metrics.py
```diff
@@ -171,10 +171,48 @@ def should_fetch_custom_tags():
...
+    # 5. Dedupe contiguous underscores
+    # 6. Remove initial underscores/digits such that the string
+    #    starts with an alpha char
+    #    FIXME: tag normalization incorrectly supports tags starting
+    #    with a ':', but this behavior should be phased out in future
+    #    as it results in unqueryable data.  See dogweb/#11193
+    # 7. Truncate to 200 characters
+    # 8. Strip trailing underscores
+    tag = convert_camel_case_to_underscore(tag)
+    tag = tag.replace(":", "_")
```
swfunc: I think we don't want to replace every colon with an underscore since we don't do that in crawlers. If I'm reading the code [here](https://github.com/DataDog/dogweb/blob/prod/integration/amazon_web_services/utils/common.py#L106-L128) right, the crawlers never replace the colons in tag values but do replace the colons in tag keys if the `aws_tags_replace_colons` FF is enabled, which it is by default for new orgs. Since it's the default for new orgs and we can't check the FF from the forwarder I think it would make sense to make that the behavior here and maybe at some point add an env var that lets customers control the behavior like the FF if customers ask for it.

DarcyRaynerDD: Hmm, good point, I'll try and update the logic.

swfunc (approved):

PR #243, opened 2020-04-03
Enable supplying DdApiKeySecret directly

jim80net:
### What does this PR do?

Works around indefinite reapply changes when using terraform to deploy
cloudformation with NoEcho parameters
(https://github.com/terraform-providers/terraform-provider-aws/issues/55)

### Motivation

Using terraform with DdApiKey (and NoEcho) results in always requiring reapplication of the API key parameter.

### Checklist

### DIFF
--- a/aws/logs_monitoring/template.yaml
+++ b/aws/logs_monitoring/template.yaml
@@ -8,10 +8,14 @@ Mappings:
 Parameters:
   DdApiKey:
     Type: String
+    Default: ""
     NoEcho: true
-    AllowedPattern: .+
-    ConstraintDescription: DdApiKey is required
-    Description: The Datadog API key, which can be found from the APIs page (/account/settings#api). It will be stored in AWS Secrets Manager securely.
+    Description: The Datadog API key, which can be found from the APIs page (/account/settings#api). It will be stored in AWS Secrets Manager securely. Either this or DdApiKeySecret must be set.
+  DdApiKeySecret:
+    Type: String
+    AllowedPattern: "arn:aws:secretsmanager:.*"
+    Default: "arn:aws:secretsmanager:DEFAULT"
+    Description: The ARN where Datadog API key can be found. Either this or DdApiKey must be set.
   DdSite:
     Type: String
     Default: datadoghq.com
@@ -142,6 +146,11 @@ Parameters:
     Default: ""
     Description: ARN for the Permissions Boundary Policy
 Conditions:
+  SetDdApiKeySecretFromApiKey:
+    Fn::Not:
+      - Fn::Equals:
+         - Ref: DdApiKey
+         - ""
   SetFunctionName:
     Fn::Not:
       - Fn::Equals:
@@ -273,7 +282,10 @@ Resources:
         Variables:
           DD_ENHANCED_METRICS: "false"
           DD_API_KEY_SECRET_ARN:
-            Ref: DdApiKeySecret
+            Fn::If:
+              - SetDdApiKeySecretFromApiKey
+              - Ref: DdApiKeySecretFromApiKey
+              - Ref: DdApiKeySecret
           DD_SITE:
             Ref: DdSite
           DD_TAGS:
@@ -379,7 +391,10 @@ Resources:
               Action:
                 - secretsmanager:GetSecretValue
               Resource:
-                Ref: DdApiKeySecret
+                Fn::If:
+                  - SetDdApiKeySecretFromApiKey
+                  - Ref: DdApiKeySecretFromApiKey
+                  - Ref: DdApiKeySecret
             - !If
               - SetDdFetchLambdaTags
               - Effect: Allow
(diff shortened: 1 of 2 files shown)


aws/logs_monitoring/template.yaml
```diff
@@ -8,10 +8,14 @@ Mappings:
 Parameters:
   DdApiKey:
     Type: String
+    Default: ""
     NoEcho: true
-    AllowedPattern: .+
```
tianchu, line 12: We actually just changed this to disallow leaving DdApiKey to be empty https://github.com/DataDog/datadog-serverless-functions/pull/242/files#. I guess not everyone read descriptions and many users actually forgot to set api key, and they only receive an error down the line when CloudFormation stack tries to create the secret. We added the AllowedPattern to give user an immediate feedback before the stack gets launched.

I understands why you need to make this change, but it conflicts with another need, and unfortunately CloudFormation doesn't provide a way out here.

Since `DdApiKeySecret` is more like an advanced feature, I would like to NOT sacrifice the experience for average users. Perhaps an ugly solution is documenting in the `DdApiKeySecret`'s description that you do have to set a **fake** `DdApiKey` as a placeholder? That's the best I can think of. Ideas are welcome!

tianchu: BTW, could `ignore_changes` be a workaround? So you don't need to re-set api key when updating? https://www.terraform.io/docs/configuration/resources.html#ignore_changes

aws/logs_monitoring/template.yaml
```diff
@@ -8,10 +8,14 @@ Mappings:
...
-    ConstraintDescription: DdApiKey is required
-    Description: The Datadog API key, which can be found from the APIs page (/account/settings#api). It will be stored in AWS Secrets Manager securely.
+    Description: The Datadog API key, which can be found from the APIs page (/account/settings#api). It will be stored in AWS Secrets Manager securely. Either this or DdApiKeySecret must be set.
+  DdApiKeySecret:
```
tianchu: Do you mind renaming it to `DdApiKeySecretArn` for explicity?

aws/logs_monitoring/template.yaml
```diff
@@ -8,10 +8,14 @@ Mappings:
...
+    Type: String
+    AllowedPattern: "arn:aws:secretsmanager:.*"
+    Default: "arn:aws:secretsmanager:DEFAULT"
+    Description: The ARN where Datadog API key can be found. Either this or DdApiKey must be set.
```
tianchu: Do you mind updating the description to read "The ARN of the secret storing the Datadog API key, if you already have it stored in Secretes Manager. You still need to set a dummy value for "DdApiKey" to satisfy the requirement, though that value won't be used."

aws/logs_monitoring/template.yaml
tianchu: Can you group this variable to `Advanced` https://github.com/DataDog/datadog-serverless-functions/pull/243/files#diff-7463e3e489db7e82f7ac591389236943R612, so it only show up at the bottom of the CFN stack launch interface.

aws/logs_monitoring/template.yaml
```diff
@@ -142,6 +146,11 @@ Parameters:
     Default: ""
     Description: ARN for the Permissions Boundary Policy
 Conditions:
+  SetDdApiKeySecretFromApiKey:
```
tianchu: Can you rename this condition to `CreateDdApiKeySecret`?

aws/logs_monitoring/template.yaml
```diff
@@ -142,6 +146,11 @@ Parameters:
...
+    Fn::Not:
+      - Fn::Equals:
+         - Ref: DdApiKey
```
tianchu: Related to my previous comment, probably need to update this to compare DdApiKeySecretArn's value with its default value to determine the condition result.

tianchu (commented):
@jim80net Thanks for your contribution 🙇! Looks good to me overall, except for a few comments.

aws/logs_monitoring/template.yaml
jim80net: Sounds good! I have incorporated the "ugly solution." I couldn't think of a better idea. With the latest commit, we preserve the mandatory DdApiKey feature, and still enable supplying a DdApiKeySecretArn to override the behavior.

aws/logs_monitoring/template.yaml
jim80net: Done!

aws/logs_monitoring/template.yaml
jim80net: Done!

aws/logs_monitoring/template.yaml
jim80net: Done!

aws/logs_monitoring/template.yaml
jim80net: Done!

jim80net:
As I indicated in comments in-line, I've adopted the proposed solution. Please let me know how it looks. Thank you very much!

tianchu (approved):
👍 LGTM! Tested, everything looks fine. Again, thanks for your contribution! @jim80net

PR #244, opened 2020-04-06
Ensure empty tags can be sanitized

william-richard:
### What does this PR do?

Adds support for lambda tags with empty values, with tests.

### Motivation

We have a few lambdas that have tags with empty values, and that was causing an exception
to be thrown during the sanitization function.  

### Additional Notes

None

### Checklist


tianchu (approved):

tianchu:
Thanks for fixing this @william-richard 🙇

PR #249, opened 2020-04-13
Refresh tags when new lambda arn is encountered

DarcyRaynerDD:
### What does this PR do?

Refreshes tags in the tag cache when a new lambda arn is encountered

### Motivation

Some issues were being encountered with new lambdas taking a while to have there tags attached

### Additional Notes

### Checklist

### DIFF
--- a/aws/logs_monitoring/enhanced_lambda_metrics.py
+++ b/aws/logs_monitoring/enhanced_lambda_metrics.py
@@ -197,7 +197,7 @@ FixInit = re.compile(r"^[_\d]*", re.UNICODE).sub
 def sanitize_aws_tag_string(tag, remove_colons=False):
     """Convert characters banned from DD but allowed in AWS tags to underscores
     """
-    global Sanitize, Dedupe, FixInit1677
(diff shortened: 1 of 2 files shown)


aws/logs_monitoring/enhanced_lambda_metrics.py
```diff
@@ -180,7 +197,7 @@ def should_fetch_custom_tags():
 def sanitize_aws_tag_string(tag, remove_colons=False):
     """Convert characters banned from DD but allowed in AWS tags to underscores
     """
-    global Sanitize, Dedupe, FixInit
+    global Sanitize, Dedupe, FixInit1677
```
swfunc: typo?

swfunc (commented):
Looks good, nice tests. I think I found a typo

aws/logs_monitoring/enhanced_lambda_metrics.py
DarcyRaynerDD: Whoops, good catch

PR #251, opened 2020-04-14
Terraform code needs to ignore changes to parameters

jikamens:
### What does this PR do?

Prevents Terraform from trying to change the AWS CloudFormation stack object every time `terraform apply` is run.

### Motivation

Because the `DdApiKey` parameter is replaced upon deployment when using Secrets Manager to store the Datadog API key, changes to the value of `parameters` need to be ignored in the resource lifecycle of the `aws_cloudformation_stack` object when deploying via Terraform.

### Checklist

### DIFF
--- a/aws/logs_monitoring/README.md
+++ b/aws/logs_monitoring/README.md
@@ -118,6 +118,9 @@ resource "aws_cloudformation_stack" "datadog-forwarder" {
     DdApiKeySecretArn  = aws_secretsmanager_secret.dd_api_key.arn
     FunctionName       = "datadog-forwarder"
   }
+  lifecycle {
+    ignore_changes = [parameters]
+  }
   template_url = "https://datadog-cloudformation-template.s3.amazonaws.com/aws/forwarder/latest.yaml"
 }
 ```


tianchu:
Looks good to me!

BTW, we now also expose the `DdApiKeySecretArn` parameter so you can create the secret yourself and supply the ARN instead, which won't change in every deploy. https://github.com/DataDog/datadog-serverless-functions/pull/243

I'm not a Terraform expert, does `ignore_changes` come with any downsides? Maybe, we should add a comment to mention that you only need one of these solutions?

jikamens:
>I'm not a Terraform expert, does ignore_changes come with any downsides? Maybe, we should add a comment to mention that you only need one of these solutions?

The risk is that if someone uses the AWS web console or CLI to modify the CloudFormation stack's parameters directly, i.e., without editing them in Terraform, then the next time someone applies their Terraform project, those changes could get overwritten. It's a low but non-zero risk. But the alternative -- not doing this -- is that Terraform will _definitely_ screw up the CloudFormation parameters and/or force the CloudFormation stack to be redeployed every time Terraform is run, as long as the functionality for automatically generating the secret is used. So that seems like a much worse scenario.

I suppose we could put in a comment saying that if the ARN is specified explicitly then the `lifecycle` stanza isn't necessary, but :shrug: I'm not sure it's worth the increased clutter. Up to you, I guess.

tianchu:
Hmm, doesn't `ignore_changes` ignore changes to other parameters (the ones other than the api key) too? 

After a second thought, I'd like to keep recommending the current template, since many customers already have a datadog api key in Secrets Manager for other use cases, and using the ARN rather than the actual API key helps avoid exposing the secret through command line history or the Terraform backend.

Certainly, this doesn't prevent you from adopting the `ingore_changes` approach, and thanks for submitting the PR! 🙇 But I'm going to close it for now, until we get additional feedbacks.

PR #259, opened 2020-04-30
Store lowercase ARN keys in tags cache

swfunc:
### What does this PR do?

Fixes an issue with the `DdFetchCustomTags` feature where in some cases custom tags were not being applied to metrics or logs. The problem was that the cache keys were in the original ARN casing but the ARN attribute that is given to the logs in post processing is in lowercase.

### Checklist


DarcyRaynerDD (approved):
Nice catch, LGTM

PR #274, opened 2020-06-04
Jordan.storms/add init duration metric and cold_start tags

jcstorms1:
### What does this PR do?

Adds an `init duration` enhanced metric for cold starts
Adds a `cold_start` tag for all metrics

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/enhanced_lambda_metrics.py
+++ b/aws/logs_monitoring/enhanced_lambda_metrics.py
@@ -40,18 +40,7 @@ REPORT_LOG_REGEX = re.compile(
     )
     + r"Memory\s+Size:\s+(?P<{}>\d+)\s+MB\s+".format(MEMORY_ALLOCATED_FIELD_NAME)
     + r"Max\s+Memory\s+Used:\s+(?P<{}>\d+)\s+MB".format(MAX_MEMORY_USED_METRIC_NAME)
-)
-# Make separate regex to account for cold start
-REPORT_LOG_REGEX_INIT = re.compile(
-    r"REPORT\s+"
-    + r"RequestId:\s+(?P<{}>[\w-]+)\s+".format(REQUEST_ID_FIELD_NAME)
-    + r"Duration:\s+(?P<{}>[\d\.]+)\s+ms\s+".format(DURATION_METRIC_NAME)
-    + r"Billed\s+Duration:\s+(?P<{}>[\d\.]+)\s+ms\s+".format(
-        BILLED_DURATION_METRIC_NAME
-    )
-    + r"Memory\s+Size:\s+(?P<{}>\d+)\s+MB\s+".format(MEMORY_ALLOCATED_FIELD_NAME)
-    + r"Max\s+Memory\s+Used:\s+(?P<{}>\d+)\s+MB\s+".format(MAX_MEMORY_USED_METRIC_NAME)
-    + r"Init\s+Duration:\s+(?P<{}>[\d\.]+)\s+ms".format(INIT_DURATION_METRIC_NAME)
+    + r"(\s+Init\s+Duration:\s+(?P<{}>[\d\.]+)\s+ms)?".format(INIT_DURATION_METRIC_NAME)
 )
 
 METRICS_TO_PARSE_FROM_REPORT = [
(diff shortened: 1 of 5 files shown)


tianchu:
Can you add a test case for the new metric and tag https://github.com/DataDog/datadog-serverless-functions/blob/master/aws/logs_monitoring/tests/test_enhanced_lambda_metrics.py?

aws/logs_monitoring/enhanced_lambda_metrics.py
```diff
@@ -40,11 +41,18 @@
     + r"Memory\s+Size:\s+(?P<{}>\d+)\s+MB\s+".format(MEMORY_ALLOCATED_FIELD_NAME)
     + r"Max\s+Memory\s+Used:\s+(?P<{}>\d+)\s+MB".format(MAX_MEMORY_USED_METRIC_NAME)
 )
-
-# Pull memorysize tag and cold start from report
-TAGS_TO_PARSE_FROM_REPORT = [
-    MEMORY_ALLOCATED_FIELD_NAME,
-]
+# Make separate regex to account for cold start
+REPORT_LOG_REGEX_INIT = re.compile(
```
tianchu: Any reason why we can't combine this with `REPORT_LOG_REGEX` by marking "Init Duration" being optional?

jcstorms1: No. I'll update it since that's a better solution.

aws/logs_monitoring/enhanced_lambda_metrics.py
```diff
@@ -439,17 +437,34 @@ def parse_metrics_from_report_log(report_log_line):
     Returns:
         metrics - DatadogMetricPoint[]
     """
+    metrics = []
+    tags = ["cold_start:false"]
```
tianchu: Probably a nits, why do you need to front load these two lines before the regex match? The original sequence reflects better to me, because you first check whether the log line is the REPORT log, if not return []...

aws/logs_monitoring/enhanced_lambda_metrics.py
```diff
@@ -439,17 +437,34 @@ def parse_metrics_from_report_log(report_log_line):
...
     if not regex_match:
         return []
 
-    metrics = []
-    tags = []
+    # we always want coldstart and memorysize tags
+    tags.append("memorysize:" + regex_match.group(MEMORY_ALLOCATED_FIELD_NAME))
+
+    # if cold_start:
+    if regex_match.group(INIT_DURATION_METRIC_NAME):
```
tianchu, line 455: Nits, do you mind keeping the tags and metrics processing in separate sections, like the original code does? processing the tags first, then working on metrics?

I would leave the if..else... to generate the cold_start tag here, and move the logic to generate the init_duration metric to the existing for loop `For for metric_name in METRICS_TO_PARSE_FROM_REPORT` (probably you need to add a if statement inside the loop to handle the case when there is no regex match since init duration is not always present).

tianchu (approved):
👍  LGTM! One last thing, do you mind check all the boxes that apply from the PR template?

aws/logs_monitoring/enhanced_lambda_metrics.py
```diff
@@ -439,17 +437,36 @@ def parse_metrics_from_report_log(report_log_line):
...
+        initial_duration = DatadogMetricPoint(
+            "{}.{}".format(
+                ENHANCED_METRICS_NAMESPACE_PREFIX, INIT_DURATION_METRIC_NAME
+            ),
+            metric_point_value,
+        )
+
+        initial_duration.add_tags(tags)
+
+        metrics.append(initial_duration)
```
tianchu, line 469: @jcstorms1 my apologies, after a weekend, I totally forgot that another suggestion I pointed out last week was merging line 454-469 (after change) to the for loop of line 471 (after change), their logic are 99% same. I suggest adding `INIT_DURATION_METRIC_NAME` to `METRICS_TO_PARSE_FROM_REPORT` and add a check within the for loop like this

```python
for metric_name in METRICS_TO_PARSE_FROM_REPORT:
    if not regex_match.group(metric_name):
        continue
    metric_point_value = float(regex_match.group(metric_name))
    ....
```

PR #292, opened 2020-07-06
[azure event hub forwarder] Convert strings to json logs

claudiadadamo:
### What does this PR do?
Updates the azure event hub forwarder to handle string logs and convert them to json if possible. Also updates unit tests, and adds `sinon` to the requirements so we can use spies to mock certain behavior. Fixes a few things and ran the linter on the code.

### Motivation
https://datadoghq.atlassian.net/jira/software/projects/AGAI/boards/207?selectedIssue=AGAI-414

### Testing Guidelines

Copied this code into the azure portal for a function I created, set the type to string and confirmed logs come into the test org I'm using correctly. Also ran unittests to confirm the correct format of the logs are passed to the right method

Before: 
image

After:
image

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/azure/test/client.test.js
+++ b/azure/test/client.test.js
@@ -16,10 +16,6 @@ describe('Azure Log Monitoring', function() {
       eventHubMessages = ['', 'foobar'];
       assert.equal(constants.STRING_ARRAY, client.getLogFormat(eventHubMessages));
     });
-//    it('should return json with records', function() {
-//      eventHubMessages = [{'records': [{}, {}]}, {'records': [{}, {}]}];
-//      assert.equal(constants.JSON_RECORDS, client.getLogFormat(eventHubMessages));
-//    });
     it('should return json object', function() {
       eventHubMessages = {'key': 'value', 'otherkey':'othervalue'};
       assert.equal(constants.JSON_OBJECT, client.getLogFormat(eventHubMessages));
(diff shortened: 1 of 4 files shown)


azure/test/client.test.js
```diff
@@ -78,4 +75,101 @@ describe('Azure Log Monitoring', function() {
...
+    it('should handle json-object properly', function() {
+      record = {'hello': 'there'}
+      expected = {'hello': 'there'}
+      assert.equal(client.getLogFormat(record), constants.JSON_OBJECT)
+      testHandleLogs(record, expected, true)
+    });
+
+    it('should handle string-array properly', function() {
+      record = ['one message', 'two message']
+      expected = 'one message'
```
garnermccloud: why do we only expect one record to be returned here for these tests with multiple records?

garnermccloud (commented):
Looks good overall - just one question on the test cases

azure/test/client.test.js
claudiadadamo: It's just testing that the mock is called with the expected value, and it will be called once with each message. I can make this more clear/test that it is called once per message

garnermccloud (approved):

PR #299, opened 2020-07-13
Refactor parse_event_source to prevent mis-identification of sources

swfunc:
### What does this PR do?
Refactors the logic in parse_event_source to prevent bugs where the event's key (s3 bucket or CW log group) has a substring of an AWS service (for example, `waf` or `sns`), causing the function to mis-identify the log's source as the substring service.

Also adds some comments and unit tests for the function.

### Motivation
Received customer ticket SLES-225

### Testing Guidelines
Added unit tests, ran integration tests

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -65,7 +65,7 @@ def get_env_var(envvar, default, boolean=False):
 ############# PARAMETERS ############
 #####################################
 
-## @param DD_API_KEY - String - conditional - default: none:
+## @param DD_API_KEY - String - conditional - default: none
 ## The Datadog API key associated with your Datadog Account
 ## It can be found here:
 ##
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -65,7 +65,7 @@ def get_env_var(envvar, default, boolean=False):
 ############# PARAMETERS ############
 #####################################
 
-## @param DD_API_KEY - String - conditional - default: none
+## @param DD_API_KEY - String - conditional - default: none:
 ## The Datadog API key associated with your Datadog Account
 ## It can be found here:
 ##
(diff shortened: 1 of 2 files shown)


aws/logs_monitoring/lambda_function.py
```diff
@@ -999,7 +1019,7 @@ def cwevent_handler(event, metadata):
 def sns_handler(event, metadata):
     data = event
     # Set the source on the log
-    metadata[DD_SOURCE] = parse_event_source(event, "sns")
+    metadata[DD_SOURCE] = "sns"
```
swfunc, line 1022: No need to pass the event to parse_event_source since we already know this is SNS

aws/logs_monitoring/lambda_function.py
```diff
@@ -307,6 +307,26 @@ def compileRegex(rule, pattern):
 DD_FORWARDER_VERSION = "3.15.0"
 
 
+# Used to identify and assign sources to logs
+LOG_SOURCE_SUBSTRINGS = [
```
agocs, line 311: I don't have evidence to support this at the moment, but it feels like all of these would probably be preceded by a `/` or a `.` of some sort. If that hunch is correct, would it be worth including those in the substrings in order to make similar errors less likely?

agocs (approved):

aws/logs_monitoring/lambda_function.py
swfunc: Good point, I thought the same thing but I didn't have time to find examples of all the event types to confirm so I left it for now to prevent introducing any new bugs. But I do agree we should refactor this to be more thorough, I just made this card in triage https://datadoghq.atlassian.net/jira/software/projects/SLS/boards/205?selectedIssue=SLS-617

PR #310, opened 2020-07-28
Translate binary data type to be human readable

Ricky-Thomas:
### What does this PR do?

We’d like to allow customers to be able to send logs as binary format to the datadog forwarder. We should handle the binary type (example of where it is set is here: https://a.cl.ly/E0uzNmg0)

Here is what this type of log would look like in DD prior to this change: https://a.cl.ly/yAub7zbb

Here is the same logs after the change: https://a.cl.ly/z8uZEoPv

### Motivation

https://datadoghq.atlassian.net/jira/software/projects/AGAI/boards/207?selectedIssue=AGAI-417

### Testing Guidelines

Tested in the Azure portal and locally with unit tests.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/azure/activity_logs_monitoring/index.js
+++ b/azure/activity_logs_monitoring/index.js
@@ -102,6 +102,10 @@ function handleJSONArrayLogs(sender, context, logs, logsType) {
                 return;
             }
         }
+        // If the message is a buffer object, the data type has been set to binary.
+        if (Buffer.isBuffer(message)) {
+            message = JSON.parse(message.toString());
+        }
         if (message.records != undefined) {
             message.records.forEach(sender(addTagsToJsonLog));
         } else {
--- a/azure/activity_logs_monitoring/index.js
+++ b/azure/activity_logs_monitoring/index.js
@@ -11,6 +11,7 @@ const STRING = 'string'; // example: 'some message'
 const STRING_ARRAY = 'string-array'; // example: ['one message', 'two message', ...]
 const JSON_OBJECT = 'json-object'; // example: {"key": "value"}
 const JSON_ARRAY = 'json-array'; // example: [{"key": "value"}, {"key": "value"}, ...] or [{"records": [{}, {}, ...]}, {"records": [{}, {}, ...]}, ...]
+const BUFFER_ARRAY = 'buffer-array'; // example: <Buffer 7b 22 72 65 63 6f 72 64 73 22 3a 20 5b 7b 20 22 74 69 6d 65 22 3a 20 22 74 65 73 74 69 6e 67 22 7d 5d 7d>
 const JSON_STRING = 'json-string'; // example: '{"key": "value"}'
 const JSON_STRING_ARRAY = 'json-string-array'; // example: ['{"records": [{}, {}]}'] or ['{"key": "value"}']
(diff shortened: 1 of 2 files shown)


azure/activity_logs_monitoring/index.js
```diff
@@ -102,6 +102,10 @@ function handleJSONArrayLogs(sender, context, logs, logsType) {
                 return;
             }
         }
+        // If the message is a buffer object, the data type has been set to binary.
+        if (Buffer.isBuffer(message)) {
+            message = JSON.parse(message.toString());
+        }
```
claudiadadamo: I think it could be confusing that this is considered json-array when it's not actually a json array. What do you think about some kind of check in this conditional below

```
    if (typeof logs[0] === 'object') {
       return JSON_ARRAY;
    }	
```
checking if if it is a buffer object, and then returning a logType of BUFFER_ARRAY? then we also don't need to check if it is a buffer for every message, just that it is only for the first

Ricky-Thomas: Good idea. I've made this change in the latest refactor. Lemme know what you think!

azure/activity_logs_monitoring/index.js
```diff
@@ -11,6 +11,7 @@ const STRING = 'string'; // example: 'some message'
 const STRING_ARRAY = 'string-array'; // example: ['one message', 'two message', ...]
 const JSON_OBJECT = 'json-object'; // example: {"key": "value"}
 const JSON_ARRAY = 'json-array'; // example: [{"key": "value"}, {"key": "value"}, ...] or [{"records": [{}, {}, ...]}, {"records": [{}, {}, ...]}, ...]
+const BUFFER_ARRAY = 'buffer-array'; // example: <Buffer 7b 22 72 65 63 6f 72 64 73 22 3a 20 5b 7b 20 22 74 69 6d 65 22 3a 20 22 74 65 73 74 69 6e 67 22 7d 5d 7d>
```
claudiadadamo, line 14: This would technically be in an array right? I think we could shorten this to something like `[<Buffer obj>, <Buffer obj>]`.

azure/activity_logs_monitoring/index.js
```diff
@@ -102,6 +106,10 @@ function handleJSONArrayLogs(sender, context, logs, logsType) {
                 return;
             }
         }
+        // If the message is a buffer object, the data type has been set to binary.
+        if (logsType == BUFFER_ARRAY) {
+            message = JSON.parse(message.toString());
```
claudiadadamo, line 111: Let's call this JSON.parse within a try/catch similar to what we do for JSON_STRING_ARRAY in case the string passed in isn't actually json. I'm not sure that we can assume it'll always be JSON

azure/test/client.test.js
```diff
@@ -154,6 +154,13 @@ describe('Azure Log Monitoring', function() {
       testHandleLogs(record, expected, true)
     });
 
+    it('should handle byte array properly', function() {
```
claudiadadamo, line 157: Can you add a test for the case where the message does not have `records` as well as for a buffer array of string logs?

PR #315, opened 2020-07-30
[azure] Refactor and convert Azure eventhub log forwarder to use HTTP

claudiadadamo:
### What does this PR do?
Refactors the azure eventhub log forwarder to use HTTP instead of TCP. These changes include:
- Moving methods into the EventhubLogForwarder class for ease of use, removes the need to pass the context object to everything. Also streamlines what we export for tests.
- Makes the function async so that we can return a promise and the function will wait on the promise to finish running. Wraps the send calls in promises wrapped around the `http.request` call. 
    - Return `Promise.allSettled()` to wait for all promises to finish running even if some fail.
- For tests, ran `npm lint` on the file so there are a lot of changes - The main code changes are creating a new mock eventhublogforwarder object to reference methods to test.

Logic for the http requests is as follows:
  - tries to make request, if it fails then will wait 1 second and then retry. If it fails again will output log message but does not fail the function/
  - http request has a timeout of 2 seconds
  - the function has a timeout of 5 minutes so this should be fine

### Motivation
Better reliability for logs using HTTP

### Testing Guidelines
- Tested by running in my function, for differnet types of log message types. By commenting out the dd-api-key from the headers I can test a request coming back with a 400 and confirmed that it retries, that it waits to retry, and that it correctly logs when it is unsuccessful in sending the message.

Confirmed when the api key is present that logs are sent to datadog: 
image
(I had the api key commented out until a little before 11:15, so that's why there are no logs before that)

Ran unit tests, they all pass.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

- [x] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future

### DIFF
--- a/azure/test/client.test.js
+++ b/azure/test/client.test.js
@@ -52,7 +52,7 @@ describe('Azure Log Monitoring', function() {
                 EventhubLogForwarderInstance.getLogFormat(eventHubMessages)
             );
         });
-        it('should return json no records', function() {
+        it('should return json array when there are no records', function() {
             eventHubMessages = [
                 { key: 'value', otherkey: 'othervalue' },
                 { key: 'value', otherkey: 'othervalue' }
--- a/azure/test/client.test.js
+++ b/azure/test/client.test.js
@@ -1,180 +1,292 @@
 var assert = require('assert');
 var client = require('../activity_logs_monitoring').forTests;
-var constants = client.constants
-var sinon = require('sinon')
+var constants = client.constants;
+var sinon = require('sinon');
 
+function fakeContext() {
+    // create a fake context object to pass into handleLogs
+    contextSpy = sinon.spy();
+    contextSpy.log = sinon.spy();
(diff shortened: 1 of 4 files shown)


azure/test/client.test.js
```diff
@@ -1,180 +1,292 @@
...
+            );
+        });
+        it('should return json object', function() {
+            eventHubMessages = { key: 'value', otherkey: 'othervalue' };
+            assert.equal(
+                constants.JSON_OBJECT,
+                EventhubLogForwarderInstance.getLogFormat(eventHubMessages)
+            );
+        });
+        it('should return json no records', function() {
```
graybjones: The should string is confusing. Can you add a comment about how that maps to a JSON array?

claudiadadamo: This was added before I started working on the log forwarder, but I will make it more clear.

graybjones (approved):
LGTM

PR #319, opened 2020-08-14
Update integration tests to use raw JSON data

nhinsch:
### What does this PR do?

CloudWatch events contain a data field which is a base64-encoded ZIP file of the logs data. Currently we store the input events for the integration tests in this encoded format. For example, the payload for one of our tests is:

```
"awslogs": {
    "data": "H4sIAAAAAAAAAHWPwQqCQBCGX0Xm7EFtK+smZBEUgXoLCdMhFtKV3akI8d0bLYmibvPPN3wz00CJxmQnTO41whwWQRIctmEcB6sQbFC3CjW3XW8kxpOpP+OC22d1Wml1qZkQGtoMsScxaczKN3plG8zlaHIta5KqWsozoTYw3/djzwhpLwivWFGHGpAFe7DL68JlBUk+l7KSN7tCOEJ4M3/qOI49vMHj+zCKdlFqLaU2ZHV2a4Ct/an0/ivdX8oYc1UVX860fQDQiMdxRQEAAA=="
}
```

where the `data` field actually represents:
```
{
	"messageType": "DATA_MESSAGE",
	"owner": "123456789123",
	"logGroup": "testLogGroup",
	"logStream": "testLogStream",
	"subscriptionFilters": ["testFilter"],
	"logEvents": [
		{
			"id": "eventId1",
			"timestamp": 1440442987000,
			"message": "[ERROR] First test message"
		},
		{
			"id": "eventId2",
			"timestamp": 1440442987001,
			"message": "[ERROR] Second test message"
		}
	]
}
```

Storing the input events in the encoded format makes it difficult to see what each test is actually testing. It also makes it more difficult to update tests, because each engineer will need to figure out the encoding format and then decode/re-encode the data.

This PR changes the format of our input events so that we store the raw JSON in our repository. The test runner will encode this on the fly before sending it to the mock forwarder. That way engineers won't need to worry about the encoding any more.

This PR also makes some other, more minor changes to the tests:

- Previously, we had a single test case that ran every snapshot test. This PR breaks that into one test case per snapshot, which will let us see when multiple snapshot tests fail at the same time. This will also allow us to add more tests in the future for different event types; we wouldn't apply the same encoding logic to events from Kinesis, for example.

- Previously, there was a separate test case for initialization. There was a required order for our tests: the initialization test had to run before the snapshot test(s). If the initialization test was skipped, the subsequent snapshot test would fail. This order is not guaranteed and only worked because the initialization test method came first in alphabetical order when compared to the previous snapshot test. In this PR, the initialization check is moved to the `setUp` method that runs before every test. This means that we can run any number of the snapshot tests we want in any order, and they will still pass.

### Motivation

We want to add more integration test coverage. This will allow us to do so more easily.

### Testing Guidelines

The integration tests still pass :) 

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


aws/logs_monitoring/tools/test_harness/tests/test_snapshots.py
```diff
@@ -33,44 +45,57 @@ def compare_snapshot(self, input_filename, snapshot_filename):
...
                 snapshot_file.write(json.dumps(output_data, indent=2))
         else:
-            message = f"Snapshot's didn't match for {input_filename}. To update run `UPDATE_SNAPSHOTS=true ./tools/integration_test.sh"
+            message = f"Snapshots didn't match for {input_filename}. To update run `UPDATE_SNAPSHOTS=true ./tools/integration_test.sh"
             self.assertEqual(output_data, snapshot_data, message)
             pass
 
-    def setup(self):
-        # Clears any recorded state from the mock server
+    def setUp(self):
```
nhinsch, line 68: This function was not being called before due to a capitalization issue

hghotra (approved):

PR #326, opened 2020-08-18
Merge application added ddtags with forwarder added ddtags

tianchu:
### What does this PR do?

When the logs intake pipeline detects a `message` field with a JSON content, it extracts the content to the top-level. The fields of same name from the top-level will be overridden. E.g. the application adds some tags to the log, which appear in the `message.ddtags` field, and the forwarder adds some common tags, such as `aws_account`, which appear in the top-level `ddtags` field:

```python
{
    "message": {
        "ddtags": "mytag:value", # tags added by the application
        ...
    },
    "ddtags": "env:xxx,aws_account", # tags added by the forwarder
    ...
}
```

Only the custom tags added by the application will be kept. We might want to change the intake pipeline to "merge" the conflicting fields rather than "overridding" in the future, but for now we should extract `message.ddtags` and merge it with the top-level `ddtags` field.

### Motivation

Allow customers to add custom tags to logs in Lambda function code when logging in JSON using the `ddtags` field.

### Testing Guidelines

- [x] When `ddtags` present in the log `message` field, its values get extracted and merged with the top-level `ddtags` field. See the added integration test case.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/tests/snapshots/cloudwatch_log_custom_tags.json
+++ b/aws/logs_monitoring/tests/snapshots/cloudwatch_log_custom_tags.json
@@ -1,0 +1,14 @@
+{
+	"messageType": "DATA_MESSAGE",
+	"owner": "123456789123",
+	"logGroup": "testLogGroup",
+	"logStream": "testLogStream",
+	"subscriptionFilters": ["testFilter"],
+	"logEvents": [
+		{
+			"id": "eventId1",
+			"timestamp": 1440442987000,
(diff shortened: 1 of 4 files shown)


aws/logs_monitoring/tests/snapshots/cloudwatch_log_custom_tags.json
```diff
@@ -0,0 +1,19 @@
+{
```
nhinsch, line 1: I think you also need to add a test case to `test_harness/tests/test_snapshots.py`

nhinsch (approved):

PR #327, opened 2020-08-18
David.kim/updated lambda forwarder

Chronobreak:
### What does this PR do?

Adds the following metrics to the log forwarder:
- aws.dd_forwarder.logs_forwarded
- aws.dd_forwarder.metrics_forwarded
- aws.dd_forwarder.traces_forwarded
- aws.dd_forwarder.incoming_events

Along with the following tags:
- forwardername
- forwarder_memorysize
- forwarder_version
- event_type

### Motivation

https://datadoghq.atlassian.net/jira/software/projects/SLS/boards/205?selectedIssue=SLS-572

### Testing Guidelines

- End to end testing on AWS, using apigateway and a test Lambda, both writing to a Cloudwatch log group that triggers the forwarder.
- Integration tests from this repository.

### Additional Notes

There is currently a known issue with dropped payloads when the function is rapidly invoked; for example, for the testing path used, there is a distribution metric payload for apigateway and for Lambda; one of the metric payloads would get dropped. This shouldn't be a blocker here, but once this is resolved there should be 1:1 correlation with the logs/metrics/traces forwarded.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


aws/logs_monitoring/lambda_function.py
```diff
@@ -390,14 +399,24 @@ def datadog_forwarder(event, context):
 
     if DD_FORWARD_LOG:
         forward_logs(filter_logs(map(json.dumps, logs)))
+        DD_FORWARDER_TELEMETRY["logs_forwarded"]+=len(list(filter_logs(map(json.dumps, logs))))
```
tianchu: I don't think it's a good idea to duplicate `filter_logs(map(...` here. Can you move logic into `forward_logs`, where you have filtered logs (and actual count)?

aws/logs_monitoring/lambda_function.py
```diff
@@ -390,14 +399,24 @@ def datadog_forwarder(event, context):
...
 
     forward_metrics(metrics)
+    DD_FORWARDER_TELEMETRY["metrics_forwarded"]+=len(metrics)
```
tianchu: I think it makes more sense to have this logic happen inside the `forward_metrics` function, since it's a small detail which shouldn't /pollute the main (i.e., `datadog_forwarder`) function. I.e., when you read the main function, you can quickly see what this function does on the high level without being distracted by lots of "details".

aws/logs_monitoring/lambda_function.py
```diff
@@ -390,14 +399,24 @@ def datadog_forwarder(event, context):
...
     if len(trace_payloads) > 0:
+        DD_FORWARDER_TELEMETRY["traces_forwarded"]+=len(trace_payloads)
         forward_traces(trace_payloads)
 
     parse_and_submit_enhanced_metrics(logs)
 
+    send_dd_forwarder_telemetry()
+
+    # Reset DD_FORWARDER_TELEMETRY counters and tags after sending
+    DD_FORWARDER_TELEMETRY["logs_forwarded"] = 0
```
tianchu: I don't think you need to keep a counter at all. Why not just submit the `len(metrics)` right away?

aws/logs_monitoring/lambda_function.py
Chronobreak: Hi Tian - moved to `forward_logs` and removed duplication of the `filter_logs` call

aws/logs_monitoring/lambda_function.py
Chronobreak: Refactored/cleaned up so the main function is clean

aws/logs_monitoring/lambda_function.py
Chronobreak: Removed counter construction and submitting telemetry metrics right away (but still need to build the tags list)

aws/logs_monitoring/lambda_function.py
```diff
@@ -389,7 +393,7 @@ def datadog_forwarder(event, context):
     metrics, logs, trace_payloads = split(enrich(parse(event, context)))
 
     if DD_FORWARD_LOG:
-        forward_logs(filter_logs(map(json.dumps, logs)))
```
tianchu, line 392: Nits, I'm not sure about this change, since `forward_logs` should be ONLY about FORWARDING logs.

aws/logs_monitoring/lambda_function.py
```diff
@@ -424,6 +429,15 @@ def forward_logs(logs):
                 if logger.isEnabledFor(logging.DEBUG):
                     logger.debug(f"Forwarded log batch: {json.dumps(batch)}")
 
+    timestamp = time.time()
+
+    lambda_stats.distribution(
+        "aws.dd_forwarder.logs_forwarded",
+        len(list(logs_to_forward)),
```
tianchu: `logs_to_forward` is already a `list`, i don't think you need `list()` here.

aws/logs_monitoring/lambda_function.py
```diff
@@ -424,6 +429,15 @@ def forward_logs(logs):
...
+        timestamp=timestamp,
```
tianchu: I believe `timestamp` is optional, which defaults to the current time.

aws/logs_monitoring/lambda_function.py
```diff
@@ -588,6 +616,10 @@ def split(events):
             trace_payloads.append(trace_payload)
         else:
             logs.append(event)
+
+    """Dedup telemetry tags"""
+    DD_FORWARDER_TELEMETRY_TAGS = list(set(DD_FORWARDER_TELEMETRY_TAGS))
```
tianchu: It's better to avoid the duplicates in the first place. I think you can make `DD_FORWARDER_TELEMETRY_TAGS` to be a dict instead of list, such as:

```python
{
    "forwardername": ...,
    "forwarder_memorysize": ....,
    "forwarder_version": ...,
    "source": ...
}
```

BTW, these four tags are the only tags I expect we add to the telemetries, to be conservative on cardinality.

aws/logs_monitoring/lambda_function.py
```diff
@@ -532,6 +550,15 @@ def generate_metadata(context):
         "forwarder_memorysize": context.memory_limit_in_mb,
         "forwarder_version": DD_FORWARDER_VERSION,
     }
+
+    DD_FORWARDER_TELEMETRY_TAGS.append(
```
tianchu: It's better to avoid piggybacking here. I would rather create a helper function `set_forwarder_telemetry_tags`, and call it from somewhere in the `parse()` function (because we need to know the `source`).

aws/logs_monitoring/lambda_function.py
```diff
@@ -515,6 +529,10 @@ def add_metadata_to_lambda_log(event):
     tags = list(set(tags))
     tags.sort()  # Keep order deterministic
 
+    # Set custom tags on DD_FORWARDER_TELEMETRY
+    DD_FORWARDER_TELEMETRY_TAGS.extend(custom_lambda_tags)
```
tianchu: I don't expect us to add lambda tags to the forwarder, it would dramatically increase the cardinality.

tianchu:
As pointed in other comments, we should also add a `source` tag to the telemetries to tell us where the forwarded data comes from -- s3, cw_logs, cw_events, sns or kinesis.

tianchu:
Can you also help update telemetry [get_resources_api_calls](https://github.com/DataDog/datadog-serverless-functions/blob/45198c262b6b046a3c3d88fab1986cd3413c8dba/aws/logs_monitoring/enhanced_lambda_metrics.py#L316) to use the new `aws.dd_forwarder` namespace? And add the three `forwarder_xxx` tags to it (i don't think `source` makes sense for that telemetry)?

aws/logs_monitoring/lambda_function.py
```diff
@@ -148,6 +149,10 @@ def compileRegex(rule, pattern):
 ]
 
 
+# Used to build and pass aws.dd_forwarder.* telemetry tags
+DD_FORWARDER_TELEMETRY_TAGS = {"source": "", "tags": []}
```
tianchu: I don't think this structure makes too much sense to me, `source` isn't that special except that you may want to exclude it from submission for certain metrics --- in fact, let's make it simpler here and forget about that exception, i.e., let's include `source` and all other tags in all metrics.

tianchu:
```python
DD_FORWARDER_TELEMETRY_TAGS = {}
```

tianchu: If you want to go further you could even make `DD_FORWARDER_TELEMETRY_TAGS` a class inheriting `dict` (so it can be used like a dict) and then you can add a method to this class to help covert the tags to the datadog expected `list`.

aws/logs_monitoring/lambda_function.py
```diff
@@ -424,10 +431,17 @@ def forward_logs(logs):
...
+        "aws.dd_forwarder.logs_forwarded",
+        len(logs_to_forward),
+        tags=DD_FORWARDER_TELEMETRY_TAGS["tags"],
+    )
+
 
 def parse(event, context):
     """Parse Lambda input to normalized events"""
     metadata = generate_metadata(context)
+    event_type = "unknown"
```
tianchu, line 443: I don't think you need to worry about event_type being unknown, `parse_event_type` will raise exception if it cannot be parsed.

aws/logs_monitoring/enhanced_lambda_metrics.py
```diff
@@ -297,6 +301,11 @@ def parse_get_resources_response_for_tags_by_arn(get_resources_page):
     return tags_by_arn
 
 
+def set_enhanced_metrics_tags(tags):
```
tianchu: OK, let's make this simpler and not worry about excluding `source` and simply reuse `DD_FORWARDER_TELEMETRY_TAGS`. If you get circular dependency, it probably make sense to define `DD_FORWARDER_TELEMETRY_TAGS` in a third party file (i.e., not in enhanced_lambda_metrics.py or lambda_function.py`)

tianchu (commented):
Looks like we are getting closer.

aws/logs_monitoring/lambda_function.py
```diff
@@ -448,9 +461,36 @@ def parse(event, context):
...
+    """Helper function to set tags on telemetry metrics
+    Do not submit telemetry metrics before this helper function is invoked
+    """
+    global DD_FORWARDER_TELEMETRY_TAGS
+
+    DD_FORWARDER_TELEMETRY_TAGS = [
+        f"forwardername:{context.function_name.lower()}",
+        f"forwarder_memorysize:{context.memory_limit_in_mb}",
+        f"forwarder_version:{DD_FORWARDER_VERSION}",
+        f"source:{event_type}",
```
tianchu: Let's change the tag key to `event_type` as well for consistency?

aws/logs_monitoring/enhanced_lambda_metrics.py
```diff
@@ -9,6 +9,7 @@
 from botocore.exceptions import ClientError
 
 ENHANCED_METRICS_NAMESPACE_PREFIX = "aws.lambda.enhanced"
+DD_FORWARDER_TELEMETRY_NAMESPACE_PREFIX = "aws.dd_forwarder"
```
tianchu: Perhaps move this to the helper function too? so you don't define it twice?

tianchu (approved):
LGTM, except for two minor comments.

aws/logs_monitoring/tools/integration_tests/snapshots/cloudwatch_log.json~snapshot
```diff
@@ -1,49 +1,55 @@
...
     {
-      "path": "/v1/input/abcdefghijklmnopqrstuvwxyz012345",
+      "path": "/api/v1/distribution_points?api_key=abcdefghijklmnopqrstuvwxyz012345",
       "verb": "POST",
       "content-type": "application/json",
-      "data": [
-        {
-          "id": "eventId1",
-          "timestamp": 1440442987000,
-          "message": "[ERROR] First test message",
```
tianchu: We no longer send logs?

tianchu (approved):
👍 and congrats 🎉

aws/logs_monitoring/tools/integration_tests/snapshots/cloudwatch_log_lambda_invocation.json~snapshot
```diff
@@ -379,19 +379,19 @@
                 "start": "1583425836164691700",
                 "duration": "458278320",
                 "meta": {
-                  "account_id": "0",
```
tianchu, line 382: @nhinsch for later -- probably need to sort the keys of the json snapshot to avoid the flakiness?

PR #329, opened 2020-08-18
Run integration tests on wrapped Forwarder (captures metrics)

nhinsch:
### What does this PR do?

Right now we run integration tests on the Forwarder without the Datadog Lambda wrapper. This is a problem because the wrapper contains logic to flush metrics from a background thread at the end of each function invocation. Without this, the metrics do not appear in the integration test snapshots as they are not sent during the actual function invocation.

By running the integration tests against the wrapped Forwarder, we can capture metrics in our snapshots, including enhanced metrics from the wrapper. We need to remove the specific metric values from the snapshots, however, as they can vary between test runs.

### Motivation

Capturing calls to the metrics intake API will improve our integration test coverage. We also may be losing metrics data in production because of this issue.

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


aws/logs_monitoring/tools/test_harness/recorder/recorder.py
```diff
@@ -12,7 +12,7 @@
 
 class RecorderHandler(BaseHTTPRequestHandler):
     def __init__(self, request, client_address, server):
-        super().__init__(request, client_address, server)()
+        super().__init__(request, client_address, server)
```
nhinsch, line 15: This was raising exceptions in the integration tests, so I cleaned it up

aws/logs_monitoring/tools/Dockerfile_integration
```diff
@@ -5,6 +5,3 @@ ARG forwarder
 
 # Add the code into /var/task (will unzip files)
 ADD $forwarder /var/task/
-
-ENV DD_API_KEY "fake-api-key"
-ENV DD_SITE "datadog.com"
```
nhinsch, line 10: I moved these to `docker-compose` to be with the other env vars

aws/logs_monitoring/lambda_function.py
```diff
@@ -635,6 +635,10 @@ def forward_metrics(metrics):
             if logger.isEnabledFor(logging.DEBUG):
                 logger.debug(f"Forwarded metric: {json.dumps(metric)}")
 
+    # Force the background thread to send metrics
+    # If this does not happen before the function ends, the metrics may not be sent
+    lambda_stats.flush()
```
tianchu: We do flush the metrics (and it blocks) after executing the handler function https://github.com/DataDog/datadog-lambda-python/blob/a1139b4940229821d29a6fd4c20cf06d03e81de4/datadog_lambda/wrapper.py#L141. Perhaps you can add some debugging statements around that line or in the wrapper code to see what's going on?

nhinsch: The integration tests were being run against the unwrapped handler 🤬 

I'll fix this in the morning!

tianchu (approved):

PR #332, opened 2020-08-19
Translate binary data type to be human readable

Ricky-Thomas:
### What does this PR do?

We’d like to allow customers to be able to send logs as binary format to the datadog forwarder. We should handle the binary type (example of where it is set is here: https://a.cl.ly/E0uzNmg0)

Here is what this type of log would look like in DD prior to this change: https://a.cl.ly/yAub7zbb

Here is the same logs after the change: https://a.cl.ly/z8uZEoPv

### Motivation

https://datadoghq.atlassian.net/jira/software/projects/AGAI/boards/207?selectedIssue=AGAI-417

### Testing Guidelines

Tested in the Azure portal and locally with unit tests.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/azure/test/client.test.js
+++ b/azure/test/client.test.js
@@ -268,18 +268,8 @@ describe('Azure Log Monitoring', function() {
         });
 
         it('should handle buffer array without records properly', function() {
-            record = [Buffer.from('{[{ "test": "testing"}]}')];
-            expected = [{ test: 'testing' }];
-            assert.equal(
-                EventhubLogForwarderInstance.getLogFormat(record),
-                constants.BUFFER_ARRAY
-            );
-            testHandleLogs(record, expected, true);
-        });
-
-        it('should handle buffer array without records properly', function() {
-            record = [Buffer.from('{[{ "test": "testing"}]}')];
-            expected = [{ test: 'testing' }];
+            record = [Buffer.from('{ "test": "example"}')];
+            expected = [{ test: 'example' }];
             assert.equal(
                 EventhubLogForwarderInstance.getLogFormat(record),
                 constants.BUFFER_ARRAY
@@ -294,8 +284,6 @@ describe('Azure Log Monitoring', function() {
                 EventhubLogForwarderInstance.getLogFormat(record),
                 constants.BUFFER_ARRAY
             );
-            // just assert that the string method is called for the second message,
-            // we don't care about the first one for this test
             testHandleLogs(record, expected, false);
         });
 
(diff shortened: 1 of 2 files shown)


azure/test/client.test.js
```diff
@@ -257,6 +257,48 @@ describe('Azure Log Monitoring', function() {
...
+            record = [Buffer.from('{[{ "test": "testing"}]}')];
+            expected = [{ test: 'testing' }];
+            assert.equal(
+                EventhubLogForwarderInstance.getLogFormat(record),
+                constants.BUFFER_ARRAY
+            );
+            testHandleLogs(record, expected, true);
+        });
+
+        it('should handle buffer array without records properly', function() {
```
claudiadadamo: This is a duplicate of the above test i think?

azure/test/client.test.js
```diff
@@ -257,6 +257,48 @@ describe('Azure Log Monitoring', function() {
...
+
+        it('should handle buffer array with malformed string', function() {
+            record = [Buffer.from('{"time": "xy')];
+            expected = ['{"time": "xy'];
+            assert.equal(
+                EventhubLogForwarderInstance.getLogFormat(record),
+                constants.BUFFER_ARRAY
+            );
+            // just assert that the string method is called for the second message,
+            // we don't care about the first one for this test
```
claudiadadamo: You can remove this comment, it doesn't apply to the test case you're using

azure/test/client.test.js
```diff
@@ -257,6 +257,48 @@ describe('Azure Log Monitoring', function() {
...
+            expected = [{ test: 'testing' }];
+            assert.equal(
+                EventhubLogForwarderInstance.getLogFormat(record),
+                constants.BUFFER_ARRAY
+            );
+            testHandleLogs(record, expected, true);
+        });
+
+        it('should handle buffer array without records properly', function() {
+            record = [Buffer.from('{[{ "test": "testing"}]}')];
```
claudiadadamo: This isn't actually valid json input, so it's sending this one as a string log instead of json. I think what's happening is that the spies are created once, so with the test above this testing with the same expected output, the spy is saying it was called with this input but that was because it was called above (hard to explain over text, can explain in person tomorrow)

to test this, you can change "testing" to "ricky" and it'll fail. i'll put up a small pr to fix this.

claudiadadamo (approved):

PR #335, opened 2020-08-22
Forwarder: Additional target lambdas

Czechh:
### What does this PR do?

This PR adds an optional parameter in the cloudformation stack to have an additional lambda that gets called with the same event payload the forwarder receives. This can allow customers to ship logs to other vendors such as Logstash (ELK). The parameter is `additionalLambdaTargetARNs`.

### Motivation

This is always a blocker for customers that want to use DD instrumentation, but also ships to another log vendor.

### Testing Guidelines

I **extended** the existing integration tests to deploy two new lambdas. After calling the normal flow of integration tests I then verify that the two lambdas were called with the same data as the forwarder.

Additionally,  I was able to test the Cloudformation template **manually** (via upload) - then updating the forwarder lambda code to my latest code. After that initial set up, I was able to add a lambda to to get instrumented (via the plugin with the new forwarder ARN) After calling the test lambda I verified that the logs were being dual shipped to both the forwarder and the `additionalLambdaTargets` (2 other lambdas in this case).

### Types of changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### Check all that apply

- [X] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future
- [X] This PR passes the integration tests (ask a Datadog member to run the tests)
- [X] This PR passes the unit tests 
- [X] This PR passes the installation tests (ask a Datadog member to run the tests)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -33,3 +33,4 @@
 - [ ] This PR collects user input/sensitive content into Datadog
 - [ ] This PR passes the integration tests (ask a Datadog member to run the tests)
 - [ ] This PR passes the unit tests 
+- [ ] This PR passes the installation tests (ask a Datadog member to run the tests)
```
tianchu, line 36: 👍

aws/logs_monitoring/lambda_function.py
```diff
@@ -1029,3 +1033,16 @@ def parse_service_arn(source, key, bucket, context):
...
     return
+
+
+def invoke_additional_target_lambdas(event):
+    lambda_client = boto3.client("lambda")
+    lambda_arns = DD_ADDITIONAL_TARGET_LAMBDAS.split(",")
+    lambda_payload = json.dumps(event)
+
+    for lambda_arn in lambda_arns:
+        lambda_client.invoke(
```
Czechh: One open question here is how we should handle errors here. Not super familiar with python pattern, so would love to get some feedback

aws/logs_monitoring/README.md
```diff
@@ -334,3 +334,4 @@ The CloudFormation Stack creates following IAM roles:
 - `VPCSecurityGroupIds` - Comma separated list of VPC Security Group Ids. Used when AWS PrivateLink is 
     enabled.
 - `VPCSubnetIds` - Comma separated list of VPC Subnet Ids. Used when AWS PrivateLink is enabled.
+- `AdditionalTargetLambdaARNs` - Comma separated list of lambda ARNs that will get called asynchronously with the same data the forwarder receives.
```
tianchu:
```suggestion
- `AdditionalTargetLambdaARNs` - Comma separated list of Lambda ARNs that will get called asynchronously with the same `event` the Datadog Forwarder receives.
```

aws/logs_monitoring/tools/integration_tests/docker-compose.yml
```diff
@@ -6,6 +6,11 @@ services:
     command: lambda_function.lambda_handler
     environment:
       AWS_ACCOUNT_ID: 0000000000
+      AWS_ACCESS_KEY_ID: "${AWS_ACCESS_KEY_ID}"
+      AWS_SECRET_ACCESS_KEY: "${AWS_SECRET_ACCESS_KEY}"
+      AWS_SECRET_ACCESS_KEY: "${AWS_SECRET_ACCESS_KEY}"
```
tianchu: Need to set `AWS_SECRET_ACCESS_KEY` twice?

.github/workflows/integration_test.yml
```diff
@@ -14,4 +14,4 @@ jobs:
 
       - name: Run integration tests
         run: |
-          ./aws/logs_monitoring/tools/integration_tests/integration_tests.sh --python-version=${{ matrix.python-version }}
+          ./aws/logs_monitoring/tools/integration_tests/integration_tests.sh --python-version=${{ matrix.python-version }} --no-additional-lambda
```
tianchu: Why `--no-additional-lambda`? What's the concerns?

aws/logs_monitoring/tools/integration_tests/docker-compose.yml
Czechh: Good catch, removing the dup.

aws/logs_monitoring/lambda_function.py
tianchu: How about?

```python
for lambda_arn in lambda_arns:
    try:
        lambda_client.invoke(
            FunctionName=lambda_arn, InvocationType="Event", Payload=lambda_payload,
        )
    except Exception as e:
        logger.exception(f"Failed to invoke additional target lambda {lambda_arn} due to {e}")
```

.github/workflows/integration_test.yml
Czechh: For the additional lambda case we need to provision a couple of lambdas and get their logs from cloudwatch. So locally you would run this using `aws-vault` since these run in Github I'm not sure if we have a secure way to set this up.

tianchu: Got it, perhaps we could drop the requirements here to just have a unit test? Given the nature that AWS is responsible for invoking the lambda function, integration test might be an overkill and we just to ensure `invoke` is indeed called?

My concern is that after a while the integration tests for this case will be broken and unmaintained since it does not run automatically.

Czechh: that makes sense. I'll add a unit test instead and delete that integration test changes

Czechh: Alternatively, I could make the default to run without those lambda cases. And make sure that an option is required to actually run with the lambda case.

Czechh: curious to get @DarcyRaynerDD thoughts here too.

tianchu: Yeah, definitely does not hurt to keep the code if you like, assuming you change the default to not run them.

tianchu (approved):

Czechh:
Added unit tests and changed the default behavior for the integration tests so they don't include the external lambda case.

PR #351, opened 2020-09-17
Add support for S3 events triggered via SNS

sagar-khanna:
### What does this PR do?
- When S3 events are configured to push to SNS topics
- The DD Lambda Forwarder subscribes to the SNS topic
- Then in DD console the event shows up as SNS with SNS event data being logged rather than the S3 log being ingested
- This change extracts the S3 event from the SNS topic and then extract and ingests the S3 log from it

The new behavior should be desired/expected to most of users, but technically speaking it is a breaking change, because SNS logs carrying S3 events will stop appearing in Datadog, while the corresponding S3 objects will start showing up.

### Motivation
Talked to Tian (at Datadog NY offices) and we needed a solution for our organisation so wrote this PR.

### Testing Guidelines
We are using this codebase in our organisation now.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [x] Breaking change

### Check all that apply

- [x] This PR contains breaking changes that are documented in the description

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -797,11 +797,16 @@ def parse_event_type(event):
         if "s3" in event["Records"][0]:
             return "s3"
         elif "Sns" in event["Records"][0]:
-            snsS3EventMessage = event["Records"][0]["Sns"]["Message"]
-            if len(snsS3EventMessage) > 0:
-                snsS3Event = json.loads(snsS3EventMessage)
-                if "Records" in snsS3Event and "s3" in snsS3Event["Records"][0]:
+            # it's not uncommon to fan out s3 notifications through SNS,
+            # should treat it as an s3 event rather than sns event.
+            sns_msg = event["Records"][0]["Sns"]["Message"]
+            try:
+                sns_msg_dict = json.loads(sns_msg)
+                if "Records" in sns_msg_dict and "s3" in sns_msg_dict["Records"][0]:
                     return "s3"
+            except Exception:
+                if logger.isEnabledFor(logging.DEBUG):
+                    logger.debug(f"No s3 event detected from SNS message: {sns_msg}")
             return "sns"
         elif "kinesis" in event["Records"][0]:
             return "kinesis"
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -821,12 +821,9 @@ def s3_handler(event, context, metadata):
         os.environ["AWS_REGION"],
         config=botocore.config.Config(s3={"addressing_style": "path"}),
     )
+    # if this is a S3 event carried in a SNS message, extract it and override the event
     if "Sns" in event["Records"][0]:
-        snsS3EventMessage = event["Records"][0]["Sns"]["Message"]
-        if len(snsS3EventMessage) > 0:
-            snsS3Event = json.loads(snsS3EventMessage)
-            if "Records" in snsS3Event and len(snsS3Event["Records"]) > 0:
-                event = snsS3Event
+        event = json.loads(event["Records"][0]["Sns"]["Message"])
 
     # Get the object from the event and show its content type
     bucket = event["Records"][0]["s3"]["bucket"]["name"]
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -797,6 +797,11 @@ def parse_event_type(event):
         if "s3" in event["Records"][0]:
             return "s3"
         elif "Sns" in event["Records"][0]:
+            snsS3EventMessage = event["Records"][0]["Sns"]["Message"]
+            if len(snsS3EventMessage) > 0:
+                snsS3Event = json.loads(snsS3EventMessage)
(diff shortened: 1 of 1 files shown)


sagar-khanna:
@tianchu - FYI

aws/logs_monitoring/lambda_function.py
```diff
@@ -797,6 +797,11 @@ def parse_event_type(event):
         if "s3" in event["Records"][0]:
             return "s3"
         elif "Sns" in event["Records"][0]:
+            snsS3EventMessage = event["Records"][0]["Sns"]["Message"]
+            if len(snsS3EventMessage) > 0:
+                snsS3Event = json.loads(snsS3EventMessage)
+                if "Records" in snsS3Event and "s3" in snsS3Event["Records"][0]:
+                    return "s3"
```
tianchu: Some nits and adding a try...except...to avoid validating all the assumptions (e.g., whether the sns message is a valid json, etc.). 
```suggestion
            # it's not uncommon to fan out s3 notifications through SNS,
            # should treat it as an s3 event rather than sns event.
            sns_msg = event["Records"][0]["Sns"]["Message"]
            try:
                sns_msg_dict = json.loads(sns_msg)
                if "Records" in sns_msg_dict and "s3" in sns_msg_dict["Records"][0]:
                    return "s3"
            except Exception:
                if logger.isEnabledFor(logging.DEBUG):
                    logger.debug(f"No s3 event detected from SNS message: {sns_msg}")
```

aws/logs_monitoring/lambda_function.py
```diff
@@ -816,6 +821,12 @@ def s3_handler(event, context, metadata):
         os.environ["AWS_REGION"],
         config=botocore.config.Config(s3={"addressing_style": "path"}),
     )
+    if "Sns" in event["Records"][0]:
+        snsS3EventMessage = event["Records"][0]["Sns"]["Message"]
+        if len(snsS3EventMessage) > 0:
+            snsS3Event = json.loads(snsS3EventMessage)
+            if "Records" in snsS3Event and len(snsS3Event["Records"]) > 0:
+                event = snsS3Event
```
tianchu: I think some of those checks are probably redundant since you already checked them when rerouting the event to s3?

```suggestion
    # if this is a S3 event carried in a SNS message, extract it and override the event
    if "Sns" in event["Records"][0]:
        event = json.loads(event["Records"][0]["Sns"]["Message"])
```

tianchu (commented):
@sagar-khanna made a few suggestions, can you test it again with the suggested code changes (if makes sense to you)? please let me know if it works, so we can merge the PR. Thanks!

aws/logs_monitoring/lambda_function.py
sagar-khanna: Should we put a debug message if there is an exception instead of pass?

tianchu: Updated the suggested code.

tweakster:
> @sagar-khanna made a few suggestions, can you test it again with the suggested code changes (if makes sense to you)? please let me know if it works, so we can merge the PR. Thanks!

Thanks @sagar-khanna and @tianchu for this. Just what I was looking for.

Just FYI I have had this running for us and it seems to be working okay (I had it running using S3 triggers and then switched over to SNS, so it seems fine for both).

tweakster:
> ### Check all that apply
> * [x]  This PR's description is comprehensive
> * [ ]  This PR contains breaking changes that are documented in the description

Is this a breaking change? If someone was pushing SNS triggers from S3 into logs then the behaviour will change. Personally I cannot see why anyone would do that, but you never know.

aws/logs_monitoring/lambda_function.py
tianchu: @sagar-khanna did you find a chance to test with my suggested code changes? If it works for you, do you mind updating the PR so I can merge?

sagar-khanna:
@tianchu  - tested with the changes today and its all working. Let me know if there's anything else you want me to do

sagar-khanna:
> > ### Check all that apply
> > 
> > * [x]   This PR's description is comprehensive
> > * [ ]   This PR contains breaking changes that are documented in the description
> 
> Is this a breaking change? If someone was pushing SNS triggers from S3 into logs then the behaviour will change. Personally I cannot see why anyone would do that, but you never know.

@tweakster  - I'm not sure how that would work. Are you saying an S3 bucket would subscribe to an SNS trigger?
@tianchu  - any thoughts on this?

tianchu:
> > > ### Check all that apply
> > > 
> > > * [x]    This PR's description is comprehensive
> > > * [ ]    This PR contains breaking changes that are documented in the description
> > 
> > 
> > Is this a breaking change? If someone was pushing SNS triggers from S3 into logs then the behaviour will change. Personally I cannot see why anyone would do that, but you never know.
> 
> @tweakster - I'm not sure how that would work. Are you saying an S3 bucket would subscribe to an SNS trigger?
> @tianchu - any thoughts on this?

@sagar-khanna, I think @tweakster meant that those SNS logs (carrying s3 events) will stop appearing in Datadog with your changes, which is a desired change for you and many, but could potentially be an undesired breaking change for others.

I cannot think of a good reason why would anyone prefer the old behavior, so I think we can go ahead, though to be safe, I will make a "breaking change" note in release notes.

PR #353, opened 2020-09-22
[aws-ints] change rds lambda runtime to 3.7

jvanbrie:
### What does this PR do?
This pr changes the runtime for the rds enhanced monitoring lambda to python 3.7.

### Motivation
We've changed the code to work with python3, so we should also change the runtime to be consistent.


klivan (approved):
LGTM 👍

PR #354, opened 2020-09-22
[azure] Add new source types and other cleanup

claudiadadamo:
### What does this PR do?

- Adds new source type support for `azure.subscription` and `azure.resourcegroup`
- Refactors the way we split the resource ID so indexing makes more sense
- Replaces `Promise.allSettled` with `Promise.all`, since allSettled is fairly new and may cause problems for existing customers moving to the new version of the code.
- Renames extractResourceId method to extractMetadataFromResource
- Fixes issue with some azure logs coming in with sources that aren't formatted the same way as others, since `microsoft.` isn't in field we use for parsing the source.

### Motivation

More support for log forwarder, ease of dev

### Testing Guidelines

Currently running in our demo subscription, confirmed it's still sending logs. Also tested by throwing an error to confirm the function does not fail when one Promise rejects.

Logs for the test or show both azure.subscription and azure.resourcegroup sources.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/azure/activity_logs_monitoring/index.js
+++ b/azure/activity_logs_monitoring/index.js
@@ -226,7 +226,11 @@ class EventhubLogForwarder {
     }
 
     formatSourceType(sourceType) {
-        return sourceType.replace('microsoft.', 'azure.');
+        if (sourceType.includes('microsoft')) {
+            return sourceType.replace('microsoft.', 'azure.');
+        } else {
+            return '';
+        }
     }
 
     extractMetadataFromResource(record) {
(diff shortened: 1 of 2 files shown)


azure/activity_logs_monitoring/index.js
```diff
@@ -225,39 +225,58 @@ class EventhubLogForwarder {
...
             return metadata;
-        } else if (
-            record.resourceId.toLowerCase().startsWith('/subscriptions/')
-        ) {
-            var resourceId = record.resourceId.toLowerCase().split('/');
-            if (resourceId.length > 2) {
-                metadata.tags.push('subscription_id:' + resourceId[2]);
+        }
+        var resourceId = record.resourceId.toLowerCase().split('/');
+        if (resourceId[0] === '') {
```
SamGRosen, line 245: I'm not sure what format the resource id's are, but this could have unexpected behavior if the string has the form `"// multiple slashes"`.

claudiadadamo: We really only expect the resource to come in like starting with a slash, and splitting by slash will give us an empty string as the first item. I made it conditional in case for some reason azure sends us something without a slash but I don't think that's likely, and i don't think it would be likely that they'd send something with multiple slashes.

Ricky-Thomas (approved):
Nice! LGTM 👍

Issue #375, opened 2020-10-17
apikey length error when using secretmanager


wangzz2019:
**Describe what happened:**
have got this error when using secretmanger
`
[ERROR] Exception: The API key is not the expected length. Please confirm that your API key is correctTraceback (most recent call last):  File "/var/lang/lib/python3.7/imp.py", line 234, in load_module    return load_source(name, filename, file)  File "/var/lang/lib/python3.7/imp.py", line 171, in load_source    module = _load(spec)  File "<frozen importlib._bootstrap>", line 696, in _load  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked  File "<frozen importlib._bootstrap_external>", line 728, in exec_module  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed  File "/var/task/lambda_function.py", line 79, in <module>    "The API key is not the expected length. " | [ERROR] Exception: The API key is not the expected length. Please confirm that your API key is correct Traceback (most recent call last):   File "/var/lang/lib/python3.7/imp.py", line 234, in load_module     return load_source(name, filename, file)   File "/var/lang/lib/python3.7/imp.py", line 171, in load_source     module = _load(spec)   File "<frozen importlib._bootstrap>", line 696, in _load   File "<frozen importlib._bootstrap>", line 677, in _load_unlocked   File "<frozen importlib._bootstrap_external>", line 728, in exec_module   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed   File "/var/task/lambda_function.py", line 79, in <module>     "The API key is not the expected length. "
-- | --

`

**Describe what you expected:**

**Steps to reproduce the issue:**

1. Save the dd apikey in secret manager first
2. Configure DD_API_KEY_SECRET_ARN
3. Set secretsmanager:GetSecretValue permission

tianchu:
Closing this one to keep the discussion in https://github.com/DataDog/datadog-serverless-functions/pull/376

PR #376, opened 2020-10-17
add secret_key for getting secret value

wangzz2019:
### What does this PR do?

Described in issue #375 
would like to add DD_API_KEY_SECRET_KEY for secret key setting

### Motivation

would like to solve the issue soon

### Types of changes

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/settings.py
+++ b/aws/logs_monitoring/settings.py
@@ -181,7 +181,6 @@ EXCLUDE_AT_MATCH = get_env_var("EXCLUDE_AT_MATCH", default=None)
 # DD API Key
 if "DD_API_KEY_SECRET_ARN" in os.environ:
     SECRET_ARN = os.environ["DD_API_KEY_SECRET_ARN"]
-    
     secretstring = boto3.client("secretsmanager").get_secret_value(SecretId=SECRET_ARN)[
         "SecretString"
     ]
--- a/aws/logs_monitoring/settings.py
+++ b/aws/logs_monitoring/settings.py
@@ -181,11 +181,15 @@ EXCLUDE_AT_MATCH = get_env_var("EXCLUDE_AT_MATCH", default=None)
 # DD API Key
 if "DD_API_KEY_SECRET_ARN" in os.environ:
     SECRET_ARN = os.environ["DD_API_KEY_SECRET_ARN"]
-    SECRET_KEY = os.environ["DD_API_KEY_SECRET_KEY"]
+    
     secretstring = boto3.client("secretsmanager").get_secret_value(SecretId=SECRET_ARN)[
         "SecretString"
     ]
-    DD_API_KEY = json.loads(secretstring)[SECRET_KEY]
+    if "DD_API_KEY_SECRET_KEY" in os.environ:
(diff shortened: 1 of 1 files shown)


aws/logs_monitoring/settings.py
```diff
@@ -181,9 +181,11 @@ def __init__(self, name, pattern, placeholder):
 # DD API Key
 if "DD_API_KEY_SECRET_ARN" in os.environ:
     SECRET_ARN = os.environ["DD_API_KEY_SECRET_ARN"]
-    DD_API_KEY = boto3.client("secretsmanager").get_secret_value(SecretId=SECRET_ARN)[
+    SECRET_KEY = os.environ["DD_API_KEY_SECRET_KEY"]
+    secretstring = boto3.client("secretsmanager").get_secret_value(SecretId=SECRET_ARN)[
         "SecretString"
     ]
+    DD_API_KEY = json.loads(secretstring)[SECRET_KEY]
```
bryantbiggs: SecretsManager can return plaintext, its not always json and for the forwarders I believe it is the plaintext that is used so this would fail

wangzz2019: Yes, secretstring is a plaintext as string
```
>>> import json
>>> s='{"secret_key":"apikey"}'
>>> print(s)
{"secret_key":"apikey"}
>>> print(json.loads(s)["secret_key"])
apikey
```

bryantbiggs: sorry, what I meant to say is that the value stored in SecretsManager is just the API key itself like `'xxxxxxxxxxxx'` which means this would break for everyone who has it setup this way

wangzz2019: Oh, Yes, agree with you bryanbigs.
If so, at least we should provide an options for people who used secret_key

tianchu: @wangzz2019 

> If so, at least we should provide an options for people who used secret_key

Does any AWS or Datadog doc lead you to storing the api key in the `{"secret_key":"apikey"}` JSON format, instead of in the plain-text/string format?

wangzz2019: @tianchu 
There is nothing mentioned in our docs, but here is the AWS documents:
https://docs.aws.amazon.com/secretsmanager/latest/userguide/manage_create-basic-secret.html
```
Specify the details of your custom secret as Key and Value pairs. For example, you can specify a key of UserName, and then supply the appropriate user name as the value. Add a second key with the name of Password and the password text as the value. You could also add entries for Database name, Server address, TCP port, and so on. You can add as many pairs as you need to store the information you require.

Alternatively, you can choose the Plaintext tab and enter the secret value in any format.

```
Key Value pairs is the default options on console.
If we only support plaintext, it is better to mention it in our docs

tianchu:
Closing this PR, as I think it's better to update the doc, rather than giving users two options to store the api key in secrets manager, which may cause even more confusion in the future. See the PR for doc updates: https://github.com/DataDog/datadog-serverless-functions/pull/385

PR #393, opened 2020-11-23
Send logs concurrently

tianchu:
### What does this PR do?

Send AWS logs concurrently for better performance over handling larger log files (e.g., vpc log files from s3). Also increased the batch size to reduce # of batches.

CFN parameter `DdMaxWorkers` can be used to fine control the concurrency.

### Motivation

Forwarder may time out when handling large log files.

### Testing Guidelines

For a large vpc log file containing 800k log lines, duration reduced by more than 10X while memory usage remained the same

**Before:**
REPORT RequestId: XXX Duration: 650270.61 ms Billed Duration: 650300 ms Memory Size: 3008 MB Max Memory Used: 1454 MB

**After:**
REPORT RequestId: XXX Duration: 42998.54 ms	Billed Duration: 43000 ms Memory Size: 3008 MB Max Memory Used: 1503 MB

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/template.yaml
+++ b/aws/logs_monitoring/template.yaml
@@ -166,7 +166,11 @@ Parameters:
     Type: Number
     Default: 6
     AllowedValues: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
-    Description: Set the compression level from 0 (no compression) to 9 (best compression).
+    Description: Set the compression level from 0 (no compression) to 9 (best compression) when sending logs.
+  DdMaxWorkers:
+    Type: Number
+    Default: 20
+    Description: Set the max number of workers sending logs concurrently.
   PermissionsBoundaryArn:
     Type: String
     Default: ""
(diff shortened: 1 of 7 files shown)


agocs (approved):

nhinsch:
I would add the new environment variable/parameter to the README, and I would also briefly explain why the customer might want to adjust the number of workers (I assume adding more workers means logs are forwarded faster, but the function uses more memory?).

Also, is it strictly necessary to expose the setting for the number of workers? If there is a value that you think would work for all customers, could we just hardcode that number of workers for greater simplicity?

aws/logs_monitoring/template.yaml
```diff
@@ -358,6 +367,8 @@ Resources:
               - Ref: AWS::NoValue
           DD_SITE:
             Ref: DdSite
+          DD_:
+            Ref: DdSite
```
guillaumelecerf, line 371: Is this really needed or a copy/paste error?

PR #394, opened 2020-11-25
Support for secret manager partial ARNs

c-py:
### What does this PR do?

This PR loosens the IAM policy to support [secret manager partial ARNs](https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_GetSecretValue.html#SecretsManager-GetSecretValue-request-SecretId). Secrets manager randomly adds six characters to the ARN of a secret.

In a multi region deployment, each secret ARN gets a random six character suffix. The same input ARN cannot be used across regions.

### Motivation

We deploy the datadog forwarder via CloudFormation Stack Sets. We have the datadog key in a centralised account, and have to host the datadog template locally, because the `DdApiKeySecretArn` can't use short form. This PR would fix this, so we can use the datadog provided template.

```
AWSTemplateFormatVersion: "2010-09-09"
Description: Add lambda monitoring to datadog. Designed to be used via CloudFormation stack sets.
Parameters:
  DataDogTags:
    Type: String
    Description: Tags
    Default: ""
  DataDogExcludeRegex:
    Type: String
    Description: Regex to exclude logs
Resources:
  DataDogStack:
    Type: "AWS::CloudFormation::Stack"
    Properties:
      TemplateURL: "./templates/datadog.yml"
      Parameters:
        FunctionName: DatadogForwarder
        DdApiKey: "None"
        DdApiKeySecretArn: !Sub "arn:aws:secretsmanager:${AWS::Region}:HIDDEN:secret:DATADOG_KEY"
        DdTags: !Ref DataDogTags
        ExcludeAtMatch: !Ref DataDogExcludeRegex
        DdFetchLambdaTags: True
```

### Types of changes

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/template.yaml
+++ b/aws/logs_monitoring/template.yaml
@@ -527,7 +527,7 @@ Resources:
                 Fn::If:
                   - CreateDdApiKeySecret
                   - Ref: DdApiKeySecret
-                  - Ref: DdApiKeySecretArn
+                  - !Sub "${DdApiKeySecretArn}*"
             - !If
               - SetDdFetchLambdaTags
               - Effect: Allow


tianchu:
@c-py not sure how we missed your PR initially, but just to make sure I understand the issue correctly, you create Datadog API Key in a different account, and you have one secret for each region with different suffix (generated by AWS). When you create the Datadog Forwarder stack set, you would like to provide the short ARN of the secret (without the random suffix, because they differ by regions)? The current IAM policy won't allow the Forwarder to access the secrets because the value provided to `DdApiKeySecretArn` is incomplete? If so, I think your PR does make sense to me. However, I thought we had other problems with Stack Set, e.g., https://github.com/DataDog/datadog-serverless-functions/issues/423.

c-py:
@tianchu That's correct. It's working perfectly for us except for the fact we need to host our own version of the forwarder template with the modification in this PR.

The difference between our setup and #423 is that our stack set deploys a nested template in the target account to work around the transform limitation.

#423 is like this:
```
StackSet:
  Account 1 CloudFormation (Datadog Template):
    AWS::Serverless::Function
  Account 2 CloudFormation (Datadog Template):
    AWS::Serverless::Function
```

We are like this:
```
StackSet:
  Account 1 CloudFormation (Our Template):
    AWS::CloudFormation::Stack (Datadog Template):
      AWS::Serverless::Function
  Account 2 CloudFormation (Our Template):
    AWS::CloudFormation::Stack (Datadog Template):
      AWS::Serverless::Function
```

tianchu:
@c-py thanks for your explanation and your PR 🙇‍♂️ ! They make perfect sense to me. Merging now!

PR #397, opened 2020-12-02
[azure] Handle provider-only resource IDs in log forwarder

claudiadadamo:
### What does this PR do?
Adds logic to parse resource IDs that just have the 'providers' section, without a resource group. This seems to only be a problem for azure.recoveryservices sources, and currently they fall back to the 'azure' source if they don't work but this change makes these logs more consistent.

### Motivation
Improving azure log forwarder

### Testing Guidelines
Tested by sending a test request with a resource ID that I'd expect to be parsed and confirmed that it parsed correctly. Also updated and ran unit tests.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/azure/activity_logs_monitoring/index.js
+++ b/azure/activity_logs_monitoring/index.js
@@ -285,6 +285,7 @@ class EventhubLogForwarder {
     }
 
     createResourceIdArray(record) {
+        // Convert the resource ID in the record to an array, handling beginning/ending slashes
         var resourceId = record.resourceId.toLowerCase().split('/');
         if (resourceId[0] === '') {
             resourceId = resourceId.slice(1);
@@ -296,6 +297,7 @@ class EventhubLogForwarder {
     }
 
     isSource(resourceIdPart) {
+        // Determine if a section of a resource ID counts as a "source," in our case it means it starts with 'microsoft.'
         return resourceIdPart.startsWith('microsoft.') ? true : false;
     }
 
@@ -327,6 +329,7 @@ class EventhubLogForwarder {
                     resourceId[2] === 'providers' &&
                     this.isSource(resourceId[3])
                 ) {
+                    // handle provider-only resource IDs
                     metadata.source = this.formatSourceType(resourceId[3]);
                 } else {
                     metadata.tags.push('resource_group:' + resourceId[3]);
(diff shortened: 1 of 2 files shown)


azure/activity_logs_monitoring/index.js
```diff
@@ -284,12 +284,25 @@ class EventhubLogForwarder {
...
+        }
+        if (resourceId[resourceId.length - 1] === '') {
+            resourceId.pop();
         }
+        return resourceId;
+    }
+
+    isSource(resourceIdPart) {
+        // Determine if a section of a resource ID counts as a "source," in our case it means it starts with 'microsoft.'
+        return resourceIdPart.startsWith('microsoft.') ? true : false;
```
SamGRosen: This could just be `return resourceIdPart.startsWith('microsoft.');`

claudiadadamo: Wow good catch idk what I was thinking, I'll update

SamGRosen (approved):
Does it mean I'm going crazy if I find it refreshing to read some JavaScript?

azure/test/client.test.js
```diff
@@ -292,6 +292,20 @@ describe('Azure Log Monitoring', function() {
                 this.forwarder.extractMetadataFromResource(record)
             );
         });
+        it('should correctly parse provider-only resource ids', function() {
+            record = {
+                resourceId:
+                    '/SUBSCRIPTIONS/12345678-1234-ABCD-1234-1234567890AB/PROVIDERS/MICROSOFT.RECOVERYSERVICES/SOMETHING/SOMETHINGELSE'
```
SamGRosen, line 298: Another useful test case would be a resource id with `microsoft.` in the actual resource name, to ensure we don't change customer-made names.

claudiadadamo: We don't actually change the resource id, we just use the part of the resource ID with the `microsoft.` in it to determine the source, and add that to the metadata replacing `microsoft.` with `azure.`

claudiadadamo: Or do you mean when parsing out the source that we don't grab an incorrect field?

SamGRosen: Yes mostly just parsing out the correct field. It's just a test case I like to include when I write code with a .replace in it.

PR #345, opened 2020-09-10
Implemented same Datadog API-key retrieval into AWS lambda  as log forwareder has

HQJaTu:
### What does this PR do?

AWS RDS enhanced monitoring lambda function requires Datadog API-key via `kmsEncryptedKeys` environment variable. This is completely different approach than AWS log forwarder lambda has. 

### Motivation

To harmonise the API-key interface, code from log forwarder lambda was incorporated into RDS monitor. Now I don't need to setup Datadog API-key twice.

### Testing Guidelines

This is a trivial change. I'm running both lambdas in my environment.

### Additional Notes

none

### Types of changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Misc (docs, refactoring, dependency upgrade, etc.)

### Check all that apply

- [X] This PR's description is comprehensive
- [ ] This PR contains breaking changes that are documented in the description
- [ ] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future
- [ ] This PR impacts documentation, and it has been updated (or a ticket has been logged)
- [ ] This PR's changes are covered by the automated tests
- [ ] This PR collects user input/sensitive content into Datadog
- [ ] This PR passes the integration tests (ask a Datadog member to run the tests)
- [ ] This PR passes the unit tests 
- [ ] This PR passes the installation tests (ask a Datadog member to run the tests)

PR #405, opened 2020-12-17
[RDS EH] Add more options for passing in API key

thieman:
### What does this PR do?

This is carried from a lovely contribution over in https://github.com/DataDog/datadog-serverless-functions/pull/345. 

Currently, the RDS Enhanced Monitoring Lambda can only accept the necessary Datadog API key through an encrypted payload provided under the `kmsEncryptedKeys` environment variable. This PR adds the following options, with implementation matching what's used in the log forwarder:

- `DD_API_KEY_SECRET_ARN` - an ARN in AWS Secrets Manager
- `DD_API_KEY_SSM_NAME` - a name of a secret in AWS Systems Manager
- `DD_KMS_API_KEY` - a key directly encrypted through KMS. This is differentiated from the original `kmsEncryptedKeys` payload in that this one is not wrapped in a JSON object
- `DD_API_KEY` - the key in plaintext (not recommended for production use cases)

Note that using the Secrets Manager or Systems Manager methods requires permissions that are not currently added by default through the normal setup process for this Lambda, not sure whether those should be included or left to the implementer.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/rds_enhanced_monitoring/README.md
+++ b/aws/rds_enhanced_monitoring/README.md
@@ -196,6 +196,8 @@ d. **Not Recommended**: Plaintext
 
    - Set the relevant environment variable with the API key payload you generated in step 1.
 
+   - If you use Datadog's EU platform, set the environment variable `DD_SITE` to `datadoghq.eu`
+
 2. Subscribe to the appropriate log stream.
 
 # How to update the zip file for the AWS Serverless Apps
--- a/aws/rds_enhanced_monitoring/README.md
+++ b/aws/rds_enhanced_monitoring/README.md
@@ -133,14 +133,35 @@ Process a RDS enhanced monitoring DATA_MESSAGE, coming from CLOUDWATCH LOGS
 
 # Setup
 
-1. Create a KMS key for the Datadog API key and app key
-   - Create a KMS key. Refer to the [AWS KMS Creating Keys][1] documentation for step by step instructions.
-   - Encrypt the token using the AWS CLI. `aws kms encrypt --key-id alias/<KMS key name> --plaintext '{"api_key":"<dd_api_key>", "app_key":"<dd_app_key>"}'`
-   - Make sure to save the base-64 encoded, encrypted key (`CiphertextBlob`). This is used for the `KMS_ENCRYPTED_KEYS` variable in all lambda functions.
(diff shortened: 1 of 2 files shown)


jvanbrie (commented):
This looks good to me. We probably want to make some changes to the docs to mention these new API key options: https://github.com/DataDog/datadog-serverless-functions/tree/master/aws/rds_enhanced_monitoring#setup

thieman:
Updated README with the new options. Couple things to note:

1. Actually removed the `kmsEncryptedKeys` option from the README as doing it with the current version of the Lambda requires knowing the name of the Lambda ahead of time. It'll still work if you create it using the process found at the public docs here, but it didn't make sense for the CLI setup flow. https://docs.datadoghq.com/integrations/amazon_rds/?tab=enhanced

2. Explicitly recommended `DD_KMS_API_KEY` as the type to use when going through these instructions. This will work either with or without the context. IAM policies are also only provided for KMS flows as this is the recommended one.

aws/rds_enhanced_monitoring/README.md
```diff
@@ -133,14 +133,35 @@ Process a RDS enhanced monitoring DATA_MESSAGE, coming from CLOUDWATCH LOGS
 
 # Setup
 
-1. Create a KMS key for the Datadog API key and app key
-   - Create a KMS key. Refer to the [AWS KMS Creating Keys][1] documentation for step by step instructions.
-   - Encrypt the token using the AWS CLI. `aws kms encrypt --key-id alias/<KMS key name> --plaintext '{"api_key":"<dd_api_key>", "app_key":"<dd_app_key>"}'`
-   - Make sure to save the base-64 encoded, encrypted key (`CiphertextBlob`). This is used for the `KMS_ENCRYPTED_KEYS` variable in all lambda functions.
-   - Optional: set the environment variable `DD_SITE` to `datadoghq.eu` to automatically forward data to your EU platform.
```
jvanbrie, line 140: I think we should keep something in the doc about using this env variable if in eu.

thieman: Whoops, adding that back

PR #419, opened 2021-01-27
replace host for cloudtrail with aws hostname

mellon85:
### What does this PR do?

Parses `cloudwatch` events `arn` as in case it's an `assumed-role` event the AWS hostname is contained.

As the DD_HOST field is overridden at the normalization step I had to a post processing in the normalization function.

### Motivation

Improve customer experience as the host will refer to the actual host in AWS.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


aws/logs_monitoring/lambda_function.py
```diff
@@ -762,13 +791,22 @@ def normalize_events(events, metadata):
...
-            normalized.append(merge_dicts({"message": event}, metadata))
+            normalized_event = merge_dicts({"message": event}, metadata)
         else:
             # drop this log
             continue
 
+        # in case it's a cloudtrail event replace the hostname with the
+        # hostname in the ARN if available.
+        if normalized_event.get(DD_SOURCE) == "cloudtrail":
+            extracted_arn_hostname = extract_arn_hostname(normalized_event["message"])
```
jmassey-datadog:
```suggestion
            extracted_arn_hostname = extract_arn_hostname(normalized_event)
```

There is no `message` key. Passing in the normalized_event dict will work just fine.

aws/logs_monitoring/lambda_function.py
```diff
@@ -757,13 +786,22 @@ def normalize_events(events, metadata):
     for event in events:
         events_counter += 1
         if isinstance(event, dict):
-            normalized.append(merge_dicts(event, metadata))
+            normalized_event = merge_dicts(event, metadata)
```
tianchu: Why do you drop the `append`? Won't this break everything?

aws/logs_monitoring/lambda_function.py
```diff
@@ -757,13 +786,22 @@ def normalize_events(events, metadata):
...
         elif isinstance(event, str):
-            normalized.append(merge_dicts({"message": event}, metadata))
+            normalized_event = merge_dicts({"message": event}, metadata)
         else:
             # drop this log
             continue
 
+        # in case it's a cloudtrail event replace the hostname with the
+        # hostname in the ARN if available.
+        if normalized_event.get(DD_SOURCE) == "cloudtrail":
```
tianchu: It's better to keep `normalize_events` be a generic function. The best place to parse the message and override `DD_HOST` are within the log handler functions:

https://github.com/DataDog/datadog-serverless-functions/blob/615c4e7a43d65e542fa83ac5908d47f7601baab3/aws/logs_monitoring/lambda_function.py#L883 and https://github.com/DataDog/datadog-serverless-functions/blob/615c4e7a43d65e542fa83ac5908d47f7601baab3/aws/logs_monitoring/lambda_function.py#L805

For example, this is how we override RDS's host:
https://github.com/DataDog/datadog-serverless-functions/blob/615c4e7a43d65e542fa83ac5908d47f7601baab3/aws/logs_monitoring/lambda_function.py#L925

aws/logs_monitoring/lambda_function.py
mellon85: Append is moved to line 803, instead of having many possible places in which we append to `normalized` it's done only once at the end of the loop

aws/logs_monitoring/lambda_function.py
mellon85: The code was put after the normalization merging as the metadata overrides the DD_HOST of reach log, not just the one with the `arn` that we parse.
AFAIK the various other logs in `events` in `normalize_events` don't need to be from the same host and so setting it in the metadata at this point it will assign the host to many logs

aws/logs_monitoring/lambda_function.py
```diff
@@ -525,6 +530,30 @@ def extract_ddtags_from_message(event):
...
+    >>> )
+    i-0123456789abcdef0
+    >>> extract_arn_hostname({"message": "test"})
+    None
+    """
+    if message is not None:
+        if isinstance(message, str):
+            try:
+                message = json.loads(message)
+            except json.JSONDecodeError:
```
tianchu, line 546: I would also add some debugging logs in this function to help troubleshooting when needed, e.g., https://github.com/DataDog/datadog-serverless-functions/blob/3d061da905ffaa50acecce06f7f6f0c5dde3acf7/aws/logs_monitoring/lambda_function.py#L405

aws/logs_monitoring/lambda_function.py
tianchu: I see, that makes sense since it depends on how aws writes the trails to s3 objects, i guess there is no guarantee the events in the same s3 object would be from the same host. 

But I would move these code to `enrich` where we already been doing something similar https://github.com/DataDog/datadog-serverless-functions/blob/3d061da905ffaa50acecce06f7f6f0c5dde3acf7/aws/logs_monitoring/lambda_function.py#L479-L488 For simplicity, I would also rename `extract_arn_hostname` to `extract_host_from_cloudtrails` (or something similar) and include all the checks, e.g., `if normalized_event.get(DD_SOURCE) == "cloudtrail":` all into that one single helper function.

```python
for event in events: 
    add_metadata_to_lambda_log(event) 
    extract_ddtags_from_message(event) 
    extract_host_from_cloudtrails(event)
```

tianchu (approved):
👍

PR #434, opened 2021-03-23
Add tests, docs for log filtering

nhinsch:
### What does this PR do?

- Expand docs for log filtering, including adding some example regular expressions
- Add unit tests for log filtering. A customer having trouble with log filtering could set up their own unit test of their environment variables against a given input.
- Remove a print statement that shouldn't be in `forward_logs()`

### Motivation

We want to make it easier to set up log filtering and see common use cases.

### Testing Guidelines

Unit/integration tests.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

- [x] This PR impacts documentation, and it has been updated (or a ticket has been logged)

### DIFF
--- a/aws/logs_monitoring/README.md
+++ b/aws/logs_monitoring/README.md
@@ -299,8 +299,11 @@ Some examples of regular expressions that can be used for log filtering:
 - Include (or exclude) Lambda platform logs: `^(START|REPORT|END)`
 - Include CloudTrail error messages only: `errorMessage`
 - Include only logs containing an HTTP 4XX or 5XX error code: `\b[4|5][0-9][0-9]\b`
+- Include only CloudWatch logs where the `message` field contains a specific JSON key/value pair: `\\"awsRegion\\":\\"us-east-1\\"`
+  - The message field of a CloudWatch log event is encoded as a string. `{awsRegion: "us-east-1"}` is encoded as `{\"awsRegion\":\"us-east-1\"}`.
+    The pattern you provide must therefore include extra `\` escape characters.
 
-To debug these regular expressions against your logs, turn on [debug logs](#troubleshooting).
+To test different patterns against your logs, turn on [debug logs](#troubleshooting).
 
 ### Advanced (Optional)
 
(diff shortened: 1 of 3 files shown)


aws/logs_monitoring/README.md
```diff
@@ -287,11 +287,19 @@ The Datadog Forwarder is signed by Datadog. If you would like to verify the inte
...
+Filtering rules are applied to the full JSON-formatted log, including any metadata that is automatically added by the Forwarder.
+However, transformations applied by [log pipelines](https://docs.datadoghq.com/logs/processing/pipelines/),
+which occur after logs are sent to Datadog, cannot be used to filter logs in the Forwarder.
+Using an inefficient regular expression, such as `.*`, may slow down the Forwarder.
+
+Some examples of regular expressions that can be used for log filtering:
+- Include (or exclude) Lambda platform logs: `^(START|REPORT|END)`
+- Include CloudTrail error messages only: `errorMessage`
+- Include only logs containing an HTTP 4XX or 5XX error code: `\b[4|5][0-9][0-9]\b`
+
```
nhinsch, line 305: I thought of these three common use cases for log filtering but would definitely be interested in adding a few more if anyone can think of any others that tend to come up.

tianchu: Can we add one example where the json formatting makes it difficult to come up the filter pattern?

aws/logs_monitoring/logs.py
```diff
@@ -96,36 +99,33 @@ def compileRegex(rule, pattern):
...
+        if exclude_pattern is not None or include_pattern is not None:
             logger.debug("Filtering log event:")
             logger.debug(log)
         try:
-            if EXCLUDE_AT_MATCH is not None:
+            if exclude_pattern is not None:
                 # if an exclude match is found, do not add log to logs_to_send
-                logger.debug(f"Applying EXCLUDE_AT_MATCH: {EXCLUDE_AT_MATCH}")
+                logger.debug(f"Applying exclude pattern: {exclude_pattern}")
+                exclude_regex = compileRegex("EXCLUDE_AT_MATCH", exclude_pattern)
```
nhinsch, line 119: I moved the regex compilation into this function to make it more testable, rather than relying on global variables for the compiled regex. This should not negatively impact performance because even though the compile function is called for every log line, the `re` module automatically caches the last 512 compiled regular expressions.

Per the `re` docs: "The compiled versions of the most recent patterns passed to re.compile() and the module-level matching functions are cached, so programs that use only a few regular expressions at a time needn’t worry about compiling regular expressions."

([Here](https://github.com/python/cpython/blob/master/Lib/re.py#L284) is the code for the cache)

tianchu (approved):

PR #448, opened 2021-04-22
Split SH events into separate events based on findings

NoisomePossum:
### What does this PR do?

Splits the findings array in Security Hub events and creates a separate log for each element in findings.

### Motivation

Currently the findings attribute in Security Hub logs comes in as an array of objects which can't be parsed out on the logs side. Since this attribute contains a lot of the most relevant information about these logs we need to convert it from an array of objects into a series of single objects spread across multiple logs.

### Testing Guidelines

Tested on Datadog by sending Security Hub logs through the Lambda Function (triggered by an event rule).

### Additional Notes

Unit tests need to be added pending review.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -23,7 +23,10 @@ from enhanced_lambda_metrics import (
     parse_and_submit_enhanced_metrics,
 )
 from logs import forward_logs
-from parsing import parse
+from parsing import (
+    parse,
+    separate_security_hub_findings,
+)
 from telemetry import (
     DD_FORWARDER_TELEMETRY_NAMESPACE_PREFIX,
     get_forwarder_telemetry_tags,
@@ -85,7 +88,7 @@ def datadog_forwarder(event, context):
     if DD_ADDITIONAL_TARGET_LAMBDAS:
         invoke_additional_target_lambdas(event)
 
-    metrics, logs, trace_payloads = split(enrich(parse(event, context)))
+    metrics, logs, trace_payloads = split(transform(enrich(parse(event, context))))
 
     if DD_FORWARD_LOG:
         forward_logs(logs)
@@ -176,6 +179,22 @@ def extract_trace_payload(event):
         return {"message": message, "tags": event[DD_CUSTOM_TAGS]}
     except Exception:
         return None
+
+
+def transform(events):
+    """Performs transformations on complex events
+
+    Ex: handles special cases with nested arrays of JSON objects
+    Args:
+        events (dict[]): the list of event dicts we want to transform
+    """
+    for event in reversed(events):
+        findings = separate_security_hub_findings(event)
+        if findings:
+            events.remove(event)
+            events.extend(findings)
+
+    return events
 
 
 def enrich(events):
--- a/aws/logs_monitoring/parsing.py
+++ b/aws/logs_monitoring/parsing.py
@@ -7,6 +7,7 @@ import base64
 import gzip
 import json
 import os
+import copy
 
 import boto3
 import botocore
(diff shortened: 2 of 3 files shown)


aws/logs_monitoring/lambda_function.py
```diff
@@ -184,10 +187,19 @@ def enrich(events):
...
         extract_ddtags_from_message(event)
         extract_host_from_cloudtrails(event)
+        event = separate_security_hub_findings(event)
+
+        if isinstance(event, list):
+            for item in event:
+                prepped_events.append(item)
+
+    for event in prepped_events:
+        events.append(event)
```
tianchu:
```suggestion
    events_to_remove = []
    events_to_append = []
    for event in events:
        add_metadata_to_lambda_log(event)
        extract_ddtags_from_message(event)
        extract_host_from_cloudtrails(event)
        findings = separate_security_hub_findings(event)
        if findings:
            events_to_remove.append(event)
            events_to_append.extend(findings)

    for event in events_to_remove:
        events.remove(event)
        
    for event in events_to_append:
        events.append(event)
```

aws/logs_monitoring/parsing.py
```diff
@@ -527,6 +528,47 @@ def cwevent_handler(event, metadata):
     yield data
 
 
+# Extract findings from Security Hub logs
```
tianchu: I think you should use docstring in this case like this https://github.com/DataDog/datadog-serverless-functions/blob/3d962e943bffffd658cb289e09c1357273e170b2/aws/logs_monitoring/lambda_function.py#L79-L81.

aws/logs_monitoring/parsing.py
```diff
@@ -527,6 +528,47 @@ def cwevent_handler(event, metadata):
...
+# and move them to the root of the log
+def separate_security_hub_findings(event):
+    if (
+        event.get(DD_SOURCE) == "securityhub"
+        and event.get("detail", {}).get("findings") is not None
```
tianchu: `is not None` seems redundant?

aws/logs_monitoring/parsing.py
```diff
@@ -527,6 +528,47 @@ def cwevent_handler(event, metadata):
...
+                        current_resource = item
+                        # Capture the type and use it as the distinguishing key
+                        resource_type = current_resource.get("Type", {})
+                        del current_resource["Type"]
+                        new_event["detail"]["finding"]["resources"][
+                            resource_type
+                        ] = current_resource
+                events.append(new_event)
+        return events
+    else:
```
tianchu: By default python functions return `None`, so this else is redundant. If you do prefer explicit returns (which is a good practice), the common practice is to do an early return to make the code easier to read.

```python
if event.get(DD_SOURCE) != "securityhub" or not event.get("detail", {}).get("findings"):
    return None

events = []
event_copy = copy.deepcopy(event)
... 
```

This also helps reduce the indentation, which makes the code easier to read.

aws/logs_monitoring/parsing.py
NoisomePossum: It is in most cases. This is just to cover the case of customers sending a non-standard Security Hub log (one with no findings). In that case, we would be doing extra work to remove the original log in the events array only to replace it with itself. I guess it's a question of whether that savings is worth a cumbersome if statement.

NoisomePossum: Ah, sorry, I guess you mean to rewrite it simply as:

```
and event.get("detail", {}).get("findings")
```

I agree that's better. 👍

aws/logs_monitoring/lambda_function.py
```diff
@@ -189,6 +192,12 @@ def enrich(events):
         extract_ddtags_from_message(event)
         extract_host_from_cloudtrails(event)
 
+    for event in reversed(events):
```
nhinsch: It seems odd to me for this very Security Hub-specific logic to live in the top-level `enrich()` function. Can we move this logic to the parsing file somehow?

nhinsch: Also, why are the events reversed here?

aws/logs_monitoring/parsing.py
nhinsch: Big +1 on this comment -- let's see if we can reduce the indentation here by returning early

aws/logs_monitoring/lambda_function.py
tianchu: I made this suggestion, to simplify the code significantly. Iteration in the reversed order made `remove` and `extend` events  in the loop possible (no side effect).
 
Perhaps @NoisomePossum you can add a comment to explain?

tianchu (approved):
LGTM!

aws/logs_monitoring/lambda_function.py
NoisomePossum: Yeah, looping through in reverse order makes it possible to remove each event as we're iterating over it - because you need to replace the original event with a series of events where the data is split between them. Otherwise as you're going forward you end up skipping events when you remove one and then continue iterating.

I don't mind changing this to pass the entire events array into the function to perform the loop in there. However, I was also thinking that there are other logs that have a similar issue with arrays of objects (WAF logs come to mind as one that customers reach out about frequently). In that case, would it be better to keep this reverse loop in case we want to add another function to handle other similar logs? This is actually something I wouldn't mind working on myself so it could end up getting added relatively soon.

nhinsch: That makes sense -- I still think there's some value in "hiding" this loop in a function so that it's not right in the main `enrich()` function. Generally when we can abstract details of how things work into lower-level functions, it will make the code of the top-level functions easier to read. It sounds like there might be more than one use for this logic (i.e. not just Security Hub findings), so maybe the lower-level function that reverses the events and separates them could be extended in the future to handle more than one use case.

I also would look into whether this logic belongs inside the `enrich()` function, regardless of whether its abstracted into a lower-level function call or not. The enrich function already does a lot, and the description says "Adds event-specific tags and attributes to each event" -- what's happening here seems like something somewhat different, splitting an event into multiple smaller events. Maybe this logic should happen *after* enrichment, not *during* enrichment.

tianchu: Perhaps we could add an `extract` (or `transform`?) function after the `enrich` like `extract(enrich(...)`, and move these logic to the new function?

tianchu: What do you think @nhinsch ?

NoisomePossum: Went ahead and updated this with a `transform` function.

tianchu (approved):
The changes with `transform` look good to me.

nhinsch (approved):
I also like the new `transform` function! Much cleaner!

PR #451, opened 2021-05-05
Add kms encryption

cegan-pbp:
### What does this PR do?

Adds a flag to the log forwarder that encrypts the forwarder bucket object during installation.

### Motivation

An SCP deployed to our accounts prevents any s3 uploads that aren't KMS encrypted so we're unable to install the forwarder there without modifying the cloudformation template.

In our own environment, with both the flag enabled and disabled.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/template.yaml
+++ b/aws/logs_monitoring/template.yaml
@@ -187,6 +187,10 @@ Parameters:
     Type: String
     Default: ""
     Description: The endpoint URL to forward the traces to, useful for forwarding traces through a proxy
+  DdForwarderBucketKMSEncryptionEnabled:
+    Type: String
+    Default: false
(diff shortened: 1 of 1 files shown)


nhinsch:
Hi @cegan-pbp, thanks for your contribution! 🙂 We're not going to merge this PR at the moment for a few reasons:
- This isn't a request we've seen a lot from customers.
- There are a number of different ways to encrypt S3 objects, and different parameters that could be applied. What is allowed in your AWS account might be different from what is allowed in other AWS accounts. It might be preferable for customers to modify the CloudFormation template or manually install the Forwarder to suit their specific needs.
- Generally, we want to limit the number of configuration options we add to the installation process.

With that said, if we observe increased demand for this feature, we may want to revisit this, so thanks very much for putting the issue on our radar!

PR #456, opened 2021-05-13
[azure] Enable batching of requests in log forwarder and optional TCP support

claudiadadamo:
### What does this PR do?
Adds in batching for the HTTP requests, to group together logs instead of send logs individually as separate HTTP requests. These requests should be executing in parallel anyway (since using Promise.all) but it should hopefully help a bit to group them all into fewer requests overall.

Also adds in optional TCP support with all of the new logic that's been added since it was moved to only work with HTTP. TCP is much faster than HTTP because the data is streamed, versus with HTTP the data gets persisted and a 200 response means that we've acknowledged and stored the log data on our end. The tradeoff is that with TCP you don't have guarantees that we've received the message.

### Motivation

Try to speed up the HTTP log forwarder, as well as allow customers to use TCP if they want

### Testing Guidelines

Tested with my own version of the forwarder, made sure that logs are flowing in with both versions. Also updated unit tests for some of the batching logic.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/azure/activity_logs_monitoring/index.js
+++ b/azure/activity_logs_monitoring/index.js
@@ -82,12 +82,7 @@ class ScrubberRule {
 }
 
 class Batcher {
-    constructor(
-        context,
-        maxItemSizeBytes,
-        maxBatchSizeBytes,
-        maxItemsCount
-    ) {
+    constructor(context, maxItemSizeBytes, maxBatchSizeBytes, maxItemsCount) {
         this.maxItemSizeBytes = maxItemSizeBytes;
         this.maxBatchSizeBytes = maxBatchSizeBytes;
         this.maxItemsCount = maxItemsCount;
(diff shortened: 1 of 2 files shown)


azure/activity_logs_monitoring/index.js
```diff
@@ -77,6 +81,164 @@ class ScrubberRule {
...
+            256 * 1000,
+            4 * 1000 * 1000,
+            400
+        );
+    }
+
+    async sendAll(records) {
+        var batches = this.batcher.batch(records);
+        var promises = [];
+        for (var i = 0; i < batches.length; i++) {
```
mattsp1290, line 156: *nit: (extremely minor) seeing a mix of forEach and traditional fors for iterating over lists.

mattsp1290 (approved):
lgtm

mattsp1290 (approved):
lgtm

mattsp1290 (approved):
lgtm

PR #465, opened 2021-06-17
Slim cloudformation install flow

DarcyRaynerDD:
### What does this PR do?

Creates a new install flow for the forwarder using lambda layers, instead of relying on a second ZipCopier function. It makes this new install flow the default, but also keeps the ZipCopier flow as an option for regions where we can't publish lambda layers, (such as the regions in China). It also updates the installation_test script to validate both install flows.

### Motivation

The ZipCopier function, which copies an external zip from github into a local s3 bucket was proving difficult for some customers to use in restrictive AWS environments, (such as where VPC usage is mandated for all functions). This alternative flow  reduces the number of resources installed by the template by packaging code into a lambda layer.

### Testing Guidelines
You can run the ./tools/installation_test.sh script.

### Types of changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [X] Misc (docs, refactoring, dependency upgrade, etc.)

### Check all that apply

- [X] This PR contains breaking changes that are documented in the description
- [X] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future
- [X] This PR impacts documentation, and it has been updated (or a ticket has been logged)
- [X] This PR's changes are covered by the automated tests

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,7 +27,10 @@ jobs:
           pip install black
           black --check ./aws/logs_monitoring
 
-      - name: Lint CloudFormation template with cfn-lint
-        uses: scottbrenner/cfn-lint-action@master
-        with:
-          args: "aws/logs_monitoring/template.yaml"
+      - name: Setup Cloud Formation Linter with Latest Version
+        uses: scottbrenner/cfn-lint-action@v2
+
+      - name: Print the Cloud Formation Linter Version & run Linter.
+        run: |
+          cfn-lint --version
+          cfn-lint -t aws/logs_monitoring/template.yaml
--- a/aws/logs_monitoring/README.md
+++ b/aws/logs_monitoring/README.md
@@ -365,8 +365,8 @@ To test different patterns against your logs, turn on [debug logs](#troubleshoot
 
 `AdditionalTargetLambdaARNs`
 : Comma separated list of Lambda ARNs that will get called asynchronously with the same `event` the Datadog Forwarder receives.
-`InstallFromLayer`
-: Disable to use legacy installation flow, (via Github), for regions where Lambda layers aren't supported, such as regions in China. Defaults to true.
+`InstallAsLayer`
+: Whether to use the layer based installation flow. Set to false to use our legacy installation flow, which installs a second function that copies the forwarder code from Github to an S3 bucket. Disable this option if installing into AWS China regions, where we don't publish layers. Defaults to true.
 `LayerARN`
 : ARN for the layer containing the forwarder code. If empty, the script will use the version of the layer the forwarder was published with. Defaults to empty.
 
--- a/aws/logs_monitoring/README.md
+++ b/aws/logs_monitoring/README.md
@@ -366,7 +366,7 @@ To test different patterns against your logs, turn on [debug logs](#troubleshoot
 `AdditionalTargetLambdaARNs`
 : Comma separated list of Lambda ARNs that will get called asynchronously with the same `event` the Datadog Forwarder receives.
 `InstallAsLayer`
-: Whether to use the layer based installation flow. Set to false to use our legacy installation flow, which installs a second function that copies the forwarder code from Github to an S3 bucket. Disable this option if installing into AWS China regions, where we don't publish layers. Defaults to true.
+: Whether to use the layer-based installation flow. Set to false to use our legacy installation flow, which installs a second function that copies the forwarder code from Github to an S3 bucket. Defaults to true.
 `LayerARN`
 : ARN for the layer containing the forwarder code. If empty, the script will use the version of the layer the forwarder was published with. Defaults to empty.
 
--- a/aws/logs_monitoring/README.md
+++ b/aws/logs_monitoring/README.md
@@ -365,6 +365,10 @@ To test different patterns against your logs, turn on [debug logs](#troubleshoot
 
 `AdditionalTargetLambdaARNs`
 : Comma separated list of Lambda ARNs that will get called asynchronously with the same `event` the Datadog Forwarder receives.
+`InstallFromLayer`
+: Disable to use legacy installation flow, (via Github), for regions where Lambda layers aren't supported, such as regions in China. Defaults to true.
+`LayerARN`
+: ARN for the layer containing the forwarder code. If empty, the script will use the version of the layer the forwarder was published with. Defaults to empty.
 
 ## Permissions
 
--- a/aws/logs_monitoring/release.sh
+++ b/aws/logs_monitoring/release.sh
@@ -4,6 +4,8 @@
 
 set -e
 
+LAYER_NAME="Datadog-Forwarder"
+
 # Read the current version
 CURRENT_VERSION=$(grep -o 'Version: \d\+\.\d\+\.\d\+' template.yaml | cut -d' ' -f2)
 
@@ -39,8 +41,24 @@ if [ "$ACCOUNT" = "prod" ]; then
     BUCKET="datadog-cloudformation-template"
 fi
 
+get_max_layer_version() {
+    last_layer_version=$(aws lambda list-layer-versions --layer-name $LAYER_NAME --region us-west-2 | jq -r ".LayerVersions | .[0] |  .Version")
+    if [ "$last_layer_version" == "null" ]; then
+        echo 0
+    else
+        echo $last_layer_version
+    fi
+}
+
 # Validate identity
 aws sts get-caller-identity
+
+CURRENT_ACCOUNT="$(aws sts get-caller-identity --query Account --output text)"
+CURRENT_VERSION=$(get_max_layer_version)
+NEXT_LAYER_VERSION=$(($CURRENT_VERSION +1))
+
+echo
+echo "Current layer version is $CURRENT_VERSION, next layer version is $NEXT_LAYER_VERSION"
 
 # Validate the template
 echo
@@ -76,6 +94,7 @@ if [ "$ACCOUNT" = "prod" ] ; then
     echo "Bumping the version number to ${VERSION}..."
     perl -pi -e "s/DD_FORWARDER_VERSION = \"[0-9\.]+/DD_FORWARDER_VERSION = \"${VERSION}/g" settings.py
     perl -pi -e "s/Version: [0-9\.]+/Version: ${VERSION}/g" template.yaml
+    perl -pi -e "s/LayerVersion: [0-9\.]+/LayerVersion: ${NEXT_LAYER_VERSION}/g" template.yaml
 
     # Commit version number changes to git
     echo "Committing version number change..."
@@ -98,6 +117,11 @@ if [ "$ACCOUNT" = "prod" ] ; then
     echo "Releasing aws-dd-forwarder-${VERSION} to GitHub..."
     go get github.com/github/hub
     hub release create -a $BUNDLE_PATH -m "aws-dd-forwarder-${VERSION}" aws-dd-forwarder-${VERSION}
+    
+    # Upload the sandbox layers
+    echo
+    echo "Uploading layers"
+    ./tools/publish_prod.sh $NEXT_LAYER_VERSION $VERSION
 
     # Set vars for use in the installation test
     TEMPLATE_URL="https://${BUCKET}.s3.amazonaws.com/aws/forwarder/latest.yaml"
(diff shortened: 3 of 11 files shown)


aws/logs_monitoring/release.sh
```diff
@@ -98,6 +117,11 @@ if [ "$ACCOUNT" = "prod" ] ; then
     echo "Releasing aws-dd-forwarder-${VERSION} to GitHub..."
     go get github.com/github/hub
     hub release create -a $BUNDLE_PATH -m "aws-dd-forwarder-${VERSION}" aws-dd-forwarder-${VERSION}
+    
+    # Upload the sandbox layers
```
nhinsch: I think this comment is wrong -- should be prod

aws/logs_monitoring/tools/build_bundle.sh
```diff
@@ -34,12 +34,13 @@ function make_path_absolute {
 
 function docker_build_zip {
     # Args: [python version] [zip destination]
-
     destination=$(make_path_absolute $2)
```
nhinsch: can we change `destination` to `zip_destination`?

aws/logs_monitoring/tools/build_bundle.sh
```diff
@@ -51,13 +52,25 @@ function docker_build_zip {
...
 
     rm -rf $temp_dir
-    echo "Done creating archive $destination"
+    echo "Done creating forwarder zip archive $destination"
+
+    temp_dir=$(mktemp -d)
+    SUB_DIRECTORY=python
+    mkdir $temp_dir/$SUB_DIRECTORY
+
+    # Run the image by runtime tag, tar its generatd `python` directory to sdout,
```
nhinsch: Nit: generatd -> generated

aws/logs_monitoring/tools/list_layers.sh
```diff
@@ -0,0 +1,68 @@
...
+    if [[ ! "$AVAILABLE_REGIONS" == *"$2"* ]]; then
+        >&2 echo "Could not find $2 in available regions:" $AVAILABLE_REGIONS
+        >&2 echo ""
+        >&2 echo "EXITING SCRIPT."
+        exit 1
+    fi
+    REGIONS=($2)
+fi
+
+# Check region arg
```
nhinsch, line 33: This comment talks about a "region" arg, but I think the code refers to the first arg ("layer-name"). However, there is only one possible layer name so maybe that argument and all of this code can be removed?

aws/logs_monitoring/tools/publish_layers.sh
```diff
@@ -0,0 +1,145 @@
+#!/bin/bash
+
+# Unless explicitly stated otherwise all files in this repository are licensed
+# under the Apache License Version 2.0.
+# This product includes software developed at Datadog (https://www.datadoghq.com/).
+# Copyright 2019 Datadog, Inc.
+
+# Publish the datadog python lambda layer across regions, using the AWS CLI
```
nhinsch: Looks like this was copypasted from the Python layer, should be updated for the Forwarder

aws/logs_monitoring/tools/publish_layers.sh
```diff
@@ -0,0 +1,145 @@
...
+    echo "Region specified: $REGIONS"
+    if [[ ! "$AVAILABLE_REGIONS" == *"$REGIONS"* ]]; then
+        echo "Could not find $REGIONS in available regions: $AVAILABLE_REGIONS"
+        echo ""
+        echo "EXITING SCRIPT."
+        exit 1
+    fi
+fi
+
+# Determine the target layers
```
nhinsch, line 44: Likewise only one possible layer here

.github/workflows/lint.yml
```diff
@@ -27,7 +27,10 @@ jobs:
           pip install black
           black --check ./aws/logs_monitoring
 
-      - name: Lint CloudFormation template with cfn-lint
-        uses: scottbrenner/cfn-lint-action@master
-        with:
-          args: "aws/logs_monitoring/template.yaml"
+      - name: Setup Cloud Formation Linter with Latest Version
```
nhinsch, line 30: Looks like this doesn't necessarily use the latest version (it will always use version 2). I would just stick to "Setup Cloud Formation Linter"

aws/logs_monitoring/README.md
```diff
@@ -365,6 +365,10 @@ To test different patterns against your logs, turn on [debug logs](#troubleshoot
 
 `AdditionalTargetLambdaARNs`
 : Comma separated list of Lambda ARNs that will get called asynchronously with the same `event` the Datadog Forwarder receives.
+`InstallFromLayer`
+: Disable to use legacy installation flow, (via Github), for regions where Lambda layers aren't supported, such as regions in China. Defaults to true.
```
nhinsch: I would reword this, I don't think it's obvious what `legacy installation flow, (via Github)` entails.

nhinsch: Also, I think "regions where Lambda layers aren't supported, such as regions in China" is a bit misleading. It looks like Lambda layers are supported in China, but we just don't publish our layers there.

aws/logs_monitoring/README.md
```diff
@@ -365,6 +365,10 @@ To test different patterns against your logs, turn on [debug logs](#troubleshoot
 
 `AdditionalTargetLambdaARNs`
 : Comma separated list of Lambda ARNs that will get called asynchronously with the same `event` the Datadog Forwarder receives.
+`InstallFromLayer`
```
nhinsch: I would rename this to `InstallAsLayer`. `InstallFromLayer` makes it sound like we are somehow copying something off of a layer.

aws/logs_monitoring/template.yaml
```diff
@@ -52,6 +53,17 @@ Parameters:
     Type: String
     Default: ""
     Description: DO NOT CHANGE unless you know what you are doing. Override the default location of the function source code.
+  InstallFromLayer:
```
nhinsch: I left some comments about these parameters on the README file, but those comments apply to the template as well.

aws/logs_monitoring/README.md
```diff
@@ -365,6 +365,10 @@ To test different patterns against your logs, turn on [debug logs](#troubleshoot
...
+: Disable to use legacy installation flow, (via Github), for regions where Lambda layers aren't supported, such as regions in China. Defaults to true.
+`LayerARN`
+: ARN for the layer containing the forwarder code. If empty, the script will use the version of the layer the forwarder was published with. Defaults to empty.
```
nhinsch, line 372: These docs make it sound like the `InstallFromLayer` option determines whether or not we install from the layer. But looking at the template, it seems that we automatically don't install from the layer when we detect the installation is happening in AWS China or the `SourceZipUrl` parameter is set.

Also, why do we assume that the customer wants to use a Zip if they set a `SourceZipUrl`, but we don't necessarily assume they want to use a layer if they set `LayerARN`?

aws/logs_monitoring/README.md
nhinsch: We could also flip it and make this `InstallAsZip`? With a default of `false`?

aws/logs_monitoring/tools/list_layers.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
```
nhinsch, line 1: Are you sure that the `list_layers.sh` and `add_new_region.sh` scripts are needed? It looks to me like the `publish_layers.sh` script uses all available regions.

aws/logs_monitoring/tools/list_layers.sh
DarcyRaynerDD: This file is pretty much copy pasted from the python layer. I think it will make it easier to port changes back if we avoid unnecessary changes. Wondering if it might make sense to abstract our layer tooling into a reusable utility that we share between codebases, and only keep settings like LAYER_NAME in the local repo.

.github/workflows/lint.yml
DarcyRaynerDD: It's v2 of the github action, but I think it does use the latest version of the linter, at least based on the actions documentation. https://github.com/ScottBrenner/cfn-lint-action#upgrading-from-version-1

aws/logs_monitoring/tools/publish_layers.sh
```diff
@@ -0,0 +1,145 @@
...
+# Unless explicitly stated otherwise all files in this repository are licensed
+# under the Apache License Version 2.0.
+# This product includes software developed at Datadog (https://www.datadoghq.com/).
+# Copyright 2019 Datadog, Inc.
+
+# Publish the datadog python lambda layer across regions, using the AWS CLI
+# Usage: VERSION=5 REGIONS=us-east-1 LAYERS=Datadog-Python27 publish_layers.sh
+# VERSION is required.
+set -e
+
```
nhinsch, line 12: It looks like this script requires `VERSION` and `FORWARDER_VERSION`. The Forwarder version seems like a semantic version number whereas version is the layer version. However, the usage of these two terms isn't consistent across all of the scripts.

Could we switch to using `LAYER_VERSION` and `FORWARDER_VERSION` everywhere, so it is easy to keep straight which is which in all of the scripts?

aws/logs_monitoring/tools/list_layers.sh
nhinsch: Yeah, a reusable layer deploy tool could be a smart idea.

In the mean time, I buy that it's reasonable to leave this extra argument here (I don't think it does any harm). Some of my other comments still apply, I think there's potential for confusion with some of the copy-pasted stuff.

aws/logs_monitoring/tools/list_layers.sh
DarcyRaynerDD: list_layers is useful for local development, even if it isn't strictly needed during the publish flow. add_new_region is useful because it grabs a current layer zip and redeploys that, without having to go back and rebuild from the last release commit.

aws/logs_monitoring/README.md
DarcyRaynerDD: I could add a check the the China condition, to see if LayerARN was manually set. The reason for having the China condition there is so user's won't have to think decide which flow to use by default. I can probably remove the SourceZipUrl condition.

aws/logs_monitoring/README.md
```diff
@@ -365,6 +365,10 @@ To test different patterns against your logs, turn on [debug logs](#troubleshoot
 
 `AdditionalTargetLambdaARNs`
 : Comma separated list of Lambda ARNs that will get called asynchronously with the same `event` the Datadog Forwarder receives.
+`InstallAsLayer`
+: Whether to use the layer based installation flow. Set to false to use our legacy installation flow, which installs a second function that copies the forwarder code from Github to an S3 bucket. Disable this option if installing into AWS China regions, where we don't publish layers. Defaults to true.
```
nhinsch: Nits:
layer based -> layer-based
Github -> GitHub

Also, this says "Disable this option if installing into AWS China regions, where we don't publish layers." However, looking at the template, it looks like you don't actually need to disable this option when publishing to China -- it happens automatically.

aws/logs_monitoring/release.sh
```diff
@@ -63,7 +81,7 @@ if [ "$ACCOUNT" = "prod" ] ; then
 
     # Confirm to proceed
     echo
-    read -p "About to bump the version from ${CURRENT_VERSION} to ${VERSION}, create a release aws-dd-forwarder-${VERSION} on GitHub and upload the template.yaml to s3://${BUCKET}/aws/forwarder/${VERSION}.yaml. Continue (y/n)?" CONT
+    read -p "About to bump the version from ${CURRENT_VERSION} to ${FORWARDER_VERSION}, create a release aws-dd-forwarder-${FORWARDER_VERSION} on GitHub and upload the template.yaml to s3://${BUCKET}/aws/forwarder/${FORWARDER_VERSION}.yaml. Continue (y/n)?" CONT
```
nhinsch: Should we update this message to also say that we are publishing a new layer?

nhinsch (approved):
Looks great! Just a couple minor nits

aws/logs_monitoring/tools/publish_layers.sh
nhinsch: This comment should still be updated

PR #472, opened 2021-07-02
[aws] use the service from the dd_tags if it exists

guillaumelecerf:
### What does this PR do?

When we send ELB logs to Datadog, we end up having `source: elb` and `service: elb`, hence we cannot differentiate the logs from service X from those from service Y.

This PR changes the parsing so that if the tags contain `service:X`, it is used for DD_SERVICE.

### Motivation

Being able to filter the logs by service.

### Testing Guidelines

Tested by manually uploading the new code to our log forwarding AWS Lambda

### Types of changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Misc (docs, refactoring, dependency upgrade, etc.)

### Check all that apply

- [x] This PR's description is comprehensive
- [ ] This PR contains breaking changes that are documented in the description
- [ ] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future
- [ ] This PR impacts documentation, and it has been updated (or a ticket has been logged)
- [ ] This PR's changes are covered by the automated tests
- [ ] This PR collects user input/sensitive content into Datadog
- [ ] This PR passes the integration tests (ask a Datadog member to run the tests)
- [ ] This PR passes the unit tests 
- [ ] This PR passes the installation tests (ask a Datadog member to run the tests)

Issue #269, opened 2020-06-01
Feature Request: Apply Cloudwatch Log Group Tags as Datadog Tags


kellydunn:
Hello there!

We love the capabilities of the Data Logging Pipeline, and we use it quite heavily today!  One feature we've always enjoyed is the nice UX of installing an agent and having it aggregate EC2 Tagging metadata as tags in Datadog's Logging product.

We also make heavy use of the Datadog Log Forwader for our ECS Workloads, however, we've noticed that some of the metadata we used to get in EC2 isn't necessarily working how we'd expect with Cloudwatch log groups.

What would be really nice is if this same behavior applied to Cloudwatch Log Groups so we wouldn't have to encode aggregation data into our Structured Logs and try to parse it out via the Logging Pipeline.

For example, if we tag a Cloudwatch log group resource in AWS with

```
Env: 'production'
Service: 'auth'
```

We would love to have any logs from that log group be searchable and aggregated by those tags as well!

eboulfakhr:
Hello!
We have a similar need. I added tags to the aws elasticsearch slow logs logroup but I am not able to see these tags in the the logs forwarded to datadog.

thank you!

dominics:
As an additional use case: this is the only way we'd be able to tag AWS Glue logs with unified tagging.

That's because there's no Lambda for the Forwarder to look up the tags of, and the log events themselves are formatted by the service. [Datadog's Glue docs](https://docs.datadoghq.com/integrations/amazon_glue/#send-logs-to-datadog) recommend using the Forwarder, but don't mention you can't use it with tagging.

agocs:
SLS-1325

HartmannVolker:
Would love to see that 👍  we are forwarding API Gateway access logs to datadog using the lambda forwarder. Without the service and env tag it is hard to correlated these logs with the lambda invocations.

podon:
Bump. Any traction on this request?

tianchu:
Closing. because it's done by https://github.com/DataDog/datadog-serverless-functions/pull/540

PR #473, opened 2021-07-02
[aws] allow overriding service with DD_SERVICE_NAME

guillaumelecerf:
### What does this PR do?

This PR allows to override the service name by setting DD_SERVICE_NAME environment variable.

### Motivation

When we send ELB logs to Datadog, we end up having `source: elb` and `service: elb`, hence we cannot differentiate the logs from service X from those from service Y.

This PR allows us to filter the logs by service.

### Testing Guidelines

Tested by manually uploading the new code to our log forwarding AWS Lambda

### Types of changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/README.md
+++ b/aws/logs_monitoring/README.md
@@ -290,6 +290,9 @@ SSL encrypted TCP connection, set this parameter to true.
 `DdPort`
 : The endpoint port to forward the logs to, useful for forwarding logs through a proxy.
 
+`DdServiceName`
+: Allow to override the service name of the forwarded logs.
+
 `DdSkipSslValidation`
 : Send logs over HTTPS, while NOT validating the certificate provided by the endpoint. This will still encrypt the traffic between the forwarder and the log intake endpoint, but will not verify if the destination SSL certificate is valid.
 
--- a/aws/logs_monitoring/parsing.py
+++ b/aws/logs_monitoring/parsing.py
@@ -29,6 +29,7 @@ from settings import (
     DD_SOURCE,
     DD_CUSTOM_TAGS,
     DD_SERVICE,
+    DD_SERVICE_NAME,
     DD_HOST,
     DD_FORWARDER_VERSION,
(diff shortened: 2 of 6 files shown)


agocs:
Hi @guillaumelecerf, do you intend to have one forwarder per service? That's a valid workaround in this case, but I'm not 100% sure it's the long term strategy we want to adopt. I'll create a ticket to investigate this further.

This could possibly be solved by #269 (SLS-1325)

SLS-1318 (automatically-created ticket)
SLS-1332

guillaumelecerf:
Hi @agocs,
Yes, we are currently using one forwarder per service. To be more precise: for our services using an ALB, the ALB writes its logs to an S3 bucket, whose events are consumed by the datadog Lambda function.

I understand that the approach of this PR may not fit your long term strategy, thus I wrote another one which may better suit:
#475 

I also looked at #269 but this issue refers to Cloudwatch which is, AFAIK, not involved in my setup (ALB -> S3 -> Lambda -> Datadog).

tianchu:
@guillaumelecerf thanks for your amazing contribution to our code base! I agree with @agocs that this PR does not fit in our long-term strategy. https://github.com/DataDog/datadog-serverless-functions/pull/475 does make a lot of sense, though the implementation needs to be improved for efficiency.

If you do want to make the change to the forwarder so it allows you to override `service` name at the forwarder level, I would suggest using `DdTags` option to set a `service` tag, and then make a small updates in the `xxx_handler` functions to ensure they do not override the `service` tag if already present. I think that makes more sense than adding a new `DdServiceName` parameter.

guillaumelecerf:
@tianchu : thanks a ton for your review ;)
So, IIUC, something like https://github.com/DataDog/datadog-serverless-functions/pull/472 could be acceptable for you?

tianchu:
Yeah, that one seems to be exactly what I described :)

PR #482, opened 2021-08-05
Fix deploy prod deploy script for GovCloud

DarcyRaynerDD:
### What does this PR do?

Currently the forwarder deploy script breaks when submitting to gov-cloud. This is because the script is run in a nested aws-vault session. To fix, I changed the release script so no nesting is used, and aws-vault sessions are more tightly scoped within the script.

### Testing Guidelines

I've run the installation tests to verify nothing has broken.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Misc (docs, refactoring, dependency upgrade, etc.)

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/release.sh
+++ b/aws/logs_monitoring/release.sh
@@ -41,8 +41,18 @@ if [ "$ACCOUNT" = "prod" ]; then
     BUCKET="datadog-cloudformation-template"
 fi
 
+function awe-login() {
+    cfg=$@
+    shift
+    if [ "$ACCOUNT" = "prod" ] ; then
+        aws-vault exec prod-engineering --  $cfg
+    else
+        aws-vault exec sandbox-account-admin --  $cfg
+    fi
+}
+
 get_max_layer_version() {
-    last_layer_version=$(aws lambda list-layer-versions --layer-name $LAYER_NAME --region us-west-2 | jq -r ".LayerVersions | .[0] |  .Version")
+    last_layer_version=$(awe-login aws lambda list-layer-versions --layer-name $LAYER_NAME --region us-west-2 | jq -r ".LayerVersions | .[0] |  .Version")
     if [ "$last_layer_version" == "null" ]; then
         echo 0
     else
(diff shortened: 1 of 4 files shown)


aws/logs_monitoring/release.sh
```diff
@@ -41,8 +41,18 @@ if [ "$ACCOUNT" = "prod" ]; then
     BUCKET="datadog-cloudformation-template"
 fi
 
+function awe-login() {
```
nhinsch: What does `awe` stand for here? Is this supposed to be `aws-login`?

nhinsch (approved):
Looks good, just one question.

aws/logs_monitoring/release.sh
agocs: This function doesn't work on my machine and I can't figure out why.

```
awe-login aws sts get-caller-identity
Enter token for arn:aws:iam::464622532012:mfa/chris.agocs: 629648
aws-vault: error: exec: Error execing process: exec: "aws sts get-caller-identity": executable file not found in $PATH
```

I'd feel more comfortable if we could defensively code around whatever's going on here, or if we used aws-vault directly

DarcyRaynerDD: Yeah, unfortunately we can't use aws-vault directly on release.sh, because aws-vault doesn't support nested sessions, (I tried unsetting the relevant environment variables, but that broke things even more). Can you call 'aws sts get-caller-identity' directly from bash/zsh. It's weird that it's not in your path.

DarcyRaynerDD: Ahh good point. I have this command in my local bash and I reused the name. I read it as 'aws environment'. I wanted to avoid confusion with it being an official tool like 'aws' or 'aws-vault'. I can change it though if you feel it's neater.

agocs:
```suggestion
function awe-login() {
    cfg=( "$@" )
    echo "${cfg[@]}"
    shift
    if [ "$ACCOUNT" = "prod" ] ; then
        # shellcheck disable=SC2068
        aws-vault exec prod-engineering --  ${cfg[@]}
    else
        # shellcheck disable=SC2068
        aws-vault exec sandbox-account-admin --  ${cfg[@]}
    fi
}
```

agocs: Here, this works

```
function awe-login() {
    cfg=( "$@" )
    shift
    if [ "$ACCOUNT" = "prod" ] ; then
        # shellcheck disable=SC2068
        aws-vault exec prod-engineering --  ${cfg[@]}
    else
        # shellcheck disable=SC2068
        aws-vault exec sandbox-account-admin --  ${cfg[@]}
    fi
}
```

agocs (approved):

aws/logs_monitoring/release.sh
DarcyRaynerDD: I've updated the function to use this.

PR #492, opened 2021-08-13
[azure] Fix log splitting when field value is null

claudiadadamo:
### What does this PR do?

If a field that we're trying to index into is null, the function will crash - in some cases it seems like some logs may have a field that is null instead of a dictionary, so we should return null if one of the fields is null.

### Motivation

Better support for log splitting in eventhub log forwarder

### Testing Guidelines
Tested in test account, confirmed with different field types that it works, that we see the exception `TypeError: Cannot read property 'runOutput' of null` before the change, and after the log is successfully sent and not split.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/azure/activity_logs_monitoring/index.js
+++ b/azure/activity_logs_monitoring/index.js
@@ -5,7 +5,7 @@
 
 var https = require('https');
 
-const VERSION = '0.5.4';
+const VERSION = '0.5.5';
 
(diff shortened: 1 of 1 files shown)


mattsp1290 (approved):
lgtm

iotmani (approved):

PR #519, opened 2021-12-15
Restrict the logic for trace detection in logs

maxday:
### What does this PR do?

- Modify the logic to better discriminate a Datadog trace in logs
- Add units tests around the modified function

Before : traces array was enough
After : traces array + first item of the array containing a 'trace_id' field

Now logs containing the trace keyword, which are NOT Datadog traces, are correctly logged.

Before:
Screen Shot 2021-12-15 at 12 52 32 PM
After:
Screen Shot 2021-12-15 at 12 53 47 PM

### Motivation

Support ticket

### Testing Guidelines

- [x] Unit testing
- [x] redeploy a forwarder in sandbox

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -175,10 +175,16 @@ def extract_trace_payload(event):
     try:
         message = event["message"]
         obj = json.loads(event["message"])
-        if not "traces" in obj or not isinstance(obj["traces"], list):
-            return None
-        return {"message": message, "tags": event[DD_CUSTOM_TAGS]}
-    except Exception:
+
+        obj_has_traces = "traces" in obj
+        traces_is_a_list = isinstance(obj["traces"], list)
+        # check that the log is not containing a traces array unrelated to Datadog
+        trace_id_found = len(obj["traces"][0]) > 0 and obj["traces"][0][0]["trace_id"] is not None
+    
+        if obj_has_traces and traces_is_a_list and trace_id_found:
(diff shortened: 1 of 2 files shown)


aws/logs_monitoring/lambda_function.py
```diff
@@ -175,9 +175,19 @@ def extract_trace_payload(event):
...
-            return None
-        return {"message": message, "tags": event[DD_CUSTOM_TAGS]}
+
+        obj_has_traces = "traces" in obj
+        traces_is_a_list = isinstance(obj["traces"], list)
+        # check that the log is not containing a trace array unrelated to Datadog
+        trace_id_found = (
+            len(obj["traces"]) > 0
+            and len(obj["traces"][0]) > 0
+            and obj["traces"][0][0]["trace_id"] is not None
```
tianchu, line 185: Could even validate the trace_id format/length?

maxday: According to this : https://docs.datadoghq.com/tracing/faq/span_and_trace_id_format 
It seems that it's gonna be a large number and I'm afraid that a too restricted test would prevent valid traces from being sent. It seems that the agent is only checking that the trace_id is not zero here : https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/agent/normalizer.go#L43

Maybe I can add a check on the `resource` field which is not allowed to be empty?

tianchu (approved):

Issue #493, opened 2021-08-18
aws-dd-forwarder-3.38.0: Update fails with lambda:GetLayerVersion AccessDenied error


martinlindner:
**Describe what happened:**
When trying to update  datadog-forwarder stack 3.31.0 -> 3.38.0 in `ap-northeast-1`, the `Forwarder` resource update fails with the following error and triggers a rollback:
```
Resource handler returned message: "User: arn:aws:sts::<redacted> is not authorized to perform: lambda:GetLayerVersion on resource: arn:aws:lambda:ap-northeast-1:464622532012:layer:Datadog-Forwarder:6 (Service: Lambda, Status Code: 403, Request ID: <redacted>, Extended Request ID: null)" (RequestToken: <redacted>, HandlerErrorCode: AccessDenied)
```

Updating 3.31.0 -> 3.37.0 in the same environment worked as expected.

**Describe what you expected:**
Successful update to 3.38.0.

**Steps to reproduce the issue:**
Try to update datadog-forwarder stack with `https://datadog-cloudformation-template.s3.amazonaws.com/aws/forwarder/3.38.0.yaml`.

tianchu:
@martinlindner I can reproduce this issue and it appears to be a problem with the lambda layer not correctly published to all the regions. For now, you can workaround the issue by setting `InstallAsLayer` to `false`.

martinlindner:
Hi @tianchu, thanks for the quick reponse.

I can confirm that the `InstallAsLayer = false` workaround does the trick for me. I'm perfectly happy with the "legacy" install flow for now, but just FYI, the layer-based installation is still borked in `ap-northeast-1` as of ~15m ago.

tianchu:
@martinlindner good to hear that the workaround worked for you! We are still looking into the layer publishing issue on our side, will keep the issue open until we find a perfect solution for everyone.

tianchu:
@martinlindner we increased the timeout setting on AWS CLI and republished the missing layers, I can now successfully install the Forwarder in Tokyo. Thanks for reporting the issue! 🙇

PR #531, opened 2022-03-04
adding log group tags

jvanbrie:
### What does this PR do?
This pr adds support for cloudwatch log group tags. These tags will now get attached to the logs sent to datadog. 

### Motivation
Customer requests

### Testing Guidelines
Created new lambda function with this code, created some custom tags on a few log groups, and confirmed those tags get attached to the logs that make their way to Datadog. Unit and integration tests pass as well.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/parsing.py
+++ b/aws/logs_monitoring/parsing.py
@@ -453,7 +453,7 @@ def awslogs_handler(event, context, metadata):
 
     metadata[DD_SERVICE] = get_service_from_tags(metadata)
 
-    cloudwatch_logs_client = boto3.client('logs')
+    cloudwatch_logs_client = boto3.client("logs")
     response = None
     try:
         response = cloudwatch_logs_client.list_tags_log_group(
@@ -463,10 +463,15 @@ def awslogs_handler(event, context, metadata):
         logger.exception(f"Failed to get log group tags due to {e}")
     if response is not None:
         formatted_tags = [
-            "{key}:{value}".format(key=k, value=v) if v else k for k,v in response["tags"].items()
+            "{key}:{value}".format(key=k, value=v) if v else k
+            for k, v in response["tags"].items()
         ]
         if len(formatted_tags) > 0:
-            metadata[DD_CUSTOM_TAGS] = ",".join(formatted_tags) if not metadata[DD_CUSTOM_TAGS] else metadata[DD_CUSTOM_TAGS] + "," + ",".join(formatted_tags)
+            metadata[DD_CUSTOM_TAGS] = (
+                ",".join(formatted_tags)
+                if not metadata[DD_CUSTOM_TAGS]
+                else metadata[DD_CUSTOM_TAGS] + "," + ",".join(formatted_tags)
+            )
 
     # Build aws attributes
     aws_attributes = {
--- a/aws/logs_monitoring/parsing.py
+++ b/aws/logs_monitoring/parsing.py
@@ -453,6 +453,21 @@ def awslogs_handler(event, context, metadata):
 
     metadata[DD_SERVICE] = get_service_from_tags(metadata)
 
+    cloudwatch_logs_client = boto3.client('logs')
+    response = None
+    try:
+        response = cloudwatch_logs_client.list_tags_log_group(
+            logGroupName=logs["logGroup"]
+        )
+    except Exception as e:
+        logger.exception(f"Failed to get log group tags due to {e}")
+    if response is not None:
+        formatted_tags = [
+            "{key}:{value}".format(key=k, value=v) if v else k for k,v in response["tags"].items()
+        ]
+        if len(formatted_tags) > 0:
+            metadata[DD_CUSTOM_TAGS] = ",".join(formatted_tags) if not metadata[DD_CUSTOM_TAGS] else metadata[DD_CUSTOM_TAGS] + "," + ",".join(formatted_tags)
+
     # Build aws attributes
     aws_attributes = {
         "aws": {
--- a/aws/logs_monitoring/template.yaml
+++ b/aws/logs_monitoring/template.yaml
@@ -651,6 +651,11 @@ Resources:
                   - kms:Decrypt
                 Resource: "*"
                 Effect: Allow
+              # Get tags for log groups and attach them to the logs sent to Datadog
+              - Action:
+                  - logs:ListTagsLogGroup
+                Resource: "*"
+                Effect: Allow
               # Access the Datadog API key from Secrets Manager
               - Action:
                   - secretsmanager:GetSecretValue


aws/logs_monitoring/parsing.py
```diff
@@ -453,6 +453,26 @@ def awslogs_handler(event, context, metadata):
...
+    response = None
+    try:
+        response = cloudwatch_logs_client.list_tags_log_group(
+            logGroupName=logs["logGroup"]
+        )
+    except Exception as e:
+        logger.exception(f"Failed to get log group tags due to {e}")
+    if response is not None:
+        formatted_tags = [
+            "{key}:{value}".format(key=k, value=v) if v else k
```
sabiurr, line 466: I don't really get this ternary operator. Why would we set this to `k` if `not v`?

jvanbrie: This is to capture how AWS tags work with keys and values. If the value isn't present we don't want to send "KEY:", we want to just send "KEY"

buckhx: It's confusing indenting, but the list comprehension saying for each item, format if v exists, else just return k

`format(k, v) if v else k for k, v in items()`.

sabiurr: I think I understand whats going on but not why

So in this case, is `k` the entire tag?  If its a bad format, I'd think we'd exclude it

jvanbrie: Yeah, k is the whole tag in this case. It's just like tags in datadog without a ":" - a "just key" tag as opposed to a "key:value" tage

sabiurr (approved):

rl-michaelbrandeis:
@jvanbrie did datadog update something here and not publish a new forwarder lambda? we are getting an error

`not authorized to perform: lambda:GetLayerVersion on resource: arn:aws:lambda:us-east-1:464622532012:layer:Datadog-Forwarder:12 because no resource-based policy allows the lambda:GetLayerVersion action`

See this same error from 2021

https://github.com/DataDog/datadog-serverless-functions/issues/493

rl-michaelbrandeis:
Forgive me for intruding on your PR here. But it appears your have put the boto3 client connection inside the lambda handler so it will have to reconnect on every invoke, rather than caching a session - and calling a tag lookup on every invoke will surely hit throttling limits and cause issues in an account. The Datadog forwarder uses a LambdaTagsCache in other places to optimize handling of tags already. This seems like a regression and not an ideal solution

https://github.com/DataDog/datadog-serverless-functions/blob/1acfca2dd774c8fdcecab4ffc0f969f60b857669/aws/logs_monitoring/enhanced_lambda_metrics.py#L119

magnetikonline:
> The Datadog forwarder uses a LambdaTagsCache in other places to optimize handling of tags already. This seems like a regression and not an ideal solution

I'm just reviewing this PR too @rl-michaelbrandeis - note this PR adds the pulling of _CloudWatch Log Group tags_ - not _Lambda tags_. The abilities added are really nice and well worth it (I do question if it should be optional though).

I tend to agree though - creating a `boto3.client("logs")` client and then calling `cloudwatch_logs_client.list_tags_log_group()` on every invoke of the forwarder Lambda seems excessive?

I would have thought the CloudWatch log group tags could be cached in the same way the Lambda function tags are cached - by using a JSON object within the nominated S3 bucket for the forwarder itself.

magnetikonline:
@jvanbrie / @sabiurr something for consideration? Caching? ☝️

aws/logs_monitoring/parsing.py
Ashley-Smith: Are the tags cached? There could be a lot of calls to retrieve the log group tags - resulting in a lot of errors - any thought to handling HTTP 429 errors and providing a backoff/retry strategy?

jvanbrie:
Yep, we're making a few changes to this pr here. The layer version issue has been fixed and we're looking into some caching options for this in the meantime. This pr isn't in the current released version. For now we've reverted the change.

magnetikonline:
Thanks @jvanbrie / @sabiurr - appreciate the fast replies and turnaround 👍  Will keep an eye out for next releases. 👍

jvanbrie:
Hey everyone, I made a new pr to address some of these concerns: https://github.com/DataDog/datadog-serverless-functions/pull/540 
This pr changes three major things from this implementation.
1. We now cache these cloudwatch log group tags (so no api call per invoke)
2. You can now configure the tag collection (turn it on/off and configure refresh time)
3. No boto reconnect on each invoke

Feel free to take a look - I'll leave the pr up for a few days before looking to release.

magnetikonline:
@jvanbrie (sorry, maybe I should have commented on #540) but on this point:

> You can now configure the tag collection (turn it on/off and configure refresh time)

I couldn't see changes in that PR that allowed tag collection to be switched on/off for CloudWatch log groups?

Otherwise looked good.

dung-leviethoang-mox:
I want to report that with the current logic, the lambda assumes the log group is in the same account.
If I want centralize account contain the forwarder lambda, all log groups in several accounts use the lambda as targets for their log group subscriptions, it will throw below error:
`Failed to get log group tags due to An error occurred (ResourceNotFoundException) when calling the ListTagsLogGroup operation: The specified log group does not exist.`
Here's the issue I'm talking about https://github.com/DataDog/datadog-serverless-functions/issues/541

Just like @magnetikonline mentioned, would be better if we can have a switch to allow tag collection or not.

jvanbrie:
You can control tag collection through the environment variable DD_FETCH_LAMBDA_TAGS. This being set to false will disable tag collection in the lambda: https://github.com/DataDog/datadog-serverless-functions/pull/540/files#diff-023a52190a3595e5875c43668188d53b819c2c966ac78be21408d516391b5679R188 

I do not know how involved supporting a single lambda for multiple accounts is, but that is not a goal of this pr. If you want logs from multiple accounts with tags for log groups you can make a lambda in each account. If you don't want to make a lambda per-account you can disable tag collection for the lambda to avoid errors.

magnetikonline:
> You can control tag collection through the environment variable DD_FETCH_LAMBDA_TAGS. This being set to false will disable tag collection in the lambda: https://github.com/DataDog/datadog-serverless-functions/pull/540/files#diff-023a52190a3595e5875c43668188d53b819c2c966ac78be21408d516391b5679R188

Ah thanks @jvanbrie - I was hoping only your new CloudWatch log group tag collection could be disabled - basically retaining the existing behaviour before the PR.

I'd argue that `DD_FETCH_LAMBDA_TAGS` is now a confusing option name in that case.

PR #540, opened 2022-03-10
cached cloudwatch log group tags

jvanbrie:
### What does this PR do?
This pr separates out our caching logic and creates a new cache specifically for cloudwatch log tags.

### Motivation
The caching logic was built around the lambda custom tags use case, and that isn't great for future caching use cases. In addition we've gotten request for support for cloudwatch log group tags.

### Testing Guidelines
Lambdas are running this new code just fine. We can see the s3 caches updating as well. 

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


dylanburati (approved):
This looks good to me.

aws/logs_monitoring/cache.py
```diff
@@ -0,0 +1,421 @@
...
+
+        self.tags_by_id = {}
+        self.last_tags_fetch_time = 0
+
+    def write_cache_to_s3(self, data):
+        """Writes tags cache to s3"""
+        try:
+            logger.debug("Trying to write data to s3: {}".format(data))
+            s3_object = s3_client.Object(DD_S3_BUCKET_NAME, self.CACHE_FILENAME)
+            s3_object.put(Body=(bytes(json.dumps(data).encode("UTF-8"))))
```
sabiurr, line 127: Does the lambda log forwarder cloudformation template need to be updated for this IAM permission?

aws/logs_monitoring/cache.py
```diff
@@ -0,0 +1,421 @@
...
+            send_forwarder_internal_metrics("s3_cache_write_failure")
+            logger.debug("Unable to write new cache to S3", exc_info=True)
+
+    def acquire_s3_cache_lock(self):
+        """Acquire cache lock"""
+        cache_lock_object = s3_client.Object(
+            DD_S3_BUCKET_NAME, self.CACHE_LOCK_FILENAME
+        )
+        try:
+            file_content = cache_lock_object.get()
```
sabiurr, line 138: I'm not sure I fully understand how this s3 file is being used as a lock. My thought would be that there would be something to block on acquiring the lock, but I'm not seeing that

aws/logs_monitoring/cache.py
jvanbrie: Ah yeah, thanks for the catch. Forgot to add it to this new pr, added it now.

aws/logs_monitoring/cache.py
jvanbrie: To acquire the lock we create a lock s3 file, and then to remove the lock we delete it. If the file exists we know that the lock is active and there is a lambda running that's in the process of updating the tags. If the lock is active and we didn't grab it, then we don't need to update the tags in the current running lambda process as they'll be updated in the process that grabbed the lock.

aws/logs_monitoring/cache.py
```diff
@@ -0,0 +1,421 @@
...
+        """Makes API calls to GetResources to get the live tags of the account's Lambda functions
+
+        Returns an empty dict instead of fetching custom tags if the tag fetch env variable is not set to true
+
+        Returns:
+            tags_by_arn_cache (dict<str, str[]>): each Lambda's tags in a dict keyed by ARN
+        """
+        new_tags = {}
+        for log_group in self.tags_by_id.keys():
+            new_tags[log_group] = get_log_group_tags(log_group)
```
sabiurr: Shouldn't we return False (first var) if there was an exception here similar to in `LambdaCustomTagsCache`?

I think if there was an exception from `get_log_group_tags`, we'd overwrite as empty tags which we don't want to do

aws/logs_monitoring/cache.py
```diff
@@ -0,0 +1,421 @@
...
+        """
+        if self._is_expired():
+            send_forwarder_internal_metrics("local_cache_expired")
+            logger.debug("Local cache expired, fetching cache from S3")
+            self._refresh()
+
+        log_group_tags = self.tags_by_id.get(log_group, None)
+        if log_group_tags is None:
+            log_group_tags = get_log_group_tags(log_group)
+            self.tags_by_id[log_group] = log_group_tags
```
sabiurr, line 442: Is this needed? I think we're updating `self.tags_by_id` in the `_refresh` function

jvanbrie: Right, we'll get here if we're encountering a log group that isn't in `self.tags_by_id`, so it'll be a log group we haven't gotten logs from before. In this case we grab tags for it and add it temporarily to  `self.tags_by_id`. It'll get added to the s3 cache when we refresh. This lets us get tags for it right away rather than waiting for a refresh or refreshing all tags whenever we get 1 new log group.

aws/logs_monitoring/cache.py
jvanbrie: Here we're making multiple api calls to refill the cache, so I don't want to return false and not write to the cache if one of those api calls fails. I suppose if all the api calls fail we could return False here and skip writing to the s3 cache.
If the api call fails we can take the value we have stored from the local cache to avoid overwriting on a failed api call.

sabiurr (approved):

magnetikonline:
Just to echo my comments in https://github.com/DataDog/datadog-serverless-functions/pull/531#issuecomment-1068746625 - it seems that the existing `DD_FETCH_LAMBDA_TAGS` argument will be the only way to opt out of this feature?

jvanbrie:
Yep, all tag collection will be controlled by that argument.
Edit: Just made a change to separate out control over both tag fetches.

hghotra (approved):
LGTM 👍🏼

magnetikonline:
> Edit: Just made a change to separate out control over both tag fetches.

Wonderful - thanks @jvanbrie - reviewed that commit, makes this whole PR opt-in 👍

aws/logs_monitoring/cache.py
```diff
@@ -0,0 +1,442 @@
...
+    def _is_expired(self, last_modified=None):
+        """Returns bool for whether the fetch TTL has expired"""
+        if not last_modified:
+            last_modified = self.last_tags_fetch_time
+
+        earliest_time_to_refetch_tags = last_modified + self.tags_ttl_seconds
+        return time() > earliest_time_to_refetch_tags
+
+    def should_fetch_tags(self):
+        raise Exception("SHOULD FETCH TAGS MUST BE DEFINED FOR TABS CACHES")
```
magnetikonline:
```suggestion
        raise Exception("SHOULD FETCH TAGS MUST BE DEFINED FOR TAGS CACHES")
```

magnetikonline:
@jvanbrie did you cut the release right? I can't see any of your change between the latest release and the past?

https://github.com/DataDog/datadog-serverless-functions/compare/aws-dd-forwarder-3.42.0...aws-dd-forwarder-3.43.0

Maybe I'm just having a slow day here? 😄.

jvanbrie:
Something is definitely fishy here. Looks like the changes didn't make it into that release. I'll cut a new release tomorrow and see if that has the changes.

jvanbrie:
Thanks for bringing this up, just released a new version and it grabbed the changes this time. https://github.com/DataDog/datadog-serverless-functions/compare/aws-dd-forwarder-3.43.0...aws-dd-forwarder-3.44.0

magnetikonline:
> Thanks for bringing this up, just released a new version and it grabbed the changes this time.

No dramas @jvanbrie - agreed - can deff see the changes in there now!

PR #588, opened 2022-07-19
[AZI-62] Make partition count configurable in log forwarder automation

iotmani:
### What does this PR do?
Make partition count configurable in log forwarder automation. 
Useful for adding more control to scaling the event hub as needed.

### Motivation
AZI-62

### Testing Guidelines

I deployed the parent template, with Event Hub partitionCount set to 28, and I could see it deployed correctly:
image
image

### Additional Notes
Docs:
https://docs.microsoft.com/en-us/azure/templates/microsoft.eventhub/namespaces/eventhubs?tabs=json

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/azure/deploy-to-azure/event_hub.json
+++ b/azure/deploy-to-azure/event_hub.json
@@ -15,6 +15,13 @@
         "description": "Name of Event Hub"
       }
     },
+    "partitionCount": {
+      "type": "int",
+      "defaultValue": 32,
+      "metadata": {
+        "description": "The number of event hub partitions"
+      }
+    },
     "location": {
       "type": "string",
       "defaultValue": "[resourceGroup().location]",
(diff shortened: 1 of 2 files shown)


ashwingandhi-ddog (commented):
@iotmani I think you went over this in the demo, but why do we need this change in `parent_template.json` as well when its in `event_hub.json` ?

iotmani:
> @iotmani I think you went over this in the demo, but why do we need this change in `parent_template.json` as well when its in `event_hub.json` ?

The `event_hub.json` is what deploys the eventhub and can technically be deployed independently if so desired.
The `parent_template.json` is the entry point for all resources/templates we want customers to be able to tweak, hence to allow them to override the default we'd need to expose another parameter there also, otherwise it would default to 32.

Think of it as just passing parameter x to function `foo`, which subsequently needs to call function `bar`. (and we don't have the ability to pass all kwargs like in Python :) ).

ashwingandhi-ddog (approved):

PR #596, opened 2022-07-28
[AZI-529] improve log forwarder retry behavior

parsons90:
### What does this PR do?

Retries sending logs if the HTTP request failed. It will retry up to 4 times, with a 1 second timeout between retries, doubling each time. It will not retry http 4xx responses.

### Motivation

The network connection can be flaky and it is worth retrying a failed request.

### Testing Guidelines

Ran unit tests and tested in the Azure portal.

To test request errors, I constructed a bad http request with invalid options, and saw it attempt to retry the request. You can see the retry backoff in the timestamps.
Screen Shot 2022-07-28 at 9 13 46 AM

Retrying based on status code. It doesn't retry on a 404, but does retry with a 500.
Screen Shot 2022-07-28 at 9 51 33 AM

There was also a random timeout that got retried:
Screen Shot 2022-07-28 at 9 09 09 AM

### Additional Notes

Is 4 retries too much/not enough? How is a 1 sec timeout interval? Will doubling the timeout cause it to wait too long?

Without using external libraries, the only way to add a delay is with `setTimeout`, which requires a callback to be provided.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/azure/activity_logs_monitoring/index.js
+++ b/azure/activity_logs_monitoring/index.js
@@ -165,52 +165,64 @@ class HTTPClient {
         );
     }
 
+    isStatusCodeValid(statusCode) {
+        return statusCode >= 200 && statusCode <= 299;
+    }
+
+    shouldStatusCodeRetry(statusCode) {
+        // don't retry 4xx responses
+        return (
+            !this.isStatusCodeValid(statusCode) &&
+            statusCode < 400 &&
+            statusCode > 499
+        );
+    }
+
     send(record) {
+        var numRetries = MAX_RETRIES;
+        var retryInterval = RETRY_INTERVAL;
         return new Promise((resolve, reject) => {
-            var numRetries = MAX_RETRIES;
-            var retryInterval = RETRY_INTERVAL;
-            const httpRequest = (options, record) => {
-                const retryRequest = error => {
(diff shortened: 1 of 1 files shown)


gpalmz (approved):
LGTM. Maybe worth getting another set of eyes since we don't work on this much

azure/activity_logs_monitoring/index.js
```diff
@@ -29,6 +29,9 @@ const DD_SERVICE = process.env.DD_SERVICE || 'azure';
 const DD_SOURCE = process.env.DD_SOURCE || 'azure';
 const DD_SOURCE_CATEGORY = process.env.DD_SOURCE_CATEGORY || 'azure';
 
+const MAX_RETRIES = 4; // max number of times to retry a single http request
+const RETRY_INTERVAL = 1000; // amount of time (milliseconds) to wait before retrying request, doubles after every retry
```
claudiadadamo: The function is a bit sensitive to latency so what if we start at like 250 ms and double each time?

parsons90: Yeah, that makes sense.

Issue #583, opened 2022-07-11
Add ability to tune autoDeleteObjects value in CloudFormation Stack


shiron-babi:
**Describe what happened:**
DataDog Lambda Forwarder has S3 bucket in which some caches are being saved there.
The problem is when you try to delete the stack you have to go and manually delete the bucket content before deleting the stack otherwise the stack deletion will fail.
This issue makes it almost impossible to control the Forwarder deployment via IaC.

**Describe what you expected:**
We wish to be able to set the 
`autoDeleteObjects: true`
as part of the Stack parameters

**Steps to reproduce the issue:**
1. Deploy DD forwarder stack. 
2. Simulate some log shipping via Forwarder Lambda so the S3 Forwarder bucket will have some data in it.
3. Try to delete the CloudFormation Stack.

pimperator:
#614

PR #614, opened 2022-10-17
Adding DeletionPolicy to ForwarderBucket

pimperator:
### What does this PR do?
DD-Support-ticket 969071: adding configureable DeletionPolicy but keeping default for ForwarderBucket since stack deprovisioning fails when ForwarderBucket is not empty (s3 default on aws)  
here are the aws docs that describe how to use deletion policies on s3: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html

### Motivation
we use 'latest.yaml' to provision log forwarders via terraform or cdk.
In both cases (i.e. 'terraform destroy' or 'cdk destroy' the destruction is being reported in the state 'failed' because the Cloudformation stack is not destroyable which is caused by the non-empty s3 bucket. 

This is how it is integrated in cdk:
```
    new CfnStack(this, 'DataDog Forwarder', {
      templateUrl: 'https://datadog-cloudformation-template.s3.amazonaws.com/aws/forwarder/latest.yaml',
```
This is a message by cdk from a ci/cd pipeline:

`The stack named STACKNAME is in a failed state. You may need to delete it from the AWS console : DELETE_FAILED (The following resource(s) failed to delete: [DataDogForwarder]. ): Embedded stack arn:aws:cloudformation:eu-middleeast-1:123456789:stack/stackname-dd-forwarder-DataDogForwarder-C4R4CT3R5/s0m3-uu1d was not successfully deleted: The following resource(s) failed to delete: [ForwarderBucket].`

### Testing Guidelines
# create a stack with default values on the aws console
# delete the stack afterwards (this should fail in Cloudformation)
# create a stack with default values but choose `Delete` as deletion policy
# delete the stack afterwards which should not fail

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/template.yaml
+++ b/aws/logs_monitoring/template.yaml
@@ -219,6 +219,14 @@ Parameters:
     Type: String
     Default: ""
     Description: The name of the forwarder bucket to create. If not provided, AWS will generate a unique name.
+  FWBucketDeletionPolicy:
+    Description: DeletionPolicy options for the DdForwarderBucketName.
+    Type: String
+    Default: Retain
+    AllowedValues:
+      - Retain
+      - Delete
+      - Snapshot
 Conditions:
   IsAWSChina:
     Fn::Equals:
@@ -790,6 +798,7 @@ Resources:
   # A s3 bucket used by the Forwarder as a datastore
   ForwarderBucket:
     Type: AWS::S3::Bucket
+    DeletionPolicy: Ref: FWBucketDeletionPolicy
     Properties:
       BucketName:
         Fn::If:


pimperator:
resolves #583

pimperator:
Hold on... this does not work as I coded it since validation fails with 
`An error occurred (ValidationError) when calling the ValidateTemplate operation: Template format error: Every DeletionPolicy member must be a string.`

pimperator:
This does not work at all. I need to dig into this more since the documentation states 
> For Amazon S3 buckets, you must delete all objects in the bucket for deletion to succeed.

see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html

I just tested hardcoded the policy being set to 'Delete' and can confirm that the bucket needs to be emptied first and then deleted. So no matter what I will close this MR and re-open another one without DeletionPolicy but with a lambda that does a cleanup first.

PR #621, opened 2022-11-29
Unset default value for reserved concurrency to match AWS default one

adriacabeza:
### What does this PR do?

The default reserved concurrency for new AWS accounts has changed to 10. This makes the [one-click integration experience](https://www.datadoghq.com/blog/aws-1-click-integration/) fail for every new account that has not increase their default limits or modified the template:

`Specified ReservedConcurrentExecutions for function decreases account's UnreservedConcurrentExecution below its minimum value of [10].`

 This PR simply changes the default value of the template to empty as it is an optional parameter and it makes it work.

### Motivation

Improve the onboarding experience for AWS. 

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


jvanbrie (changes requested):
I think we'll want to set the default to 10. Setting no default will cause it to use unreserved concurrency which could rate-limit other lambdas in the account. We also probably want to add to the description that we recommend using 100 if your account is configured to allow 100 reserved concurrency for an individual lambda.

adriacabeza:
I agree that we should set a default value in order not to rate-limit other lambdas in the account. However, I just tried creating one forwarder using 10 as a default value (as it can be seen in the `requestParameters`) and it did not work (same error message). 

```
 {
    "eventVersion": "1.08",
    "userIdentity": {},
    "eventTime": "2022-11-29T14:38:17Z",
    "eventSource": "lambda.amazonaws.com",
    "eventName": "PutFunctionConcurrency20171031",
    "awsRegion": "",
    "sourceIPAddress": "AWS Internal",
    "userAgent": "AWS Internal",
    "errorCode": "InvalidParameterValueException",
    "errorMessage": "Specified ReservedConcurrentExecutions for function decreases account's UnreservedConcurrentExecution below its minimum value of [10].",
    "requestParameters": {
        "functionName": "datadog-forwarder-Forwarder",
        "reservedConcurrentExecutions": 10
    },
    "responseElements": null,
    "requestID": "",
    "eventID": "",
    "readOnly": false,
    "eventType": "AwsApiCall",
    "managementEvent": true,
    "recipientAccountId": "",
    "eventCategory": "Management",
    "sessionCredentialFromConsole": "true"
}
```

It turns out that AWS does not allow to have less than 10 unreserved concurrent executions (which is the default value for the full account concurrency). Thus, you can not reserve anything: 
image

I tried as well using 9 to confirm this and it is the case.

jvanbrie (approved):

PR #633, opened 2023-02-28
management group ARM

ashwingandhi-ddog:
ManagementGroup role assignment ARM template


iotmani (approved):
The failing test is irrelevant here, and we shouldn't even be running it for Azure.
LGTM!

PR #655, opened 2023-04-28
[azure][azi-1524] arm template separate api calls for filters

ashwingandhi-ddog:
Saving host filters and app_service_plan filters are each a separate API call to app registration call.
These 2 filter request are dependent on app registration request succeeding and they only happen if the filter values are provided.

Corresponding UIDefinition.json changes.

### DIFF
--- a/azure/deploy-to-azure/azure-integration/CreateUIDefinition-managementgroups.json
+++ b/azure/deploy-to-azure/azure-integration/CreateUIDefinition-managementgroups.json
@@ -87,7 +87,7 @@
                 "visible": true,
                 "options": {
                   "icon": "Info",
-                  "text": "The following steps involve navigating away from this template. Please read them carefully before continuing.\n\n1. Select the option to create a new app registration.\n2. Give it a name and account type (the account type doesn’t matter)\n3. Hit Register. This will take you away from this template screen.\n4. Create a Client Secret. \n5. Copy the Value of the Client Secret.\n6. Hit the X in the corner to return to this template."
+                  "text": "The following steps involve navigating away from this template. Please read them carefully before continuing.\n\n1. Select the option to create a new app registration.\n2. Give it a name and select any account type.\n3. Click 'Register'. This will open a new page to create a client secret.\n4. Click ‘+ New Client Secret’ to add a Client Secret.\n5. Copy the Value of the Client Secret.\n6. Click the close (X) button in the top-right corner to return to this template screen.\n7 Paste the value of the Client Secret in the corresponding field on this template."
                 }
               },
               {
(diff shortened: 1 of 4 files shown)


azure/deploy-to-azure/azure-integration/CreateUIDefinition-managementgroups.json
```diff
@@ -239,6 +239,17 @@
                 },
                 "visible": true
               },
+              {
+                "name": "appFilters",
```
gpalmz, line 243: Nit: could we name this `appServicePlanFilters`?

gpalmz: Nit: could we name this appServicePlanFilters throughout these files?

azure/deploy-to-azure/azure-integration/azuredeploy.json
```diff
@@ -198,7 +201,105 @@
...
+        },
+        {
+            "condition": "[not(empty(parameters('appFilters')))]",
+            "type": "Microsoft.Resources/deploymentScripts",
+            "apiVersion": "2020-10-01",
+            "name": "[concat('datadog-app-filter-script-', parameters('newguid'))]",
+            "location": "[parameters('location')]",
+            "kind": "AzurePowerShell",
+            "dependsOn": [
+                "[resourceId('Microsoft.Resources/deploymentScripts', concat('datadog-integration-script-', parameters('newguid')))]"
```
gpalmz, line 264: If I'm understanding, based on this, both of the filters upserts can be submitted concurrently. We should test to make sure this consistently works and doesn't cause some sort of concurrency-based failure in the backend. If we see issues, we should just make one of the filter upserts depend on the other

gpalmz (commented):
One thought about the filter upserts

azure/deploy-to-azure/azure-integration/azuredeploy.json
ashwingandhi-ddog: I had tested this a bunch of times and it works fine keeping the above in mind.

Some thoughts:
- Flask web server can handle multiple requests though it is not very scalable. We would have max 2 requests. It queues them if it is configured as single threaded.
-  the primary key for the `azure_namespace_filters` is `(org_id, tenant_id, client_id, namespace)` so there should not be any issues there

(Also, it's very unlikely that the 2 filter requests would be sent at the same time - given that 2 temporary containers need to be spun up to run these scripts)

gpalmz: I'm convinced, thanks!

gpalmz (approved):

PR #656, opened 2023-05-09
[azure][azi-1536] report telemetry params from arm template

ashwingandhi-ddog:
https://datadoghq.atlassian.net/browse/AZI-1536

report telemetry params from arm template and cosmetic changes

### DIFF
--- a/azure/deploy-to-azure/azure-integration/CreateUIDefinition-managementgroups.json
+++ b/azure/deploy-to-azure/azure-integration/CreateUIDefinition-managementgroups.json
@@ -191,16 +191,16 @@
                   "value": "datadoghq.com"
                 },
                 {
-                  "label": "EU1",
-                  "value": "datadoghq.eu"
-                },
-                {
                   "label": "US3",
                   "value": "us3.datadoghq.com"
                 },
                 {
                   "label": "US5",
                   "value": "us5.datadoghq.com"
+                },
+                {
+                  "label": "EU1",
+                  "value": "datadoghq.eu"
                 },
                 {
                   "label": "AP1",
--- a/azure/deploy-to-azure/azure-integration/CreateUIDefinition-managementgroups.json
+++ b/azure/deploy-to-azure/azure-integration/CreateUIDefinition-managementgroups.json
@@ -87,7 +87,7 @@
                 "visible": true,
                 "options": {
                   "icon": "Info",
-                  "text": "The following steps involve navigating away from this template. Please read them carefully before continuing.\n\n1. Select the option to create a new app registration.\n2. Give it a name and select any account type.\n3. Click 'Register'. This will open a new page to create a client secret.\n4. Click ‘+ New Client Secret’ to add a Client Secret.\n5. Copy the Value of the Client Secret.\n6. Click the close (X) button in the top-right corner to return to this template screen.\n7 Paste the value of the Client Secret in the corresponding field on this template."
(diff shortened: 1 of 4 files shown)


azure/deploy-to-azure/azure-integration/CreateUIDefinition-managementgroups.json
```diff
@@ -134,7 +134,7 @@
                 "type": "Microsoft.Common.TextBlock",
                 "visible": true,
                 "options": {
-                  "text": "To fill out the fields below, copy the API and App Keys from your Datadog org. If you launched this template from the Azure tile, you can copy the keys you selected directly from there. Otherwise you can find your API and App Keys in the Access section of the Organization Settings.",
+                  "text": "To fill out the fields below, copy the API and App Keys from your Datadog org. If you launched this template from the Azure tile, refer back to the tile to select and copy the keys. Otherwise you can find your API and App Keys in the Access section of the Organization Settings.",
```
gpalmz: Do we use the term "tile" to describe the page to customers? I wonder if we should be saying "Azure integration config page" or something, especially with the coming revamp

azure/deploy-to-azure/azure-integration/CreateUiDefinition.json
```diff
@@ -94,7 +94,7 @@
                 "type": "Microsoft.Common.TextBlock",
                 "visible": true,
                 "options": {
-                  "text": "To fill out the fields below, copy the API and App Keys from your Datadog org. If you launched this template from the Azure tile, you can copy the keys you selected directly from there. Otherwise you can find your API and App Keys in the Access section of the Organization Settings.",
+                  "text": "To fill out the fields below, copy the API and App Keys from your Datadog org. If you launched this template from the Azure tile, refer back to the tile to select and copy the keys. Otherwise you can find your API and App Keys in the Access section of the Organization Settings.",
```
gpalmz: Same question as above

azure/deploy-to-azure/azure-integration/azuredeploy-managementgroups.json
```diff
@@ -168,6 +168,9 @@
                     },
                     "newguid": {
                         "value": "[parameters('newguid')]"
+                    },
```
gpalmz, line 171: We see other entities like child management groups referring to the management group by its ID, not its name. I think we should follow that pattern.

azure/deploy-to-azure/azure-integration/azuredeploy.json
```diff
@@ -34,6 +34,10 @@
                 "description": "Subscriptions list to monitor"
             }
         },
+        "subscriptionCount": {
+            "type": "int",
+            "defaultValue": "[length(parameters('subscriptionsList'))]"
```
gpalmz, line 39: What's the meaning of `default` in this case? I would think this would always be the value, not just by default

gpalmz (commented):
A couple comments

azure/deploy-to-azure/azure-integration/azuredeploy.json
```diff
@@ -194,14 +202,18 @@
                     {
                         "name": "tenantName",
                         "value": "[subscription().tenantId]"
+                    },
+                    {
+                        "name":"source",
+                        "value": "arm_template_subscription"
```
gpalmz, line 208: Nit: should this be `arm_template_per_subscription`?

azure/deploy-to-azure/azure-integration/azuredeploy.json
ashwingandhi-ddog: Parameter fields in ARM templates renders as a text box and expects a value. As the length of the `subscriptionCount` is inferred from `subscriptionList`, we set the default value and do not expose this field.

gpalmz (approved):

PR #667, opened 2023-06-28
[AWSC] Updating Python version 3.8 -> 3.9

RaphaelAllier:
### What does this PR do?

Updates the Datadog forwarder lambda layer to python version 3.9, and updates the Datadog forwarder lambda to use 3.9 as well.
Modifies tests and README as well to ensure clients are aware of the runtime change
NOTE: Had to change the forwarder integration tests docker image due to a deprecation (see [issue](https://github.com/lambci/lambci/issues/138)). We now rely on a fork of the previous repository we used, since we need the custom options `DOCKER_LAMBDA_STAY_OPEN: 1` to be able to run integration tests. We should still rely on the last supported image (3.8) for backward compatibility for a while

### Testing Guidelines

Updates via the CloudFormation template will succeed as the new layer version will only work with 3.8 and the CF template will update the function to version 3.9. 
Tested by bundling up the new lambda layer and using it with a lambda using python 3.9

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9]
     steps:
       - name: Checkout source
         uses: actions/checkout@v3
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -5,6 +5,7 @@ on: [pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
+    continue-on-error: true
     strategy:
       max-parallel: 4
       matrix:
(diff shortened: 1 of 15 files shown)


.github/workflows/integration_test.yml
```diff
@@ -5,10 +5,11 @@ on: [pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
+    continue-on-error: true
```
sabiurr, line 8: What does this do?

RaphaelAllier: We run integration tests for 2 different Python version (`matrix` mode). When one of them fail, by default the other one stops as well. 
I disabled this behavior so they all run until the end when I was doing some testing. I think it's still a good thing to let them all run in case there's a difference in behavior between Python version that we don't catch since tests end early
Documentation: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures

sabiurr (approved):

PR #670, opened 2023-07-10
[AWSC] Bump py version of RDS/VPC Lambdas, align API key decoding

RaphaelAllier:
### What does this PR do?

Bumps the runtime version of the RDS enhanced metrics and VPC Flow Logs monitoring Lambdas from Python 3.7 to Python 3.9. Tested in our Sandbox environment. 
[What's new in Python 3.8](https://docs.python.org/3/whatsnew/3.8.html)
[What's new in Python 3.9](https://docs.python.org/3/whatsnew/3.9.html)

Also aligns the decoding of the API key of the Datadog Forwarder / VPC Flow logs Lambda with this [PR](https://github.com/DataDog/datadog-serverless-functions/pull/668). When creating the CypherTextBlob through the AWS CLI, we may encounter errors when decoding it with the current code. This should fix that.

### Motivation

Aligning python version across provided Lambdas to a supported one. 

### Testing Guidelines

Tested in our sandbox

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/settings.py
+++ b/aws/logs_monitoring/settings.py
@@ -214,15 +214,12 @@ elif "DD_KMS_API_KEY" in os.environ:
         except UnicodeDecodeError as e:
             print(
                 "INFO DD_KMS_API_KEY: Could not decode key in utf-8, encoding in b64. Exception:",
-                e
+                e,
             )
             DD_API_KEY = base64.b64encode(DD_API_KEY)
             DD_API_KEY = DD_API_KEY.decode("utf-8")
         except Exception as e:
-            print(
-                "ERROR DD_KMS_API_KEY Unknown exception decoding key:",
-                e
-            )
+            print("ERROR DD_KMS_API_KEY Unknown exception decoding key:", e)
 elif "DD_API_KEY" in os.environ:
     logger.debug("Fetching the Datadog API key from environment variable DD_API_KEY")
     DD_API_KEY = os.environ["DD_API_KEY"]
--- a/aws/logs_monitoring/settings.py
+++ b/aws/logs_monitoring/settings.py
@@ -211,10 +211,18 @@ elif "DD_KMS_API_KEY" in os.environ:
         # need to re-encode this in base64
         try:
             DD_API_KEY = DD_API_KEY.decode("utf-8")
-        except:
-            print("INFO DD_KMS_API_KEY: Could not decode key in utf-8, encoding in b64")
+        except UnicodeDecodeError as e:
+            print(
+                "INFO DD_KMS_API_KEY: Could not decode key in utf-8, encoding in b64. Exception:",
+                e
+            )
             DD_API_KEY = base64.b64encode(DD_API_KEY)
             DD_API_KEY = DD_API_KEY.decode("utf-8")
+        except Exception as e:
+            print(
+                "ERROR DD_KMS_API_KEY Unknown exception decoding key:",
+                e
+            )
 elif "DD_API_KEY" in os.environ:
     logger.debug("Fetching the Datadog API key from environment variable DD_API_KEY")
     DD_API_KEY = os.environ["DD_API_KEY"]
--- a/aws/logs_monitoring/settings.py
+++ b/aws/logs_monitoring/settings.py
@@ -207,7 +207,14 @@ elif "DD_KMS_API_KEY" in os.environ:
         CiphertextBlob=base64.b64decode(ENCRYPTED)
     )["Plaintext"]
     if type(DD_API_KEY) is bytes:
-        DD_API_KEY = DD_API_KEY.decode("utf-8")
+        # If the CiphertextBlob was encrypted with AWS CLI, we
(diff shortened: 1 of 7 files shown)


aws/logs_monitoring/settings.py
```diff
@@ -207,7 +207,14 @@ def __init__(self, name, pattern, placeholder):
         CiphertextBlob=base64.b64decode(ENCRYPTED)
     )["Plaintext"]
     if type(DD_API_KEY) is bytes:
-        DD_API_KEY = DD_API_KEY.decode("utf-8")
+        # If the CiphertextBlob was encrypted with AWS CLI, we
+        # need to re-encode this in base64
+        try:
+            DD_API_KEY = DD_API_KEY.decode("utf-8")
+        except:
+            print("INFO DD_KMS_API_KEY: Could not decode key in utf-8, encoding in b64")
```
mdelaurentis: I was surprised to see a bare `print` statement here, as opposed to a log statement. Is this because it's run as a lambda in the customer's environment and we don't have access to the logging libraries?

aws/logs_monitoring/settings.py
```diff
@@ -207,7 +207,14 @@ def __init__(self, name, pattern, placeholder):
         CiphertextBlob=base64.b64decode(ENCRYPTED)
     )["Plaintext"]
     if type(DD_API_KEY) is bytes:
-        DD_API_KEY = DD_API_KEY.decode("utf-8")
+        # If the CiphertextBlob was encrypted with AWS CLI, we
+        # need to re-encode this in base64
+        try:
+            DD_API_KEY = DD_API_KEY.decode("utf-8")
+        except:
```
mdelaurentis: Is there a specific exception type we can catch here?

aws/logs_monitoring/settings.py
```diff
@@ -207,7 +207,14 @@ def __init__(self, name, pattern, placeholder):
         CiphertextBlob=base64.b64decode(ENCRYPTED)
     )["Plaintext"]
     if type(DD_API_KEY) is bytes:
-        DD_API_KEY = DD_API_KEY.decode("utf-8")
+        # If the CiphertextBlob was encrypted with AWS CLI, we
+        # need to re-encode this in base64
```
mdelaurentis, line 211: Is this because the AWS CLI encodes it in base64 before encoding in utf-8? It double-encodes it, sort of?

mdelaurentis (commented):
Looks good, just a few questions

klivan (approved):

aws/logs_monitoring/settings.py
RaphaelAllier: You'd get something like `[ERROR] UnicodeDecodeError: 'utf-8' codec can't decode byte 0x86 in position 1: invalid start byte 1`  which isn't super meaningful

aws/logs_monitoring/settings.py
RaphaelAllier: Something like that. I couldn't find a reason looking through AWS documentation. 
Basically, if the key itself was first encrypted through the AWS CLI, then decrypting it with the SDK won't work without re-encoding it in b64. We reproduced this behaviour extensively in our Sandbox environment.
See [Plaintext](https://docs.aws.amazon.com/kms/latest/APIReference/API_Decrypt.html#API_Decrypt_ResponseSyntax) in the doc

aws/logs_monitoring/settings.py
RaphaelAllier: 2 reasons I had in mind are:
* We already do some `print` statements in similar lambdas (and i just wanted the code to be the same across lambda in this repository). We could also use a log statement
* Depending on the environment variables in the customer's Lambda, I believe some of the logs wouldn't show in the Log streams. This print statement if we ever get a support ticket on this topic should always be there (preventing some back & forth)

aws/logs_monitoring/settings.py
mdelaurentis: Ok, I just meant to avoid trying this encode/decode operation if the error was something else. What is the exception type we catch when we're unable to decode from utf-8 directly?

aws/logs_monitoring/settings.py
mdelaurentis: Got it, thanks

aws/logs_monitoring/settings.py
mdelaurentis: Got it, thanks

aws/logs_monitoring/settings.py
mdelaurentis: I see Ivan approved, feel free to merge. I think it's a good practice to catch the most specific exception type possible, so if it's possible to catch `UnicodeDecodeError` I would recommend that, but in this case I guess it's unlikely to be anything else.

RaphaelAllier: You're right, I'm updating the PR

Issue #679, opened 2023-08-01
Lambda log forwarder - changed AWS Lambda enhanced metrics publish behaviour


magnetikonline:
**Describe what happened:**

In reference to #677 

@RaphaelAllier / @sabiurr - I feel this is now a **change of behaviour** from the default - since the default setting for `DD_ENHANCED_METRICS` within the CloudFormation template is `false` - and there is no parameter/option to flip it `true`.

https://github.com/DataDog/datadog-serverless-functions/blob/55206a55999fa971aa1ac638a1f9fd0c0c20013b/aws/logs_monitoring/template.yaml#L489

Unless I'm reading this wrong - this means now that anyone deploying https://github.com/DataDog/datadog-serverless-functions/releases/tag/aws-dd-forwarder-3.76.0 will have enhanced Lambda metrics vanish - as they will no longer be publishing. Thus a change to the previous behaviour.

The fix really has to be either:

- Set `DD_ENHANCED_METRICS: "true"` in the CloudFormation.
- Add a CloudFormation argument to toggle `DD_ENHANCED_METRICS` - with the default being `True`.

As of right now - the https://github.com/DataDog/datadog-serverless-functions/releases/tag/aws-dd-forwarder-3.76.0 is somewhat broken and wouldn't recommend using it.

**Describe what you expected:**

AWS Lambda enhanced metrics continue to be published.

**Steps to reproduce the issue:**

N/A

RaphaelAllier:
Hello @magnetikonline 
Thank you for raising this and contributing !
We had an on-going freeze with our annual convention DASH, hence the delay to release a fix

magnetikonline:
No dramas @RaphaelAllier 👍 - I did feel this one was pretty serious, as it might have caused some unknown pain for users that jumped on this new version right away.

PR #687, opened 2023-08-14
Revert Enhanced metrics changes

RaphaelAllier:
We've been asked to revert the changes made to `DD_ENHANCED_METRICS` for a few reasons.

Automatically generated enhanced metrics may be confusing but they are free of charge (since metrics starts with `aws.`, then don't count towards metric billing)
People only get billed for Datadog Serverless by `aws.lambda.enhanced.invocations` metric, which isn't generated by the forwarder (it's generated by our [datadog-lambda library](https://github.com/Datadog/datadog-lambda-python) when a function gets instrumented. This behavior isn't ideal, but it has no real customer impact (all free), and it's inevitable, because the forwarder has no way to know if the logs' origin lambda function is instrumented or not (aka, whether ppl expect the forwarder to generate enhanced metrics or not). That's why we landed in that slightly confusing, yet no harm behavior.

The forwarder reuses [datadog-lambda library](https://github.com/Datadog/datadog-lambda-python) (which would generate `aws.lambda.enhanced.invocations`) under the hood, therefore DD_ENHANCED_METRICS was originally default to false to prevent the forwarder from generating `aws.lambda.enhanced.invocations` for the forwarder itself.

To avoid potential regressions we decided along w/ our Serverless team to revert to the original behavior

### DIFF
--- a/aws/logs_monitoring/README.md
+++ b/aws/logs_monitoring/README.md
@@ -117,7 +117,7 @@ If you can't install the Forwarder using the provided CloudFormation template, y
 1. Create a Python 3.9 Lambda function using `aws-dd-forwarder-<VERSION>.zip` from the latest [releases][101].
 2. Save your [Datadog API key][102] in AWS Secrets Manager, set environment variable `DD_API_KEY_SECRET_ARN` with the secret ARN on the Lambda function, and add the `secretsmanager:GetSecretValue` permission to the Lambda execution role.
 3. If you need to forward logs from S3 buckets, add the `s3:GetObject` permission to the Lambda execution role.
-4. Set the environment variable `DD_ENHANCED_METRICS` to `false` on the Forwarder. This stops the Forwarder from generating enhanced metrics from other lambdas.
+4. Set the environment variable `DD_ENHANCED_METRICS` to `false` on the Forwarder. This stops the Forwarder from generating enhanced metrics itself, but it will still forward custom metrics from other lambdas.
 5. Some AWS accounts are configured such that triggers will not automatically create resource-based policies allowing Cloudwatch log groups to invoke the forwarder. Reference the [CloudWatchLogPermissions][103] to see which permissions are required for the forwarder to be invoked by Cloudwatch Log Events.
 6. [Configure triggers][104].
 7. Create an S3 bucket, and set environment variable `DD_S3_BUCKET_NAME` to the bucket name. Also provide `s3:GetObject`, `s3:PutObject`, and `s3:DeleteObject` permissions on this bucket to the Lambda execution role. This bucket is used to store the Lambda tags cache.
--- a/aws/logs_monitoring/lambda_function.py
+++ b/aws/logs_monitoring/lambda_function.py
@@ -44,7 +44,6 @@ from settings import (
     DD_HOST,
     DD_FORWARDER_VERSION,
     DD_ADDITIONAL_TARGET_LAMBDAS,
-    DD_ENHANCED_METRICS,
 )
 
 
@@ -99,8 +98,8 @@ def datadog_forwarder(event, context):
 
     if len(trace_payloads) > 0:
         forward_traces(trace_payloads)
-    if DD_ENHANCED_METRICS:
-        parse_and_submit_enhanced_metrics(logs)
+
+    parse_and_submit_enhanced_metrics(logs)
 
 
 lambda_handler = datadog_lambda_wrapper(datadog_forwarder)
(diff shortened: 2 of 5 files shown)


saleelsaptarshi-dd (approved):

magnetikonline:
What's the reason for this revert @RaphaelAllier ?

RaphaelAllier:
@magnetikonline  Updated PR description to explain the change

magnetikonline:
Thanks @RaphaelAllier for that.

That does clear things up - but the fact still remains, until `3.76` the default was to push these metrics - as I mentioned in #679. So I guess this means it's been wrong all along?

> People only get billed for Datadog Serverless by aws.lambda.enhanced.invocations metric, which isn't generated by the forwarder (it's generated by our [datadog-lambda library](https://github.com/Datadog/datadog-lambda-python) when a function gets instrumented.

Does this apply only to Python Lambda functions?

All in all, a little confused now - I might have to deploy one of these new releases and see how our metrics look.

magnetikonline:
Infact - you've gone backwards now @RaphaelAllier 😄 

https://github.com/DataDog/datadog-serverless-functions/compare/aws-dd-forwarder-3.77.0...aws-dd-forwarder-3.83.0

So now:

- You once again provide `DD_ENHANCED_METRICS` in the CloudFormation stack.
- But once again _it's never used_! by the Python code.

So you've introduced a bug again?

RaphaelAllier:
Yes, I've gone backward to try and address the issue. 
To clarify, the env var `DD_ENHANCED_METRICS` isn't used in the Forwarder code, but it is used in the [Datadog Lambda Library code](https://github.com/DataDog/datadog-lambda-python/blob/main/datadog_lambda/metric.py#L92-L118) the Forwarder reuses.
If set to `True`, it would publish a metric `aws.lambda.enhanced.invocations` when invoked, which we use for Serverless billing. That's why it's set to `False` by default (and that led to me misunderstanding the first time I had a look at this)
The Forwarder does generate some enhanced metrics from other lambdas when forwarding their logs, but they are free of charge.

tianchu:
@magnetikonline I'm the person calling out for the revert and the original contributor to the forwarder. The missed historical context here is that `DD_ENHANCED_METRICS` was meant to control whether to generate enhanced metrics for the individual Lambda function where this env var was defined on, that's why it was set to `false` -- avoid generating the enhanced metrics for the forwarder lambda itself.

The recent changes (to be reverted) was overloading this env var's behavior to also control whether to generate enhanced metrics from logs for other lambda functions.

As @RaphaelAllier already mentioned, the original behavior was a bit confusing, but it had no real customer impact (the generated metrics are free of charge).

One alternative solution I can think of is to define a new env var say `DD_ENHANCED_METRICS_FROM_LAMBDA_LOGS` to specifically control wether to generate enhanced metrics from the forwarded lambda logs and that can default to `true` to avoid breaking changes. However, that may introduce a friction for customers who want to generate enhanced metrics for some lambda functions, but not all, they would need two forwarders -- that's actually why we didn't add a control for enhanced metrics generation in the first place.

That said, I don't see a perfect solution here, and reverting to the original behavior (no real issue for anyone since day 1) is probably the safest action. 

@RaphaelAllier if you don't mind, perhaps you could help add a comment to explain this next to https://github.com/DataDog/datadog-serverless-functions/blob/f5ecc1e05af9e891926ff48aec2838e2e0270184/aws/logs_monitoring/template.yaml#L489

magnetikonline:
Hey @RaphaelAllier / @tianchu - thanks so much for the detailed responses. That now all finally clicks as to they why and where 👍 

> that's why it was set to `false` -- avoid generating the enhanced metrics for the forwarder lambda itself.

Makes perfect sense - and 100% agree - so `DD_ENHANCED_METRICS=false` within the CloudFormation was for the benefit of the [Datadog Lambda Library code](https://github.com/DataDog/datadog-lambda-python/blob/8cd7e0dc83bad1359a6c11a1229edd1c6e3d935d/datadog_lambda/metric.py#L92-L118) the Forwarder uses internally 👍.

The function [`parse_and_submit_enhanced_metrics()`](https://github.com/DataDog/datadog-serverless-functions/blob/f5ecc1e05af9e891926ff48aec2838e2e0270184/aws/logs_monitoring/lambda_function.py#L102) relates to the parsing of logs and pushing out the enhanced metrics - for Lambdas that are not the log forwarder and emit enhanced metrics in their log output. That still remains and is active.

> One alternative solution I can think of is to define a new env var say `DD_ENHANCED_METRICS_FROM_LAMBDA_LOGS` to specifically control wether to generate enhanced metrics from the forwarded lambda logs and that can default to `true` to avoid breaking changes.

yeah being purely selfish - having the log forwarder generating enhanced metrics isn't something I would need/use. agreed.

All understood - thanks again for the clarification! 👍

PR #709, opened 2023-11-14
To support Lambda customized log group

nine5two7:
To support Lambda Customized Log group.
Customized log group can potentially be used by all AWS applications. 
This PR is to make it support Lambda.
It also compatible with default log group, and all other things supported by the DD forwarder.

[JIRA Link](https://datadoghq.atlassian.net/browse/SVLS-3889) 

Below screen shots can verify that it support both default and customized log groups for Lambda. Integration test also verify that it dose not impact other AWS applications.

Screenshot 2023-11-14 at 10 20 28 PM
Screenshot 2023-11-14 at 10 23 08 PM
Screenshot 2023-11-14 at 10 25 39 PM

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


aws/logs_monitoring/customized_log_group.py
```diff
@@ -0,0 +1,32 @@
+import re
+
+"""
+Customized log group is a log group shared by multiple applications of the same type. Based on the feedback from AWS, 
+customers may name the log group arbitrually. E.g they can name a lambda log group as "/aws/vendedlogs/states/**", which is typically used for Stepfunctions
```
agocs:
```suggestion
customers may name the log group arbitrarily. E.g they can name a lambda log group as "/aws/vendedlogs/states/**", which is typically used for Stepfunctions
```

aws/logs_monitoring/customized_log_group.py
```diff
@@ -0,0 +1,32 @@
...
+In addition, potentially, not just Lambda, any other AWS services can use a customized log group.
+The workaround is to parse the logstream_name to get the source of logs.
+"""
+
+# Example: "2023/11/06/test-customized-log-group1[$LATEST]13e304cba4b9446eb7ef082a00038990"
+REX_LAMBDA_CUSTOMIZE_LOGSTREAM_NAME_PATTERN = re.compile(
+    "^20[0-9]{2}\\/[01][0-9]\\/[0-3][0-9]\\/[0-9a-zA-Z_.-]{1,75}\\[(?:\\$LATEST|[0-9A-Za-z_-]{1,129})\\][0-9a-f]{32}$"
```
agocs: #todo: update this regex in 76 years

aws/logs_monitoring/parsing.py
```diff
@@ -495,6 +499,12 @@ def awslogs_handler(event, context, metadata):
         source = "transitgateway"
     metadata[DD_SOURCE] = parse_event_source(event, source)
 
+    # Special handling for customized log group of Lambda functions
+    # Multiple Lambda functions can share one single customized log group
+    # Need to parse logStream name to determine whether it is a Lambda function
+    if is_lambda_customized_log_group(logs["logStream"]):
+        metadata[DD_SOURCE] = "lambda"
+
```
agocs, line 507: It might not be a concern, but I worry about running a regular expression match on every Lambda log event.

agocs: On second thought, it might be the best way...

aws/logs_monitoring/parsing.py
```diff
@@ -863,3 +853,42 @@ def get_state_machine_arn(message):
...
+        # If there is no env specified, default to env:none
+        if not env_tag_exists:
+            metadata[DD_CUSTOM_TAGS] += ",env:none"
+
+
+# The lambda function name can be inferred from either a customized logstream name, or a loggroup name
+def get_lower_cased_lambda_function_name(logs):
+    logstream_name = logs["logStream"]
+    # function name parsed from logstream is preferred for handling some edge cases
+    function_name = get_lambda_function_name_from_logstream_name(logstream_name)
```
agocs, line 887: Rather than run this regex match a second time for every lambda log event, does it make sense to set the `function_name` in `awslogs_handler` based on whether we've discovered it to be a custom log group or not?

agocs (approved):
LGTM, but I have some concerns about how many times we're calling `fullmatch`. If you don't think it's a performance concern, it's probably okay

aws/logs_monitoring/customized_log_group.py
nine5two7: 😃 Let me make it more durable.

aws/logs_monitoring/parsing.py
nine5two7: The format for logstream names varies on the source of the logs. I originally expect it to be something more consistent, but it is not.

aws/logs_monitoring/parsing.py
nine5two7: I was thinking of avoiding pattern matching twice, and it turns out that in order to be compatible with the existing logics for default log group, the codes can be more verbose and hard to maintain. I tried to refactor the lambda log processing logic out into a separate function at the cost of it.

RaphaelAllier (approved):

PR #714, opened 2023-11-29
[CLOUDS-4012] Fix service tag setting in lambda forwarder

ge0Aja:
### What does this PR do?
Fixing a bug related to service tag setting in the lambda forwarder

This PR fixes the precedence of setting the service tag from multiple sources and removes the redundancy when having multiple values for the service tag from these sources.

The order of priority for setting the service tag is as follows:

1. `message` field in the log if it has a custom tags field set  i.e. `message.ddtags`
2. `ddtags` env var in the lambda logs forwarder 
3. LogGroup tags cache 
4. Lambda tags cache 
5. `ddsource` if non of the above is set
 
Previously the forwarder appended all the values set for the service from all sources in the custom tags field which created a confusion between the actual value of the service key in the log and the appended values in the custom tags field. 
This PR removes the redundant values based on the above priorities and ensures that the service key is set according to the value of the highest priority.
 
### Motivation
https://datadoghq.atlassian.net/browse/CLOUDS-4012

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


aws/logs_monitoring/parsing.py
```diff
@@ -242,15 +242,21 @@ def s3_handler(event, context, metadata):
             yield structured_line
 
 
-def get_service_from_tags(metadata):
-    # Get service from dd_custom_tags if it exists
+def get_service_from_tags_and_remove_duplicates(metadata):
```
tyrcho, line 245: you also need to rename the call from test_parsing.py

aws/logs_monitoring/lambda_function.py
```diff
@@ -257,14 +257,24 @@ def add_metadata_to_lambda_log(event):
...
+    if not event[DD_SERVICE] or event[DD_SERVICE] == event[DD_SOURCE]:
+        service_tag = next(
+            (tag for tag in custom_lambda_tags if tag.startswith("service:")),
+            f"service:{function_name}",
+        )
+        if service_tag:
+            tags.append(service_tag)
+            event[DD_SERVICE] = service_tag.split(":")[1]
+    else:
+        # remove the service tag from the cusotm lambda tags if it exists
```
tyrcho:
```suggestion
        # remove the service tag from the custom lambda tags if it exists
```

aws/logs_monitoring/lambda_function.py
```diff
@@ -319,7 +329,54 @@ def extract_ddtags_from_message(event):
                 if logger.isEnabledFor(logging.DEBUG):
                     logger.debug(f"Failed to extract ddtags from: {event}")
                 return
-        event[DD_CUSTOM_TAGS] = f"{event[DD_CUSTOM_TAGS]},{extracted_ddtags}"
+
+        event[DD_CUSTOM_TAGS] = merge_custom_and_application_tags(
+            event[DD_CUSTOM_TAGS], extracted_ddtags
+        )
+
+        ## Extract service tag from message.ddtags if exists
```
tyrcho:
```suggestion
        # Extract service tag from message.ddtags if exists
```

aws/logs_monitoring/tests/test_lambda_function.py
```diff
@@ -231,5 +313,15 @@ def test_extract_trace_payload_valid_trace(self):
         )
 
 
+class TestMergeMessageTags(unittest.TestCase):
+    def test_merge_custom_and_application_tags(self):
+        message_tags = "key0:value0,key1:value1"
```
tyrcho: can't we have repetition in tags here ? like `message_tags = "key0:value0,key1:value1,key1:value11"`

I'm wondering what is the expected / actual behavior with that

aws/logs_monitoring/lambda_function.py
```diff
@@ -319,7 +329,54 @@ def extract_ddtags_from_message(event):
...
+        if "service:" in extracted_ddtags:
+            event[DD_SERVICE] = next(
+                tag[8:]
+                for tag in extracted_ddtags.split(",")
+                if tag.startswith("service:")
+            )
+
+
+def merge_custom_and_application_tags(custom_tags, application_tags):
```
tyrcho: Could we see this function as simpler / more generic ?
- signature could be `merge_tags(specific_tags, default_tags)`
- comment would only insist on the key feature imo : specific overrides defaut

The rest of the comment could move to the place where it's used at the moment

My gut feeling is that this coud probably be reused elsewhere.

aws/logs_monitoring/lambda_cache.py
```diff
@@ -78,5 +78,12 @@ def get(self, key):
             logger.debug("Local cache expired, fetching cache from S3")
             self._refresh()
 
+        if not self.should_fetch_tags():
```
tyrcho, line 76: could we do this test before the refresh ?

aws/logs_monitoring/lambda_cache.py
```diff
@@ -78,5 +78,12 @@ def get(self, key):
...
+            logger.debug(
+                "Not fetching lambda function tags because the env variable DD_FETCH_LAMBDA_TAGS is "
+                "not set to true"
+            )
+            return []
         function_tags = self.tags_by_id.get(key, [])
```
tyrcho: we can merge with next line

aws/logs_monitoring/lambda_function.py
```diff
@@ -257,14 +257,24 @@ def add_metadata_to_lambda_log(event):
     # Get custom tags of the Lambda function
     custom_lambda_tags = get_enriched_lambda_log_tags(event)
 
-    # Set the `service` tag and metadata field. If the Lambda function is
+    # If not set during parsing or set with a default value
```
tyrcho: I have trouble understanding this comment. Maybe with extra indenting + punctuation at the end of this line ?

aws/logs_monitoring/lambda_function.py
```diff
@@ -257,14 +257,24 @@ def add_metadata_to_lambda_log(event):
     # Get custom tags of the Lambda function
     custom_lambda_tags = get_enriched_lambda_log_tags(event)
 
-    # Set the `service` tag and metadata field. If the Lambda function is
```
tyrcho, line 260: Would you add on the main comment (l232) the rule we defined with the precedence for service tag ?

aws/logs_monitoring/lambda_function.py
tyrcho: I'm also wondering if we also want another version of this which appends tags instead of replacing

 merge(k1:v1, k1:v2) => "k1:v1,k1:v2"

aws/logs_monitoring/lambda_function.py
ge0Aja: I've removed this comment on a previous revision

aws/logs_monitoring/tests/test_lambda_function.py
ge0Aja: Not sure on this, but looking back at the code. During parsing, it doesn't seem that we account for multiple tags with the same key but a different value.

aws/logs_monitoring/lambda_function.py
ge0Aja: For the append version we can just f"{custom_tags},{application_tags}" as it was before. 
I don't think we need a separate method for that

aws/logs_monitoring/lambda_function.py
ge0Aja: Changed the wording here

tyrcho (approved):

PR #715, opened 2023-11-29
approval testing

tyrcho:
# to run the tests

    pip install boto3 mock approvaltests
    python -m unittest discover ./aws/logs_monitoring/


aws/logs_monitoring/tests/test_parsing.py
```diff
@@ -924,32 +724,8 @@ def test_awslogs_handler_step_functions_tags_added_properly(
...
-            {
-                "ddsource": "stepfunction",
-                "ddtags": "env:dev,test_tag_key:test_tag_value",
-                "host": "/aws/vendedlogs/states/logs-to-traces-sequential-Logs",
-                "service": "stepfunction",
-            },
-            metadata,
-        )
+        verify_as_json(list(awslogs_handler(event, context, metadata)))
+        verify_as_json(metadata, options=NamerFactory.with_parameters("metadata"))
```
tyrcho, line 728: @ge0Aja that's a useful pattern when we really want to approve 2 different things in the same test. Ideally there would be only one verify/assert but this is acceptable.

ge0Aja (approved):
LGTM

PR #716, opened 2023-12-05
Fix a bug in processing AWS WAF logs data

ge0Aja:
### What does this PR do?

WAF logs are multiline logs forwarded to S3 using `\n` line separator. Split lines using `\n` only and don't use `splitlines()` call as it splits on other chars https://docs.python.org/3/library/stdtypes.html

### Motivation

Support request

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


RaphaelAllier (approved):

aws/logs_monitoring/parsing.py
```diff
@@ -230,7 +237,11 @@ def s3_handler(event, context, metadata):
                     "DD_MULTILINE_LOG_REGEX_PATTERN %s did not match start of file, splitting by line",
                     DD_MULTILINE_LOG_REGEX_PATTERN,
                 )
-            split_data = data.splitlines()
+            if source == "waf":
+                # WAF logs are \n separated
+                split_data = [d for d in data.split("\n") if d != ""]
```
dwikle, line 242: Why is this not simply `split_data = data.split("\n")`? Does WAF produce blank lines in the logs?

aws/logs_monitoring/tests/test_parsing.py
```diff
@@ -819,5 +821,49 @@ def test_get_lower_cased_lambda_function_name(self):
...
+        key = "mykey"
+        source = "waf"
+        verify_all_combinations(
+            lambda d: self.parse_lines(d, key, source),
+            [
+                [
+                    "123\n456\n789\n",
+                    "123\n456\n789",
+                    "Hello\rWorld!\f",
+                    "Hello\n\rWorld!\f",
```
dwikle: WAF logs are structured JSON, so the test data should probably match.

https://docs.aws.amazon.com/waf/latest/developerguide/logging-examples.html
https://docs.aws.amazon.com/waf/latest/developerguide/logging-s3.html

They don't clearly state the format, but they say it works with Athena, and it's JSON - https://docs.aws.amazon.com/athena/latest/ug/supported-serdes.html

aws/logs_monitoring/parsing.py
ge0Aja: data.split("\n") can generate a list with an empty string if data was an empty string we're just yielding non empty lines. 
I'm not sure if WAF log files could be empty or not sometimes, thus we added this safety measure.

aws/logs_monitoring/tests/test_parsing.py
```diff
@@ -819,5 +821,50 @@ def test_get_lower_cased_lambda_function_name(self):
...
+        ]
+
+    def test_get_structured_lines_waf(self):
+        key = "mykey"
+        source = "waf"
+        verify_all_combinations(
+            lambda d: self.parse_lines(d, key, source),
+            [
+                [
+                    '{"timstamp": 12345, "key1": "value1", "key2":"value2"}\n',
```
dwikle: nit: misspelled `timestamp` in this test data

dwikle (approved):

aws/logs_monitoring/tests/test_parsing.py
ge0Aja: fixed

PR #731, opened 2024-01-30
[AWS] Add support for Python 3.11, arm64 for RDS/VPC lambda

RaphaelAllier:
### What does this PR do?

Updates the Cloudformation SAM templates to support Python 3.11 and arm64 for the RDS and VPC lambdas.
Because there's a change in support architecture (from x86_64 to arm) and in runtime, it is advised to perform the update manually in 2 steps (AWS will throw an error otherwise)
* Update the runtime
* Update the architecture.
No code change otherwise

New installations from the Marketplace of applications or Cloudformation should work out of the box.

Tested all changes in our sandbox account.

### DIFF
--- a/aws/logs_monitoring/cloudwatch_log_group_cache.py
+++ b/aws/logs_monitoring/cloudwatch_log_group_cache.py
@@ -82,12 +82,14 @@ def get_log_group_tags(log_group):
     formatted_tags = None
     if response is not None:
         formatted_tags = [
-            "{key}:{value}".format(
-                key=sanitize_aws_tag_string(k, remove_colons=True),
-                value=sanitize_aws_tag_string(v, remove_leading_digits=False),
+            (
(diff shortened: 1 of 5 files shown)


ViBiOh (approved):
The configuration is not put in the same order if both files, but not a blocker

aws/logs_monitoring/cloudwatch_log_group_cache.py
```diff
@@ -82,12 +82,14 @@ def get_log_group_tags(log_group):
     formatted_tags = None
     if response is not None:
         formatted_tags = [
-            "{key}:{value}".format(
-                key=sanitize_aws_tag_string(k, remove_colons=True),
-                value=sanitize_aws_tag_string(v, remove_leading_digits=False),
+            (
```
RaphaelAllier, line 85: Drive by to fix a linting issue

PR #751, opened 2024-03-14
Add a prefix to tags cache file names

ge0Aja:
### What does this PR do?

Add a prefix string to tags cache filenames using a hash of the function ARN
This will allow to store cache files on a single S3 bucket for multiple Lambda forwarders. 
We can access the function ARN from the context object provided by AWS on function invocation 
https://docs.aws.amazon.com/lambda/latest/dg/python-context.html 
Also, update readme file by adding a troubleshooting section when encountering issues creating S3 triggers

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


aws/logs_monitoring/README.md
```diff
@@ -204,6 +204,14 @@ If you still couldn't figure out, please create a ticket for [Datadog Support][1
 ### JSON-formatted logs are not appearing in Datadog
 If your logs contain an attribute that Datadog parses as a timestamp, you need to make sure that the timestamp is both current and in the correct format. See [Log Date Remapper][24] to learn about which attributes are parsed as timestamps and how to make sure that the timestamp is valid.
 
+### Issue creating S3 triggers
+In the case of encounteirng the below error when creating S3 triggers related to using prefix filtering. We recommend considering following a fanout architecture proposed by AWS [in this article](https://aws.amazon.com/blogs/compute/fanout-s3-event-notifications-to-multiple-endpoints/)
```
tyrcho:
```suggestion
In case you encounter the following error when creating S3 triggers, we recommend considering following a fanout architecture proposed by AWS [in this article](https://aws.amazon.com/blogs/compute/fanout-s3-event-notifications-to-multiple-endpoints/)
```

tyrcho: also it could be useful to mention [here](https://github.com/DataDog/datadog-serverless-functions/pull/751/files#diff-6a31213916423d2eee54f995d4a33652ee5aa7b54c88285ad4ff8d239d9399acR123) that 

Since version xx of the forwarder, this bucket can be shared between several AWS lambda definitions of the forwarder since the cache files will be store under a different prefix in the S3 bucket for different lambdas.

Or sth like that ...

aws/logs_monitoring/caching/base_tags_cache.py
```diff
@@ -22,24 +22,36 @@
 
 
 class BaseTagsCache(object):
+    CACHE_PREFIX = None
```
tyrcho: Could we use lowercased, instance variables, initialized by the constructor for these 3 ?

The way this is defined here makes it possible to create an instance of BaseTagsCache (or subclass) which does not properly initialize the cache prefix

aws/logs_monitoring/lambda_function.py
```diff
@@ -66,8 +70,14 @@ def datadog_forwarder(event, context):
     if DD_ADDITIONAL_TARGET_LAMBDAS:
         invoke_additional_target_lambdas(event)
 
-    parsed = parse(event, context)
-    enriched = enrich(parsed)
+    # set the prefix for cache layer
+    if not cache_layer.prefix:
```
tyrcho: Could we instead create the CacheLayer with the proper prefix directly here ? and pass the prefix to Caches in `CacheLayer.__init__ `

aws/logs_monitoring/steps/enrichment.py
```diff
@@ -97,6 +97,25 @@ def add_metadata_to_lambda_log(event):
     event[DD_CUSTOM_TAGS] = ",".join([event[DD_CUSTOM_TAGS]] + tags)
 
 
+def get_enriched_lambda_log_tags(log_event, cache_layer):
+    """Retrieves extra tags from lambda, either read from the function arn, or by fetching lambda tags from the function itself.
+
+    Args:
+        log (dict<str, str | dict | int>): a log parsed from the event in the split method
```
tyrcho: is the best arg name log or log_event ?

ge0Aja: It will be a lambda log event here given the context. I think it should be good to use `log_event`

tyrcho: OK, please just rename it in the comment then.

tyrcho (approved):

PR #771, opened 2024-04-18
ECI-83 Add OCI functions to serverless repo

kanishktripathi:
### What does this PR do?

* Adds a new function for Oracle Cloud Infrastructure (OCI). This function will be used by customers to forward metrics to Datadog events intake endpoint.
* 

### Motivation

* Adding the metric forwarding function which can be used by the customers.

### Testing Guidelines

* The forwarding function is being used to send Oracle cloud metrics to staging environment

### DIFF
--- a/oci/README.md
+++ b/oci/README.md
@@ -1,0 +1,3 @@
+# Overview
+
+Repository of OCI functions that process OCI metrics and send data to Datadog as metrics events for further processing.
--- a/oci/dd-metrics-function.py
+++ b/oci/dd-metrics-function.py
@@ -1,0 +1,150 @@
+import io
+import json
+import logging
+import os
+import gzip
+
+from fdk import context, response
+import requests
+from requests.adapters import HTTPAdapter
+from requests.exceptions import HTTPError
+
+
+logger = logging.getLogger(__name__)
+
+
+OUTPUT_MESSAGE_VERSION = "v1.0"
+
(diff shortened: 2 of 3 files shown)


oci/dd-metrics-function.py
```diff
@@ -0,0 +1,150 @@
...
+    http_response = _session.post(url, data=serialized, headers=api_headers)
+    http_response.raise_for_status()
+
+    logger.info(
+        f"Sent payload size={len(metrics_message)} encoding={api_headers.get('content-encoding', None)}"
+    )
+    return http_response.text
+
+
+def handler(ctx: context.InvokeContext, data: io.BytesIO = None):
```
bengillmandd: nitpick ⛏️ Would be good to have type hint for response of this function even though you have it in the doc string

oci/README.md
```diff
@@ -0,0 +1,3 @@
+# Overview
```
bengillmandd, line 1: To be consistent with Azure and AWS we can add a link to our section in the main README https://github.com/DataDog/datadog-serverless-functions/blob/master/README.md

bengillmandd (changes requested):

oci/dd-metrics-function.py
bengillmandd: still missing return type 😬

bengillmandd (approved):

PR #618, opened 2022-11-10
Forward Step Functions Tags with logs to the backend

kimi-p:
### What does this PR do?
- Fetch Step Functions tags from AWS `resourcegroupstaggingapi` api and attach these tags to logs.
- An integration test for step function is added.
- The same S3 caching strategy is used as Lambda tags. 
- CloudFormation template is updated and `DD_FETCH_STEP_FUNCTIONS_TAGS` flag is added with default to `true`.
- `DEPLOY_TO_SERVERLESS_SANDBOX` env var flag is add for `installation_test.sh` to deploy to Serverless sandbox account.
- Two S3 buckets are created in Serverless sandbox account, `datadog-cloudformation-template-serverless-sandbox` and `dd-lambda-signing-bucket-serverless-sandbox`.
- This issue about CloudWatch logs not using S3 cache at all is not address in this PR. I will discuss with the team and have another PR to fix it. Hopefully will release this PR and the fix to CW tags caching together after that work is done.

### Motivation
`logs-to-traces` project build Step Functions traces from aws logs. To get the `env` tag, we'd like to fetch Step Function tags on the forwarder and send these tags with logs to the logs intake. `logs-to-traces` reducer will then pick up these tags and label traces with the correct `env`.
https://datadoghq.atlassian.net/browse/SLS-2718

### Testing Guidelines
Tested in Serverless sandbox account by running `./installation_test.sh` with stack deletion line commented out.

- logs are tagged with `env:staging123` and `kimi_test:kimi-test` (tags are labeled on the testing state machine) 

image

image

- JSON file used for caching also works as expected
image

- Verified that CloudWatch logs are also attached
image

### Additional Notes
[resourcegroupstaggingapi API doc](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/resourcegroupstaggingapi.html#ResourceGroupsTaggingAPI.Client.get_resources)

[Testing Forwarder Logs](https://sa-east-1.console.aws.amazon.com/cloudwatch/home?region=sa-east-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fdatadog-forwarder-integration-stack-GatR-Forwarder-5iCyL4vZrHEC/log-events/2022$252F11$252F15$252F$255B$2524LATEST$255D05c742fe8bbb4d4e875d721f039dfbc1)

```json
{
    "PaginationToken": "",
    "ResourceTagMappingList": [
        {
[...]

PR #831, opened 2024-07-30
Enable step function tracing at forwarder level

avedmala:
### What does this PR do?

This PR adds logic to allow users to enable tracing for all their Step Functions via the `dd_step_function_trace_enabled` tag

This tag can be set via the `DdStepFunctionTraceEnabled` CloudFormation parameter which is used to set the `DD_STEP_FUNCTION_TRACE_ENABLED` env var

[Link](https://ddserverless.datadoghq.com/logs?query=source%3Astepfunction%20statemachinearn%3A%22arn%3Aaws%3Astates%3Asa-east-1%3A425362996713%3Astatemachine%3Aabhinav-test-1%22%20&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&event=AgAAAZEKJDdh1-OHNwAAAAAAAAAYAAAAAEFaRUtKRjhRQUFCN3U3X2xjQ2xIZlFBQwAAACQAAAAAMDE5MTBhMjQtNWYxOS00NTliLWFkMTktNGQwN2M4YTNkNjM5&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1722446220000&to_ts=1722449820000&live=true) to log with the new tag and [screenshot](https://github.com/user-attachments/assets/16b387b7-06e4-4aaa-9258-297d211fbe90) of the tag

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/README.md
+++ b/aws/logs_monitoring/README.md
@@ -388,18 +388,6 @@ SSL encrypted TCP connection, set this parameter to true.
 `DdForwardLog`
 : Set to false to disable log forwarding, while continuing to forward other observability data, such as metrics and traces from Lambda functions.
 
-`DdStepFunctionTraceEnabled`
-: Set to false to disable log to trace conversion for Step Functions, while continuing to forward logs.
-
-`DdFetchLambdaTags`
-: Let the Forwarder fetch Lambda tags using GetResources API calls and apply them to logs, metrics, and traces. If set to true, permission `tag:GetResources` will be automatically added to the Lambda execution IAM role.
-
-`DdFetchLogGroupTags`
-: Let the forwarder fetch Log Group tags using ListTagsLogGroup and apply them to logs, metrics, and traces. If set to true, permission `logs:ListTagsLogGroup` will be automatically added to the Lambda execution IAM role.
-
-`DdFetchStepFunctionsTags`
-: Let the Forwarder fetch Step Functions tags using GetResources API calls and apply them to logs and traces (if Step Functions tracing is enabled). If set to true, permission `tag:GetResources` will be automatically added to the Lambda execution IAM role.
-
 ### Log scrubbing (optional)
 
 `RedactIp`
@@ -435,6 +423,18 @@ Some examples of regular expressions that can be used for log filtering:
 To test different patterns against your logs, turn on [debug logs](#troubleshooting).
 
 ### Advanced (optional)
+
+`DdFetchLambdaTags`
+: Let the Forwarder fetch Lambda tags using GetResources API calls and apply them to logs, metrics, and traces. If set to true, permission `tag:GetResources` will be automatically added to the Lambda execution IAM role.
+
+`DdFetchLogGroupTags`
+: Let the forwarder fetch Log Group tags using ListTagsLogGroup and apply them to logs, metrics, and traces. If set to true, permission `logs:ListTagsLogGroup` will be automatically added to the Lambda execution IAM role.
+
+`DdFetchStepFunctionsTags`
+: Let the Forwarder fetch Step Functions tags using GetResources API calls and apply them to logs and traces (if Step Functions tracing is enabled). If set to true, permission `tag:GetResources` will be automatically added to the Lambda execution IAM role.
+
+`DdStepFunctionTraceEnabled`
+: Set to false to disable log to trace conversion for Step Functions, while continuing to forward logs.
 
 `SourceZipUrl`
 : Do not change unless you know what you are doing. Override the default location of the function source code.
--- a/aws/logs_monitoring/README.md
+++ b/aws/logs_monitoring/README.md
@@ -148,7 +148,7 @@ Starting version 3.107.0 a new feature is added to enable Lambda function to sto
 
 ### Upgrade an older version to +3.106.0
 
-Starting version 3.106.0 Lambda function has been updated to add a prefix to cache filenames stored in the S3 bucket configured in `DD_S3_BUCKET_NAME`. This allows to use the same bucket to store cache files from several functions.  
+Starting version 3.106.0 Lambda function has been updated to add a prefix to cache filenames stored in the S3 bucket configured in `DD_S3_BUCKET_NAME`. This allows to use the same bucket to store cache files from several functions.
 Additionally, starting this version, the forwarder will attach custom S3 bucket tags by default to all logs exported to S3. For example, if a service is configured to send logs to a destiantion S3 bucket, the forwarder will add the bucket's tags to the logs while pulling and forwarding the logs.
 
 ### Upgrade an older version to +3.99.0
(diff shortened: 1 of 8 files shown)


aws/logs_monitoring/template.yaml
```diff
@@ -391,6 +398,10 @@ Conditions:
     Fn::Equals:
       - Ref: DdForwardLog
       - false
+  SetDdStepFunctionTraceEnabled:
+    Fn::Equals:
+      - Ref: DdStepFunctionTraceEnabled
+      - true
```
avedmala, line 404: Confused if this should be true or false

`DdForwardLog` case above sets it to false but doesn't that mean the variable will be true if the parameter is set to false?

aws/logs_monitoring/template.yaml
```diff
@@ -161,6 +161,13 @@ Parameters:
       - true
       - false
     Description: Set to false to disable log forwarding, while continuing to forward other observability data, such as metrics and traces from Lambda functions.
+  DdStepFunctionTraceEnabled:
+    Type: String
+    Default: true
```
nine5two7: I think the default value should be false, so that customers are not billed unexpectedly.

aws/logs_monitoring/template.yaml
nine5two7: These parameters are shown to customers when they are deploying the DD-Forwarder (as a Lambda).

I think the workflow is Cloud Formation Parameters (Default false for customers) => Forwarder env variables
Then, during the runtime, read this env variable and handle logs accordingly.
 
aws cloudformation config when [install new Forwarder ](https://docs.datadoghq.com/logs/guide/forwarder/?tab=cloudformation)
Screenshot 2024-07-31 at 11 27 05 AM
// aws console
Screenshot 2024-07-31 at 11 29 13 AM

Above is my understanding, you may double check with @kimi-p to verify it.

aws/logs_monitoring/steps/enrichment.py
```diff
@@ -32,6 +33,12 @@ def enrich(events, cache_layer):
         extract_host_from_guardduty(event)
         extract_host_from_route53(event)
 
+        # Set tracing behavior for all step functions
+        if DD_STEP_FUNCTION_TRACE_ENABLED:
+            event[DD_CUSTOM_TAGS] = ",".join(
```
nine5two7: Null check for `event[DD_CUSTOM_TAGS]` ?

aws/logs_monitoring/template.yaml
nine5two7, line 404: What will be the default value in CloudFormation console (shown as below) when customer install a new DD forwarder?
It should be default false, but customers can set it to be true if they want.

Screenshot 2024-07-31 at 11 27 05 AM

aws/logs_monitoring/README.md
```diff
@@ -433,6 +424,18 @@ To test different patterns against your logs, turn on [debug logs](#troubleshoot
...
+: Let the Forwarder fetch Lambda tags using GetResources API calls and apply them to logs, metrics, and traces. If set to true, permission `tag:GetResources` will be automatically added to the Lambda execution IAM role.
+
+`DdFetchLogGroupTags`
+: Let the forwarder fetch Log Group tags using ListTagsLogGroup and apply them to logs, metrics, and traces. If set to true, permission `logs:ListTagsLogGroup` will be automatically added to the Lambda execution IAM role.
+
+`DdFetchStepFunctionsTags`
+: Let the Forwarder fetch Step Functions tags using GetResources API calls and apply them to logs and traces (if Step Functions tracing is enabled). If set to true, permission `tag:GetResources` will be automatically added to the Lambda execution IAM role.
+
+`DdStepFunctionTraceEnabled`
+: Set to true to enable log to trace conversion for Step Functions.
```
nine5two7: "for all Step Functions"?

kimi-p: Maybe `Set to true to enable tracing for all Step Functions.`

aws/logs_monitoring/steps/enrichment.py
```diff
@@ -32,6 +33,12 @@ def enrich(events, cache_layer):
...
+                [event[DD_CUSTOM_TAGS]] + ["dd_step_function_trace_enabled:true"]
+            )
```
kimi-p: I think we should to add the step functions specific logic to `handle_step_function_source()`. 
https://github.com/DataDog/datadog-serverless-functions/blob/93c228160c6faffaf3482037aac5a2a10f6704fb/aws/logs_monitoring/steps/handlers/awslogs_handler.py#L161

We'd also want to do the empty value (e.g.  `""`) check like this one:
https://github.com/DataDog/datadog-serverless-functions/blob/93c228160c6faffaf3482037aac5a2a10f6704fb/aws/logs_monitoring/steps/handlers/awslogs_handler.py#L181

aws/logs_monitoring/template.yaml
kimi-p: The console will use the default value in `Parameters:` section, so in this case it would be `false`.

https://github.com/DataDog/datadog-serverless-functions/pull/831/files#diff-7c5bda646bf6435d9285913888d0473f0ac98ed0fdd025ae5d9ea357ad9b3ffaR164-R166

kimi-p (commented):
Sorry that I forgot to share the original PR with you before. Could you also add a test like this test from the previous PR?

https://github.com/DataDog/datadog-serverless-functions/pull/618/files#diff-caaaee34b52da49a4f9f45b1d6f59c91d60b24d7570e75aab8876df91f5324a7R874

aws/logs_monitoring/template.yaml
avedmala: This should be the default behavior, here is the flow

* Template has the default for the `DdStepFunctionTraceEnabled` to be `false`
* `SetDdStepFunctionTraceEnabled` condition will then be `false` because `DdStepFunctionTraceEnabled != true`
* `DD_STEP_FUNCTION_TRACE_ENABLED` will be `AWS::NoValue`, which means the env var won't be set

aws/logs_monitoring/steps/enrichment.py
avedmala: Will be moving this logic as per Kimi's comment and adding a null check on the entry

ViBiOh (approved):

aws/logs_monitoring/steps/enrichment.py
avedmala: I moved the logic to the `handle_step_function_source()` but I used a slightly different approach with the join so I don't think need the empty check

aws/logs_monitoring/steps/handlers/awslogs_handler.py
```diff
@@ -184,6 +185,12 @@ def handle_step_function_source(self):
                 + ",".join(formatted_stepfunctions_tags)
             )
 
+        if DD_STEP_FUNCTION_TRACE_ENABLED:
```
nine5two7: 👍

nine5two7:
LGTM.
Nit: We typically use the term 'Step Functions' instead of 'Step Function' to be consistent with AWS naming conventions.

nine5two7 (approved):
LGTM

PR #833, opened 2024-08-07
refactor(s3): Move all cache files to the `cache/` folder

ViBiOh:
### What does this PR do?

Change the location of caches files under a `cache/` folder prefix. So that it's allowed by the Policy and will result in a NotFoundException instead of AccessDeniedException, as explained [here](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html).

### Motivation

avoid CloudTrail pollution for AccessDeniedException

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/template.yaml
+++ b/aws/logs_monitoring/template.yaml
@@ -586,7 +586,7 @@ Resources:
                   Condition:
                     StringLike:
                       s3:prefix:
-                        - "retry/*"
+                        - "failed_events/*"
                         - "cache/*"
                   Effect: Allow
                 - !Ref AWS::NoValue
--- a/aws/logs_monitoring/template.yaml
+++ b/aws/logs_monitoring/template.yaml
@@ -587,7 +587,7 @@ Resources:
                     StringLike:
                       s3:prefix:
                         - "retry/*"
-                        - "log-group-cache/*"
+                        - "cache/*"
                   Effect: Allow
                 - !Ref AWS::NoValue
               - Action:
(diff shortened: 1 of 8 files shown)


aws/logs_monitoring/lambda_function.py
```diff
@@ -4,29 +4,30 @@
 # Copyright 2021 Datadog, Inc.
 
 import json
+import logging
```
ViBiOh: Just formatting when passing by.

aws/logs_monitoring/template.yaml
```diff
@@ -794,14 +807,6 @@ Resources:
...
                     Effect: Allow
                   - Ref: AWS::NoValue
-              # Fetch Step Functions resource tags for data enrichment
-              - Fn::If:
-                  - SetDdFetchStepFunctionsTags
-                  - Action:
-                      - tag:GetResources
-                    Resource: "*"
-                    Effect: Allow
-                  - Ref: AWS::NoValue
```
ViBiOh: This was a duplicate I fixed along the way.

tyrcho (approved):

klivan:
My concern is that `List*` operations are 10x more expensive than `Get*`, so we'll 10x customers' costs for these checks.

tyrcho:
> My concern is that `List*` operations are 10x more expensive than `Get*`, so we'll 10x customers' costs for these checks.

If this cost becomes a concern, I guess we could offer both alternatives, configurable with an env variable. 
By default we should keep the existing behaviour, and if the variable is set we would switch to List operations to reduce CloudTrail pollution with some extra S3 cost.

PR #838, opened 2024-08-16
Reformat CloudFormationTemplate

tyrcho:
No functional changes here.
- consistently use the short !form for all intrinsic functions 
- inline some definitions to make the file easier to read

The changes are easier to review [here](https://github.com/DataDog/datadog-serverless-functions/pull/838/files?diff=split&w=1#diff-7c5bda646bf6435d9285913888d0473f0ac98ed0fdd025ae5d9ea357ad9b3ffa)

### DIFF
--- a/aws/logs_monitoring/template.yaml
+++ b/aws/logs_monitoring/template.yaml
@@ -250,459 +250,287 @@ Parameters:
     Default: ""
     Description: The name of an existing s3 bucket to use. If not provided, a new bucket will be created.
 Conditions:
-  IsAWSChina:
-    Fn::Equals:
-      - Ref: AWS::Partition
-      - "aws-cn"
-  IsGovCloud:
-    Fn::Equals:
-      - Ref: AWS::Partition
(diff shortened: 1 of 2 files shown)


aws/logs_monitoring/template.yaml
```diff
@@ -718,51 +546,47 @@ Resources:
...
+                    - s3:ListBucket
+                  Resource:
+                    - !If
+                      - CreateS3Bucket
+                      - !GetAtt ForwarderBucket.Arn
+                      - !Sub "arn:aws:s3:::${DdForwarderExistingBucketName}"
+                  Condition:
+                    StringLike:
+                      s3:prefix:
+                        - "failed_events/*"
```
klivan: is change `retry` -> `failed_events` intended?

tyrcho: good catch, it will be in the next PR

klivan (approved):

PR #876, opened 2024-12-03
fix(enrichment): search for service as key, not as value in extracted ddtags

phildp:
### What does this PR do?
When your `ddtags` in log message look like this: 
```
{
  ...,
  "ddtags": "custom_key:contains-the-word-service",
  ...
}
```
meaning that there is no `service:` tag, but instead the word `service` is in a tag value, then it causes a `StopIteration` error which leads to function fail, and no logs are forwarded.
image

### Motivation

Without this PR, users are required to exclude this specific word, or find other ways to mitigate the problem. 
### Testing Guidelines

- [x] unit tests
- [x] manual test with uploaded function to AWS

### Additional Notes

The reason this causes a `StopIteration` error is that the `next()` function that is used, raises this error when an empty iterator is passed as an argument. So, if there is just the word `service` (without the colon symbol) in the `ddtags` string, returns an empty iterator.

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/steps/enrichment.py
+++ b/aws/logs_monitoring/steps/enrichment.py
@@ -154,7 +154,7 @@ def extract_ddtags_from_message(event):
                 return
 
         # Extract service tag from message.ddtags if exists
-        if "service" in extracted_ddtags:
+        if "service:" in extracted_ddtags:
             event[DD_SERVICE] = next(
                 tag[8:]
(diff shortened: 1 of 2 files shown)


ge0Aja (approved):
lgtm

PR #894, opened 2025-01-31
[AZURE] Fix event hub log forwarder duplication

ava-silver:
Finally found the source of log duplication that was happening in node 18 but got worse in node 20. Re-wrote the send function which fixed all the problems we were having.

Tested in both node 18 and 20 and saw log duplication drop to basically zero duplicated logs.


ava-silver:
* **#894** <a href="https://app.graphite.dev/github/pr/DataDog/datadog-serverless-functions/894?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/datadog-serverless-functions/894?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

azure/activity_logs_monitoring/index.js
```diff
@@ -153,35 +154,21 @@ class Batcher {
...
-        };
+        this.url = `https://${DD_HTTP_URL}:${DD_HTTP_PORT}/api/v2/logs`;
         this.scrubber = new Scrubber(this.context, SCRUBBER_RULE_CONFIGS);
-        this.batcher = new Batcher(
-            this.context,
-            256 * 1000,
-            4 * 1000 * 1000,
-            400
-        );
+        this.batcher = new Batcher(256 * 1000, 4 * 1000 * 1000, 400);
```
PCaponetti, line 159: nit: would be great to have these in top level constants so the names could inform the reader of their intent.

azure/activity_logs_monitoring/index.js
```diff
@@ -197,58 +184,52 @@ class HTTPClient {
...
+        const retryRequest = async errMsg => {
+            if (retries === 0) {
+                throw new Error(errMsg);
+            }
+            this.context.log.warn(
+                `Unable to send request, with error: ${errMsg}. Retrying ${retries} more times`
+            );
+            retries--;
+            retryInterval *= 2;
+            await sleep(retryInterval);
```
PCaponetti, line 197: Might be an easy value add to setup a more configurable backoff here. i.e. don't double retryInterval, but `await sleep(retryInterval + (MAX_RETRIES-retries)*backoffConstant + smallRandomSeed)`.

azure/activity_logs_monitoring/index.js
```diff
@@ -440,7 +418,7 @@ class EventhubLogHandler {
             if (logsType == BUFFER_ARRAY) {
                 try {
                     message = JSON.parse(message.toString());
-                } catch (err) {
+                } catch {
```
PCaponetti, line 421: seems odd to me to have catches that log stuff and don't log err.message, but I'll leave that with you as to whether or not there is value in it 🤷

PCaponetti (approved):
looks better than where it was. left a couple of suggestions.

azure/activity_logs_monitoring/index.js
ava-silver: yep, might do some updates in a future PR to add more docstrings/types as well. primarily wanted to focus on just fixing the bug here

azure/activity_logs_monitoring/index.js
ava-silver: true! again just trying to keep down the scope of this PR

azure/activity_logs_monitoring/index.js
ava-silver: yeah this code is quite old, but eventually we should be able to have customers move off it entirely with the future to come 😉

gpalmz (approved):

PR #946, opened 2025-06-13
[AWSX] feat(lambda logs forwarder): Allow to opt-out from S3 tags enrichment

ge0Aja:
### What does this PR do?

Allow to opt out from S3 tags enrichment + re-order fetched tags to have a higher priority for Forwarder set tags, notable `service`. 

If S3 tags enrichment is enabled, the fetched tags will be added infront, which should be deduplicated if the same tags are set as ENV vars on the forwarder's side.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/caching/s3_tags_cache.py
+++ b/aws/logs_monitoring/caching/s3_tags_cache.py
@@ -17,7 +17,8 @@ class S3TagsCache(BaseTagsCache):
         )
 
     def should_fetch_tags(self):
-        return os.environ.get("DD_FETCH_S3_TAGS", "false").lower() == "true"
+        # set it to true if we don't have the environment variable set to keep the default behavior
+        return os.environ.get("DD_FETCH_S3_TAGS", "true").lower() == "true"
 
     def build_tags_cache(self):
         """Makes API calls to GetResources to get the live tags of the account's S3 buckets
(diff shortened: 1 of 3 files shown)


aws/logs_monitoring/caching/s3_tags_cache.py
```diff
@@ -16,7 +17,7 @@ def __init__(self, prefix):
         )
 
     def should_fetch_tags(self):
-        return True
+        return os.environ.get("DD_FETCH_S3_TAGS", "false").lower() == "true"
```
ViBiOh: The default is `false` here but `true` in the CloudFormation. If we set it to `false` here, it's a breaking change I think.

ge0Aja: it will be false if we don't find it in the env vars, but it should be always set by the template. The behavior here is to not bring in the tags if we don't have the env var set which should be by the template.

ge0Aja: However, I'll mark it as a breaking change since it won't be automatic anymore

ge0Aja: changed to keep the current behavior

ViBiOh (approved):

PR #993, opened 2025-09-19
Feat/ AWS Log monitoring: set source using template parameter

chaimaa-kadaoui:
### What does this PR do?

- Add a new optional parameter `DdSource` for the cloudformation template to set the source of forwarded logs (setting the env var `DD_SOURCE`)

### Motivation

- We can set tags with `DdTags` , however when setting the source using this parameter : (eg. DdTags: source:test), the forwarded logs still keep the `source:s3` tag 
- By using `DdSource` to set `DD_SOURCE` this solves the issue: we have only one source which is the one set with the new parameter

Config | Output
--- | ---
`DdTags: source:kandji,service:kandji` | Screenshot 2025-09-19 at 12 13 07
`DdTags: service:kandji` <br /> `DdSource: kandji` | Screenshot 2025-09-19 at 12 17 24

### Testing Guidelines

- ✅ ran python unitests
- ✅ uploaded the template and tested the new parameter on our AWS account

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

- [x] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future

### DIFF
--- a/aws/logs_monitoring/template.yaml
+++ b/aws/logs_monitoring/template.yaml
@@ -75,6 +75,10 @@ Parameters:
     Type: String
     Default: ""
     Description: Add custom tags to forwarded logs, comma-delimited string, no trailing comma, e.g., env:prod,stack:classic
+  DdSource:
+    Type: String
+    Default: ""
+    Description: Define forwarded logs source (default to s3)
   DdFetchLambdaTags:
     Type: String
     Default: true
@@ -310,6 +314,8 @@ Conditions:
   SetS3SourceZip: !Equals [!Select [0, !Split [/, !Ref SourceZipUrl]], "s3:"]
   SetDdTags: !Not
     - !Equals [!Ref DdTags, ""]
+  SetDdSource: !Not
+    - !Equals [ !Ref DdSource, "" ]
   SetDdUseTcp: !Equals [!Ref DdUseTcp, true]
   SetDdNoSsl: !Equals [!Ref DdNoSsl, true]
   SetDdUrl: !Not
@@ -461,6 +467,10 @@ Resources:
           DD_TAGS: !If
             - SetDdTags
             - !Ref DdTags
+            - !Ref AWS::NoValue
+          DD_SOURCE: !If
+            - SetDdSource
+            - !Ref DdSource
             - !Ref AWS::NoValue
           DD_TAGS_CACHE_TTL_SECONDS: !Ref TagsCacheTTLSeconds
           DD_FETCH_LAMBDA_TAGS: !If
@@ -1017,6 +1027,7 @@ Metadata:
           default: Log Forwarding (Optional)
         Parameters:
           - DdTags
+          - DdSource
           - DdMultilineLogRegexPattern
           - DdUseTcp
           - DdNoSsl


aws/logs_monitoring/template.yaml
```diff
@@ -75,6 +75,10 @@ Parameters:
     Type: String
     Default: ""
     Description: Add custom tags to forwarded logs, comma-delimited string, no trailing comma, e.g., env:prod,stack:classic
+  DdSource:
+    Type: String
+    Default: ""
+    Description: Define forwarded logs source (default to s3)
```
ViBiOh, line 81: That's not exactly the default behavior. Putting this value change the `source` tag for **every** logs forwarded by the Forwarder, a lambda, rds or eks log will have the `source` defined here.

ViBiOh (commented):
Hi @chaimaa-kadaoui,

Thanks for your contribution, unfortunately it was not in the CloudFormation on purpose.

This environment variable is a [recent addition](https://github.com/DataDog/datadog-serverless-functions/commit/d2297a3bbd7ed54efa49bdb152be03eb442495b5) (6 months) and was responding to a customer support for a very specific use case: when our automatic identification fails to determine the correct source of an event (due to a string contains approach).

We don't want to spread usage of this variable, and so don't want to make it visible in every CloudFormation installation because it's tricky to use:
- all logs of the forwarder will have the given source, if a customer reuse a forwarder configured like that (via [autosubscription](https://docs.datadoghq.com/logs/guide/send-aws-services-logs-with-the-datadog-lambda-function/?tab=awsconsole#automatically-set-up-triggers) for example), it will lead to more support on our side
- we do tag enrichment on backend side based on the `source`. Enforcing a value will likely break that and can lead to more support too (tag are very important for database monitoring for example).

The `source` is intended to express the "AWS Product" that emitted the logs (rds, s3, eks, vpc, etc.) and the `service` to refine "which instance" (some exceptions occur to that rule, due to legacy behavior that we can't change without breaking).

In your example, `kandji` is not a `source` that we can manage or attach to any Datadog AWS Integration and you lost some tags (`region`) and information (`host`) by changing it.

chaimaa-kadaoui:
Thanks for your in-depth answer @ViBiOh ! 🙏 

> The `source` is intended to express the "AWS Product" that emitted the logs (rds, s3, eks, vpc, etc.) and the `service` to refine "which instance" (some exceptions occur to that rule, due to legacy behavior that we can't change without breaking).
> 
> In your example, `kandji` is not a `source` that we can manage or attach to any Datadog AWS Integration and you lost some tags (`region`) and information (`host`) by changing it.

- I wanted to replicate the same behavior as for cloudtrail logs:
	-  They are pushed to S3 and forwarded using datadog forwarder lambda function 
	-  Yet, S3 is not listed as a source, we only have cloudtrail

	Screenshot 2025-09-23 at 18 31 42

- Some context about our setup:
	- Kandji is our EDR 
	- They offer the possibility to forward activity logs in S3 to be picked up by other SIEM integrations ([doc](https://support.kandji.io/kb/amazon-s3-activity-log-events-integration))
	- We use a dedicated Datadog forwarder 
	- So in this case, S3 is only an intermediate storage, the source origin is kandji and the host & other attributes should be related to Kandji's events

What do you recommend for our use case?

ViBiOh:
I was not aware of Kandji tool. Yes, in this case, it can be relevant for you to force the `DD_SOURCE` env variable to the name you want (there is no [Datadog X Kandji Integration](https://docs.datadoghq.com/integrations/)).

But given the limitation of this feature (one dedicated forwarder, "broke" tag enrichment, etc.), we don't want to put it in the CloudFormation. Is it something manageable on your side to add the env variable afterward or you can only rely on CloudFormation for that?

chaimaa-kadaoui:
> But given the limitation of this feature (one dedicated forwarder, "broke" tag enrichment, etc.), we don't want to put it in the CloudFormation. Is it something manageable on your side to add the env variable afterward or you can only rely on CloudFormation for that?

Sure - we'll add it on our side in the lambda function! Thanks, I'm closing the PR

PR #996, opened 2025-09-25
refactor(aws)!: AWSX-1703 Change matching of regex. Previously done against JSON, now only log message

ViBiOh:
### What does this PR do?

Fix a behavior of running the `INCLUDE_AT_MATCH` and `EXCLUDE_AT_MATCH` against the JSON dump of the log being sent to Datadog instead of the log as received.

This is a BREAKING CHANGE if regexes where defined for JSON escaped string.

Previously, filtering a the `status` field of a JSON log was possible like this

```
\\\"status:\":\s*5\d\d
```

Now, it will be like this

```
"status":\s*5\d\d
```

### Motivation

The Regex was hard to reason about. Searching for the regex in a log file was working but not in the forwarder, this leaded to some support from customers.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [x] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/forwarder.py
+++ b/aws/logs_monitoring/forwarder.py
@@ -95,7 +95,7 @@ class Forwarder(object):
 
             evaluated_log = log
 
-            # apply scrubbing rules to inner log message if exists
+            # apply scrubbing rules to inner log message
             if isinstance(log, dict) and log.get("message"):
                 try:
                     log["message"] = scrubber.scrub(log["message"])
--- a/aws/logs_monitoring/forwarder.py
+++ b/aws/logs_monitoring/forwarder.py
@@ -4,34 +4,35 @@
 # Copyright 2021 Datadog, Inc.
 
 
-import logging
 import json
+import logging
 import os
 
-from telemetry import send_event_metric, send_log_metric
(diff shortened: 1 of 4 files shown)


aws/logs_monitoring/forwarder.py
```diff
@@ -83,25 +84,29 @@ def _forward_logs(self, logs, key=None):
...
                     logger.exception(
                         f"Exception while scrubbing log message {log['message']}: {e}"
                     )
 
-            logs_to_forward.append(json.dumps(log, ensure_ascii=False))
-
-        logs_to_forward = filter_logs(
-            logs_to_forward, INCLUDE_AT_MATCH, EXCLUDE_AT_MATCH
-        )
+            if matcher.match(evaluated_log):
```
ge0Aja, line 108: We should properly catch the exception raised by the matcher here and maybe log a warn?

ViBiOh: The previous behavior was to catch the exception and raise a new one. I think it's better to crash in this case. Otherwise Lambda will "succeed" but with bad filtering which is hidding a problem.

ge0Aja (approved):
lgtm, minor comment about exception handling

PR #1004, opened 2025-10-03
feat(aws): Move tags enablement environment variable to settings

ViBiOh:
### What does this PR do?

Refactor the load of the environment variables related to tags enrichment.

### Motivation

Reduce number of calls to `os.environ` and centralize the value into the settings.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/settings.py
+++ b/aws/logs_monitoring/settings.py
@@ -271,6 +271,37 @@ DD_MULTILINE_LOG_REGEX_PATTERN = get_env_var(
     "DD_MULTILINE_LOG_REGEX_PATTERN", default=None
 )
 
+DD_FETCH_S3_TAGS = get_env_var("DD_FETCH_S3_TAGS", default="true", boolean=True)
+
+DD_FETCH_LOG_GROUP_TAGS = get_env_var(
+    "DD_FETCH_LOG_GROUP_TAGS", default="true", boolean=True
+)
+
+DD_FETCH_LAMBDA_TAGS = get_env_var(
+    "DD_FETCH_LAMBDA_TAGS", default="false", boolean=True
+)
+
+DD_FETCH_STEP_FUNCTIONS_TAGS = get_env_var(
+    "DD_FETCH_STEP_FUNCTIONS_TAGS", default="false", boolean=True
+)
+
+
+def get_fetch_s3_tags():
+    return DD_FETCH_S3_TAGS
+
+
+def get_fetch_log_group_tags():
+    return DD_FETCH_LOG_GROUP_TAGS
+
+
+def get_fetch_lambda_tags():
+    return DD_FETCH_LAMBDA_TAGS
+
+
+def get_fetch_step_functions_tags():
+    return DD_FETCH_STEP_FUNCTIONS_TAGS
+
+
 DD_SOURCE = "ddsource"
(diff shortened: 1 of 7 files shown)


RaphaelAllier (approved):

aws/logs_monitoring/settings.py
```diff
@@ -271,6 +271,37 @@ def __init__(self, name, pattern, placeholder, enabled=True):
...
+DD_FETCH_LAMBDA_TAGS = get_env_var(
+    "DD_FETCH_LAMBDA_TAGS", default="false", boolean=True
+)
+
+DD_FETCH_STEP_FUNCTIONS_TAGS = get_env_var(
+    "DD_FETCH_STEP_FUNCTIONS_TAGS", default="false", boolean=True
+)
+
+
+def get_fetch_s3_tags():
```
ge0Aja, line 289: imo, we could just import the setting directly in the cache function an use it. not a strict opinion, but it will be consistent with the other settings consts that we import.

aws/logs_monitoring/tests/test_enhanced_lambda_metrics.py
```diff
@@ -303,7 +305,10 @@ def test_generate_enhanced_lambda_metrics_once_with_missing_arn(self):
 
     @patch("caching.base_tags_cache.send_forwarder_internal_metrics")
     @patch("caching.lambda_cache.send_forwarder_internal_metrics")
+    @patch.dict(os.environ, {"DD_FETCH_LAMBDA_TAGS": "true"})
```
ge0Aja, line 308: we can use something like
```
with patch.dict(os.environ, {"MY_ENV_VAR": "test_value"}):
```
to have local env  changes, I think it would also allow not to reload `settings` module

ge0Aja (approved):
lgtm, a couple of minor comments

aws/logs_monitoring/settings.py
ViBiOh: It was my first attempt, but with mocking involved, I needed to reload `settings`, then `lambda_cache` (for example), but `lambda_cache` was already patched with mocks and tests were deeply broken.

Using a getter allows me to only reload settings, and the rest of components will dynamically benefit from it.

aws/logs_monitoring/tests/test_enhanced_lambda_metrics.py
ViBiOh: It's used [here](https://github.com/DataDog/datadog-serverless-functions/blob/d7f492e/aws/logs_monitoring/tests/test_awslogs_handler.py#L55) for example but it doesn't prevent the use of `reload`. If I remove it, the test is broken.

ge0Aja: ok, at least the local env change part still holds if we're sharing / re-setting the same env vars amongst tests

aws/logs_monitoring/settings.py
ge0Aja: yeah I added this comment regarding consistency in settings vars imports. Otherwise, it's not really changing anything. 

Regarding the patching, I think it's a workaround here to do getters. However, I think we could use a code refactor for reworking patching. 

We can check that later

Issue #1014, opened 2025-10-25
Azure Logs - Malformed Properties Field


d1g1tal-lore-34:
### Describe what happened:
Azure logs that are generated by functions sometimes generate in proper JSON Objects under the `properties` field. This issue is documented in the follow [Microsoft Post](https://learn.microsoft.com/en-us/answers/questions/1001797/invalid-json-logs-produced-for-function-apps)

### Describe what you expected:
Azure log forwarder to check if it is a malformed JSON String, and convert it to a proper JSON Object. This will allow Datadog Log Processor to extract the proper message and log level field.

PR #1015, opened 2025-10-25
(Azure) Resolve invalid JSON Object under properties field

d1g1tal-lore-34:
### What does this PR do?

Added a function call `fixPropertiesJsonString`, that works on converting malformed JSON String that is part of the properties field into a proper JSON object. This issue is documented in the follow [Microsoft Post](https://learn.microsoft.com/en-us/answers/questions/1001797/invalid-json-logs-produced-for-function-apps).

### Motivation

This will allow Datadog Log Processor to extract the proper message and log level field. This fixes issue #1014 

### Testing Guidelines

It has been tested in two manners:
- Tested in my Datadog Instance
- Test cases have been created under `EventhubLogHandler Fix Properties Json String`

### Additional Notes

This issue is documented in the follow [Microsoft Post](https://learn.microsoft.com/en-us/answers/questions/1001797/invalid-json-logs-produced-for-function-apps).

### Types of changes

- [x] Bug fix

### Check all that apply

### DIFF
--- a/azure/activity_logs_monitoring/index.js
+++ b/azure/activity_logs_monitoring/index.js
@@ -396,12 +396,15 @@ class EventhubLogHandler {
                 // If the check is true, find and replace single quotes
                 // with double quotes, to make a proper JSON
                 // which is then converted into a JSON Object
-                parsedProperties = JSON.parse(record.properties.replace(/'/g, '"'));
-                record.properties = parsedProperties;
+                record.properties = JSON.parse(record.properties.replace(/'/g, '"'));
+                return record;
             } catch {
                 this.context.error('Unable to fix properties field to JSON Object');
+                return record;
             }
-        return record;
+        } else {
+            return record;
+        }
     }
 
     handleLogs(logs) {
--- a/azure/activity_logs_monitoring/index.js
+++ b/azure/activity_logs_monitoring/index.js
@@ -381,11 +381,35 @@ class EventhubLogHandler {
                     this.records.push(originalRecord);
                 }
             } else {
-                this.records.push(originalRecord);
+                this.records.push(this.fixPropertiesJsonString(originalRecord));
             }
         } else {
             record = this.addTagsToStringLog(record);
             this.records.push(record);
+        }
+    }
+
+    fixPropertiesJsonString(record) {
+        try {
+            // Check if properties field is a string
+            if (typeof record.properties === 'string') {
+                // If it is a string, check if it is a malformed JSON String
+                // By checking for ':'
+                if (record.properties.includes("':'")) {
+                    // If the check is true, find / replace single quotes
+                    // with double quotes, to make a proper JSON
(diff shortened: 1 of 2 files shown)


azure/activity_logs_monitoring/index.js
```diff
@@ -381,14 +381,38 @@ class EventhubLogHandler {
...
+                    return record;
+                }
+            } else {
+                return record;
+            }
+        } catch {
+            this.context.error('Unable to fix properties field to JSON Object');
+            return record;
+        }
+    }
```
mattsp1290, line 408:
```suggestion
    fixPropertiesJsonString(record) {
        // Check if properties field is a malformed JSON String
        if (Object.hasOwn(record, 'properties') && (typeof record.properties === 'string') && (record.properties.includes("':'"))) {
            try {
                // If the check is true, find and replace single quotes
                // with double quotes, to make a proper JSON
                // which is then converted into a JSON Object
                parsedProperties = JSON.parse(record.properties.replace(/'/g, '"'));
                record.properties = parsedProperties;
            } catch {
                this.context.error('Unable to fix properties field to JSON Object');
            }
        return record;
    }
```
This function has a lot of return paths that all seem to do the same thing. This could be simplified a bit to have a single return and combine our checks into a single statement to verify whether or not record.properties has one of these malformed objects.

We'll need to do a little verification on our end to make this work with live data and double check the behavior of nested message and level fields.

mattsp1290 (changes requested):
Overall this seems like a useful change. Thanks for putting this together. We'd just like to see a little simplification in the added function to reduce the surface of any issues.

azure/activity_logs_monitoring/index.js
d1g1tal-lore-34: @mattsp1290 I went ahead and took the changes that you've recommended. I added a couple of return clauses that ensures:
- When the if statement fails, we should return the record as is
- If the conversion fails, we should return the record as is

```js
    fixPropertiesJsonString(record) {
        // Check if properties field is a malformed JSON String
        if (Object.hasOwn(record, 'properties') && (typeof record.properties === 'string') && (record.properties.includes("':'"))) {
            try {
                // If the check is true, find and replace single quotes
                // with double quotes, to make a proper JSON
                // which is then converted into a JSON Object
                record.properties = JSON.parse(record.properties.replace(/'/g, '"'));
                return record;
            } catch {
                this.context.error('Unable to fix properties field to JSON Object');
                return record;
            }
        } else {
            return record;
        }
    }
```

I have also added a couple of additional test cases, that will test nested objects.

mattsp1290 (approved):
I have been able to validate and will make the remaining required changes in a follow up PR.

mattsp1290:
You can track [this PR](https://github.com/DataDog/datadog-serverless-functions/pull/1024) for when this will be available via the deploy to azure button in this repo

PR #1020, opened 2025-11-05
AWS KMS.1 compliant policy with dedicated kms key list

n0pants:
### What does this PR do?

Adding an input parameter **KmsKeyList** to the CFN template.yaml to define a comma seperated list of KMS Key ARNs.
That list will be used to tailor kms:decrypt resource access in the IAM policy of the datadog serverless forwarder.

### Motivation

AWS security Hub finding KMS.1 as well as CKV_AWS_356 in the default template.
Having wildcard permissions on all AWS KMS keys is a security issue. 
Allowing access to only dedicated KMS keys keeps the setup secure.
Keys using default encryption aws/secretsmanager are accessible by default.

### Additional Notes
While the original CFN template contains a wildcard permission for "customer managed" keys, integrating this version will break access to those CMKs, if they are not listed in the parameter.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/template.yaml
+++ b/aws/logs_monitoring/template.yaml
@@ -289,6 +289,9 @@ Parameters:
       - "true"
       - "false"
     Description: Set to true to enable enhanced Lambda metrics. This will generate additional custom metrics for Lambda functions, including cold starts, estimated AWS costs, and custom tags. Default is false.
+  KmsKeyList:
+    Type: CommaDelimitedList
+    Description: List of KMS Key ARNs the Lambda forwarder function can use to decrypt, seperated by comma  
 Conditions:
   IsAWSChina: !Equals [!Ref "AWS::Partition", aws-cn]
   IsGovCloud: !Equals [!Ref "AWS::Partition", aws-us-gov]
@@ -378,6 +381,8 @@ Conditions:
     - !Equals [!Join ["", !Ref VPCSubnetIds], ""]
   SetDdLogLevel: !Not
     - !Equals [!Ref DdLogLevel, ""]
+  SetDdForwarderDecrypt: !Not
+    - !Equals [!Join ["", !Ref KmsKeyList], ""]
 Rules:
   MustSetDdApiKey:
     Assertions:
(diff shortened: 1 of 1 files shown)


ge0Aja:
👋 @n0pants 
There's already a `PermissionsBoundaryArn` template variable to set the permission boundary for the forwarder. 
The forwarder still needs `kms:Decrypt` action for SSE-KMS keys.

n0pants:
Hi @ge0Aja 👋
thanks for having a look and your feedback.
Permission boundaries are useful to handle generic actions or resources. Obviously secrets are pretty unique per AWS account. Running hundreds of AWS accounts makes it kind of impossible to set a generic boundary.

Rather than opting out of wildcard secret resource permissions, it should be an explicit opt in to keep it secure.

I do completely agree that you need `kms:Decrypt` for SSE-KMS access, like for all encryption using CMKs. But that´s exactly what the change does. Providing an allow list to access to CMKs and prevent access to all other keys.

ge0Aja:
> Hi @ge0Aja 👋 thanks for having a look and your feedback. Permission boundaries are useful to handle generic actions or resources. Obviously secrets are pretty unique per AWS account. Running hundreds of AWS accounts makes it kind of impossible to set a generic boundary.
> 
> Rather than opting out of wildcard secret resource permissions, it should be an explicit opt in to keep it secure.
> 
> I do completely agree that you need `kms:Decrypt` for SSE-KMS access, like for all encryption using CMKs. But that´s exactly what the change does. Providing an allow list to access to CMKs and prevent access to all other keys.

Thanks for the explanation, I'd suggest to pass the list of keys as an optional parameter then. I believe we should still keep the option to use a wildcard if no keys were specified.

n0pants:
Thanks & agree. As it would also remove the breaking character of the change.
I´ve turned the default of the input to empty which means kms:Decrypt on all resources (equals current state)
or explicit access in case the input parameter is set.

ge0Aja (approved):

Issue #1043, opened 2026-01-14
VPC Flow Log header line forwarded to Datadog as log entry


rohanpower:
When processing VPC Flow Logs from S3, the Datadog Forwarder forwards the header line as a regular log entry to Datadog instead of skipping it.

**Expected Behavior:**
VPC Flow Log header lines (e.g. version account-id interface-id srcaddr dstaddr srcport dstport protocol packets bytes start end action log-status) should be skipped and not forwarded to Datadog.

**Actual Behavior:**
The header line appears in Datadog as a log entry with the message field containing the header text.

**Example Log in Datadog:**
```
{
  "message": "version account-id interface-id srcaddr dstaddr srcport dstport protocol packets bytes start end action log-status",
  "ddsource": "vpc",
  "service": "vpc"
}
```

**Root Cause:**
In 'aws/logs_monitoring/steps/handlers/s3_handler.py', the _extract_other_logs() method processes all lines from VPC Flow Log files without
detecting or skipping the header line.

**Proposed Solution:**
Detect VPC Flow Logs by checking if "vpcflowlogs" is in the S3 key path, then skip the first line (header) for these files.

**Code Location:**
- **File:** aws/logs_monitoring/steps/handlers/s3_handler.py
- **Function:** _extract_other_logs()

**Suggested Fix:**
```
is_vpc_flowlog = "vpcflowlogs" in self.data_store.key.lower()

for idx, line in enumerate(self.data_store.data.splitlines()):
    line = line.decode("utf-8", errors="ignore").strip()
    if len(line) == 0:
        continue

    # Skip first line of VPC Flow Logs (header)
    if is_vpc_flowlog and idx == 0:
        continue

    yield self._format_event(line)
```

**Environment:**
- Forwarder Version: 5.1.1
- S3 Key Pattern: AWSLogs/{account-id}/vpcflowlogs/{region}/{date}/...

ge0Aja:
Thanks @rohanpower This fix should be in the next release

PR #1044, opened 2026-01-14
[AWSX] fix(logs fowarder): Skip header line on VPC flow logs parsing

ge0Aja:
### What does this PR do?

* Skip header line on AWS VPC flow logs parsing

### Motivation

See https://github.com/DataDog/datadog-serverless-functions/issues/1043

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring/steps/common.py
+++ b/aws/logs_monitoring/steps/common.py
@@ -55,6 +55,9 @@ def is_cloudtrail(key):
     match = CLOUDTRAIL_REGEX.search(key)
     return bool(match)
 
+def is_vpc_flowlog(key):
+    return "vpcflowlogs" in key
+
 
 def find_cloudwatch_source(log_group):
     for prefix in AwsCwEventSourcePrefix:
--- a/aws/logs_monitoring/steps/common.py
+++ b/aws/logs_monitoring/steps/common.py
@@ -55,6 +55,7 @@ def is_cloudtrail(key):
     match = CLOUDTRAIL_REGEX.search(key)
     return bool(match)
 
+
 def is_vpc_flowlog(key):
     return "vpcflowlogs" in key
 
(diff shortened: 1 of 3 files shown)


aws/logs_monitoring/steps/common.py
```diff
@@ -56,6 +56,10 @@ def is_cloudtrail(key):
     return bool(match)
 
 
+def is_vpc_flowlog(key):
+    return "vpcflowlogs" in key
```
ViBiOh, line 60: **💭 thought:** ‏ I'm not fan of this as it puts back some source identification in the forwarder and also some business logic.

While this check is the current method for S3-based VPC Flow Log, we also support CloudWatch logs for VPC Flow Logs (they seems to not face the issue). Is there we can implement this on backend side so that if we update the source identification we don't have two places to maintain it?

ge0Aja: I think this is doable from the backend, but requires a change in the logic as we won't know which line comes first on the intake side. For that we need to perform a keyword check maybe ? For the file key, it's already included in the log payload so we could use that instead of doing the check here. 

The only downside is that we're going to still send the log and filter it out in the backend. 

Regarding the source identification I agree that this brings back some business logic, but it doesn't add any extra fields to the log itself. We're still going to keep some business logic eventually (same goes for cloudtrail), but I think it should be fine if we don't use it amending payload to the log itself. 

I'm fine with either choices by the way, this one seemed more logical as we trim sending the log at the source and not filter it out on the back-end side.

ViBiOh: You got me at the "we don't know the order on backend side" so Ok for this.

ViBiOh (approved):

PR #1063, opened 2026-02-10
feat(forwarder): add SQS support as DLQ backend for failed events

LorisFriedel:
## Summary

- Add AWS SQS as an alternative Dead Letter Queue backend for failed event storage, alongside the existing S3 backend
- Users set `DD_SQS_QUEUE_URL` to point to a pre-existing SQS queue; when unset, existing S3 behavior is unchanged
- Pluggable storage pattern: `BaseStorage` ABC with `S3Storage` (renamed from `Storage`) and new `SQSStorage`
- Factory function `create_storage()` selects backend based on configuration
- CloudFormation template updated with `DdSqsQueueUrl` parameter, scoped IAM permissions, and `HasStorageBackend` condition

### SQS design decisions
- Single queue with `MessageAttributes` for prefix separation (`retry_prefix`, `function_prefix`)
- Data chunked to fit under SQS 256KB limit (240KB with overhead margin)
- `get_data()` polls up to 10 iterations, releases non-matching messages immediately
- `ReceiptHandle` used as key for deletion (idempotent)
- IAM permissions scoped to exact queue ARN (derived from URL via `!Split`/`!Select`)

### Operational note
When using SQS, the queue's **VisibilityTimeout** should be >= the Lambda function's **Timeout** (default 120s) to prevent duplicate processing during retries.

---

[OBSPLTF-947](https://datadoghq.atlassian.net/browse/OBSPLTF-947)

[OBSPLTF-947]: https://datadoghq.atlassian.net/browse/OBSPLTF-947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/aws/logs_monitoring/retry/storage.py
+++ b/aws/logs_monitoring/retry/storage.py
@@ -6,13 +6,14 @@ from time import time
 import boto3
 from botocore.exceptions import ClientError
 
+from retry.base_storage import BaseStorage
 from settings import DD_S3_BUCKET_NAME, DD_S3_RETRY_DIRNAME
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.getLevelName(os.environ.get("DD_LOG_LEVEL", "INFO").upper()))
 
 
-class Storage(object):
+class S3Storage(BaseStorage):
     def __init__(self, function_prefix):
         self.bucket_name = DD_S3_BUCKET_NAME
         self.s3_client = boto3.client("s3")
@@ -85,3 +86,7 @@ class Storage(object):
 
     def _deserialize(self, data):
         return json.loads(data.decode("UTF-8"))
+
+
+# Backward-compatible alias
+Storage = S3Storage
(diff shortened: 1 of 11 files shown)


ViBiOh (approved):

aws/logs_monitoring/retry/storage.py
```diff
@@ -81,7 +82,11 @@ def _get_key_prefix(self, retry_prefix):
...
 
     def _serialize(self, data):
-        return bytes(json.dumps(data).encode("UTF-8"))
+        return json.dumps(data).encode("UTF-8")
 
     def _deserialize(self, data):
         return json.loads(data.decode("UTF-8"))
+
+
+# Backward-compatible alias
```
ge0Aja: Could you explain why we need this ? I think it should be sufficient that we have introduced a factory method with a base class

LorisFriedel: You're right, that's an old artefact not needed anymore, I've cleaned it up Thanks!

PR #1064, opened 2026-02-11
feat(forwarder): add SQS support as event source for S3 notifications

LorisFriedel:
## Summary

- Add SQS event source support so S3 notifications delivered via SQS queues (`S3 -> SQS -> Lambda` and `S3 -> SNS -> SQS -> Lambda`) are correctly parsed and forwarded to Datadog
- SQS records are detected by `eventSource: "aws:sqs"`, unwrapped (direct S3 or SNS-wrapped S3), and delegated to the existing `S3EventHandler` with fresh per-record metadata
- Add required SQS IAM permissions (`sqs:ReceiveMessage`, `sqs:DeleteMessage`, `sqs:GetQueueAttributes`) to the CloudFormation template

## Changes

| File | Change |
|------|--------|
| `steps/enums.py` | Add `SQS` to `AwsEventType` enum |
| `steps/parsing.py` | Add SQS detection in `parse_event_type()`, `sqs_handler`, `_extract_inner_event_from_sqs`, `_contains_s3_records` |
| `template.yaml` | Add SQS IAM permissions to `ForwarderRolePolicy0` |
| `tests/test_parsing.py` | Add 9 unit tests (event type detection + parsing) |
| `tests/events/sqs_s3.json` | New fixture: S3 -> SQS event |
| `tests/events/sqs_sns_s3.json` | New fixture: S3 -> SNS -> SQS event |

---

[OBSPLTF-945](https://datadoghq.atlassian.net/browse/OBSPLTF-945)

### DIFF
--- a/aws/logs_monitoring/template.yaml
+++ b/aws/logs_monitoring/template.yaml
@@ -754,6 +754,13 @@ Resources:
                   Resource: "*"
                   Effect: Allow
                 - !Ref AWS::NoValue
+              # Required for Lambda event source mappings that poll SQS queues
+              - Action:
+                  - sqs:ReceiveMessage
+                  - sqs:DeleteMessage
+                  - sqs:GetQueueAttributes
+                Resource: "*"
+                Effect: Allow
       Tags:
         - Value: !FindInMap [Constants, DdForwarder, Version]
           Key: dd_forwarder_version
--- a/aws/logs_monitoring/template.yaml
+++ b/aws/logs_monitoring/template.yaml
@@ -311,6 +311,10 @@ Parameters:
     Type: CommaDelimitedList
     Default: ""
     Description: List of S3 bucket ARNs the forwarder can read from (e.g. arn:aws:s3:::my-bucket/*), separated by comma. Leave empty to allow all buckets. Note that restricting this list may break the automatic creation of S3 triggers for buckets not included.
+  SqsQueueArnList:
+    Type: CommaDelimitedList
+    Default: ""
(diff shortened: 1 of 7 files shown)


aws/logs_monitoring/template.yaml
```diff
@@ -754,6 +754,13 @@ Resources:
                   Resource: "*"
                   Effect: Allow
                 - !Ref AWS::NoValue
+              # Required for Lambda event source mappings that poll SQS queues
+              - Action:
+                  - sqs:ReceiveMessage
+                  - sqs:DeleteMessage
+                  - sqs:GetQueueAttributes
+                Resource: "*"
```
ge0Aja: Something that we've done recently was to allow targeting specific S3 buckets to read logs from. 
I think we can do the same for SQS permissions here. Could you add an optional parameter to target specific SQS queues rather than using `*`

ge0Aja (approved):
lgtm, needs a lint check (we're using py3.13) for the moment since cf-lint is not supported in py3.14

aws/logs_monitoring/template.yaml
LorisFriedel: Done! Added optional `SqsQueueArnList` parameter following the `S3BucketArnList` / `SetS3BucketArns` pattern:

- **New parameter** `SqsQueueArnList` (CommaDelimitedList, default empty) — list of SQS queue ARNs the forwarder can receive from
- **New condition** `SetSqsQueueArns` — true when list is non-empty
- **Updated SQS IAM Resource** — uses `!If` to scope to provided ARNs when set, falls back to `"*"` otherwise

Template validates cleanly with `cfn-lint`.

PR #1080, opened 2026-03-16
perf(forwarder): parse event message JSON once in split()

LorisFriedel:
## Summary
- `extract_metric()` and `extract_trace_payload()` both called `json.loads(event["message"])` independently
- For plain log events (the majority), both fail with exceptions — 2x parse + 2x exception overhead per event
- Parse once in `split()` and pass the pre-parsed dict directly to the extractors
- `extract_metric(parsed, event)` validates and enriches a pre-parsed dict
- `is_trace(parsed)` is a pure predicate — the trace payload dict is built in `split()` since it only needs the raw message and tags
- Events that fail JSON parsing are classified as logs immediately, skipping both extractors

## Test plan
- [ ] `test_splitting.py` tests pass (12 tests covering metric validation, trace detection, and edge cases)
- [ ] Verify log/metric/trace splitting works end-to-end via integration tests

### DIFF
--- a/aws/logs_monitoring/steps/splitting.py
+++ b/aws/logs_monitoring/steps/splitting.py
@@ -7,29 +7,23 @@ logger = logging.getLogger()
 logger.setLevel(logging.getLevelName(os.environ.get("DD_LOG_LEVEL", "INFO").upper()))
 
 
-_PARSE_FAILED = object()
-
-
-def _try_parse_message(event):
-    """Parse event message JSON once. Returns parsed object or _PARSE_FAILED sentinel."""
-    try:
-        return json.loads(event["message"])
-    except Exception:
-        return _PARSE_FAILED
-
-
 def split(events):
     """Split events into metrics, logs, and trace payloads"""
     metrics, logs, trace_payloads = [], [], []
     for event in events:
-        parsed = _try_parse_message(event)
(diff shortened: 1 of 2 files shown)


aws/logs_monitoring/steps/splitting.py
```diff
@@ -7,15 +7,28 @@
...
+
+
 def split(events):
     """Split events into metrics, logs, and trace payloads"""
     metrics, logs, trace_payloads = [], [], []
     for event in events:
-        metric = extract_metric(event)
-        trace_payload = extract_trace_payload(event)
+        parsed = _try_parse_message(event)
+        metric = extract_metric(event, parsed_message=parsed)
```
ViBiOh: **💬 suggestion:** ‏ I think we can avoid adding a parameter it and wrap everything into a try/catch, passing the parsed message as long as the raw event.

The first instruction of each function is indeed to parse the message.

For traces, we could change it to just `is_trace` because it returns something that doesn't depend on anything parsed.

For metrics, we use the parsed message and event, so we can keep both

LorisFriedel: Great idea, I've made the changes!

ViBiOh (approved):

PR #1081, opened 2026-03-16
perf(forwarder): create scrubber, matcher, and batcher once in __init__

LorisFriedel:
## Summary
- `_forward_logs()` created a new `DatadogScrubber` (compiles regex per scrubbing rule), `DatadogMatcher` (compiles include/exclude regex), and `DatadogBatcher` on every call
- These are config-driven and stateless after construction
- Move all three to `Forwarder.__init__()` as `self._scrubber`, `self._matcher`, and `self._batcher`
- Since `Forwarder` is a container-global singleton, this also eliminates re-compilation across warm Lambda invocations
- Also simplifies the log evaluation loop in `_forward_logs` and uses idiomatic emptiness checks

## Test plan
- [ ] Existing `test_forwarder.py` tests pass
- [ ] Verify log forwarding with include/exclude rules works end-to-end via integration tests

### DIFF
--- a/aws/logs_monitoring/forwarder.py
+++ b/aws/logs_monitoring/forwarder.py
@@ -36,12 +36,21 @@ logger = logging.getLogger()
 logger.setLevel(logging.getLevelName(os.environ.get("DD_LOG_LEVEL", "INFO").upper()))
 
 
-class Forwarder(object):
+class Forwarder:
     def __init__(self, function_prefix):
         self.trace_connection = TraceConnection(
             DD_TRACE_INTAKE_URL, DD_API_KEY, DD_SKIP_SSL_VALIDATION
         )
         self.storage = create_storage(function_prefix)
+        self._scrubber = DatadogScrubber(SCRUBBING_RULE_CONFIGS)
+        self._matcher = DatadogMatcher(
+            include_pattern=INCLUDE_AT_MATCH, exclude_pattern=EXCLUDE_AT_MATCH
+        )
+        self._batcher = DatadogBatcher(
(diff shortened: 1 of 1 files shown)


aws/logs_monitoring/forwarder.py
```diff
@@ -81,40 +90,33 @@ def _forward_logs(self, logs, key=None):
...
-                if log.get("message"):
-                    evaluated_log = log["message"]
-                else:
-                    to_forward = dump_event(log)
-                    evaluated_log = to_forward
+                serialized = dump_event(log)
+                evaluated_log = log.get("message") or serialized
+            else:
+                serialized = dump_event(log)
+                evaluated_log = log
```
ViBiOh: **💬 suggestion:** ‏ We'll dump event to JSON even if we don't forward them with this change. Some event, especially cloudtrail, can be very large and that's why we wanted to dump only if required.

LorisFriedel: 👍 I've updated the code to defer `json.dumps` until after matching. For dicts with a message field (CloudTrail), it now matches on the message string directly and only serialize if the event passes the filter. This avoids wasted serialization for excluded events!

ViBiOh: Great catch on the serialization concern! That said, this is actually what the code did before your change — you've essentially circled back to the previous behavior, just with the matching logic split across three branches now. The original version intentionally serialized once and matched once in a single path, which I'd argue was easier to follow. 🤖

ViBiOh (approved):

PR #1084, opened 2026-03-20
[AWSX-2144] feat(go-forwarder): add basic api key and config logic

ndakkoune:
### What does this PR do?
Add the logging initialization, the configuration, the api key retrieval/verification strategies and their tests.

### Motivation
Following the migration of the Log Forwarder to Go, the logging, the api key retrieval/verification and the loading/declaration of the configuration are required to test the most basic version of the new forwarder.

### Additional Notes
- The proxy configuration should be configured through `HTTP_PROXY` and `HTTPS_PROXY` environment variables. We will probably deprecate the SSL skipping related flags for now.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


aws/logs_monitoring_go/internal/config/config.go
```diff
@@ -0,0 +1,111 @@
...
+
+	cfg.deriveURLs()
+	slog.Debug("config loaded", "site", cfg.Site, "intakeURL", cfg.IntakeURL, "apiURL", cfg.APIURL, "port", cfg.Port, "noSSL", cfg.NoSSL, "useFIPS", cfg.UseFIPS)
+
+	logDroppedEnvVars()
+
+	if err := cfg.resolveAPIKey(ctx); err != nil {
+		return nil, fmt.Errorf("resolving API key: %w", err)
+	}
+
```
ge0Aja, line 33: can be one function

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -0,0 +1,167 @@
...
+		return fmt.Errorf("invalid Datadog API key format: expected 32 characters, got %d. Verify your API key at https://app.%s/organization-settings/api-keys", len(c.APIKey), c.Site)
+	}
+
+	slog.Debug("validating Datadog API key")
+
+	client := &http.Client{
+		Timeout: httpClientTimeout,
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: c.SkipSSLValidation,
```
ge0Aja: let's break early if skip SSL is active and we have a Datadog endpoint

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -0,0 +1,167 @@
...
+type resolveOptions struct {
+	AWSCfg aws.Config
+	Value  string
+}
+
+type apiKeyResolver func(ctx context.Context, opts resolveOptions) (string, error)
+
+var retryableStatusCodes = map[int]bool{
+	429: true,
+	500: true,
```
ge0Aja: Imo, the default logic here should be to send to DLT in case of 500s on API key validation

aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -9,15 +9,48 @@ import (
...
+		slogLevel = slog.LevelDebug
+	case "INFO":
+		slogLevel = slog.LevelInfo
+	case "WARNING", "WARN":
+		slogLevel = slog.LevelWarn
+	case "ERROR":
+		slogLevel = slog.LevelError
+	default:
+		slogLevel = slog.LevelInfo
+	}
```
ViBiOh:
```suggestion
	var slogLevel slog.Level

	if err := slogLevel.UnmarshalText([]byte(level)); err != nil {
		slog.LogAttrs(ctx, slog.LevelError, err.Error(), slog.String("level", level))
		slogLevel = slog.LevelInfo
	}
```

aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -9,15 +9,48 @@ import (
...
+	slog.SetDefault(slog.New(slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{
+		Level: slogLevel,
+	})))
+}
+
+func handleRequest(cfg *config.Config) func(context.Context, json.RawMessage) error {
+	return func(ctx context.Context, event json.RawMessage) error {
+		slog.Info("received event", "event", string(event))
```
ViBiOh: **💬 suggestion:** ‏ Avoid the "modulus 2" key/val when using log attribute and prefer using the `slog.<Typed>` thing

```suggestion
		slog.Info("received event", slog.String("event", fmt.Sprintf("%s", event)))
```

aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -9,15 +9,48 @@ import (
...
+
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/config"
 
 	"github.com/aws/aws-lambda-go/lambda"
 )
 
-func handleRequest(ctx context.Context, event json.RawMessage) error {
-	log.Printf("Received event: %s", string(event))
-	return nil
+func initLogger(level string) {
```
ViBiOh: **💬 suggestion:** ‏ You can move the log initialization into its own file or in config package too (because it depends on it)

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -0,0 +1,167 @@
...
+			slog.Debug("API key resolved", "source", resolver.envVar)
+			return nil
+		}
+	}
+	return errors.New("no API key configured: set DD_API_KEY_SECRET_ARN, DD_API_KEY_SSM_NAME, DD_KMS_API_KEY, or DD_API_KEY. See: https://docs.datadoghq.com/serverless/forwarder/")
+}
+
+// Note: the API key verification could fail (e.g. Datadog verification endpoint or network problem)
+// Instead of failing the whole lambda at startup, it should run up to the log sending part and verify the
+// key at this moment, adding the run to the future retry logic in such case.
```
ViBiOh: In such case, I'll add a parameter to no retry on error at startup.

If the API key validation fails at startup, we can take up to 1min (5 * 10s timeout + 5s interval) before we start processing logs.

At the very last moment we need to ensure the API key is valid, let's consume time to retry.

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -0,0 +1,167 @@
...
+		resp, err := client.Get(url)
+		if err != nil {
+			lastErr = err
+			slog.Debug("API key validation request failed, retrying", "attempt", attempt+1, "error", err)
+			time.Sleep(retryBackoffFactor * time.Duration(1<<attempt))
+			continue
+		}
+		if err := resp.Body.Close(); err != nil {
+			slog.Debug("failed to close response body", "error", err)
+		}
```
ViBiOh, line 93: **🥜 nitpick:** ‏ You need to drain the body before closing is, with something like

```go
io.Discard.ReadFrom(resp.Body)
```

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -0,0 +1,167 @@
...
+
+		if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+			slog.Debug("API key validated successfully")
+			return nil
+		}
+
+		if !retryableStatusCodes[resp.StatusCode] {
+			slog.Warn("API key validation failed. Verify your API key is correct and DD_SITE matches your Datadog account region. See: https://docs.datadoghq.com/getting_started/site/",
+				"status", resp.StatusCode,
+				"site", c.Site,
```
ViBiOh: Same comment here regarding slog attribute

aws/logs_monitoring_go/internal/config/apikey_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestResolveAPIKey(t *testing.T) {
+	t.Run("no env vars set", func(t *testing.T) {
```
ViBiOh: **📋 todo:** ‏ Rewrite them to perform table-drive testing https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

Extra nits, use a map and not a slice, Claude usually knows about it

aws/logs_monitoring_go/internal/config/config.go
```diff
@@ -0,0 +1,111 @@
...
+	return n
+}
+
+func logDroppedEnvVars() {
+	for _, name := range deprecatedEnvironmentVariables {
+		if _, ok := os.LookupEnv(name); ok {
+			slog.Warn("deprecated env var set, will be ignored", "name", name)
+		}
+	}
+}
```
ViBiOh: **💬 suggestion:** ‏ Move these function into a dedicated file to not "pollute" the main config file.

The idea is to come to the `config.go` and have a clear understanding on what it does. Those utils function are not primordial.

aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -9,15 +9,48 @@ import (
 	"context"
```
RaphaelAllier, line 9: Generic comment. Please write a PR description of what you're doing an why. This is a public repository so git history and context on changes / technical decisions matters to other contributors

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+var resolvers = []struct {
+	envVar  string
+	resolve apiKeyResolver
+}{
+	{"DD_API_KEY_SECRET_ARN", resolveFromSecretsManager},
+	{"DD_API_KEY_SSM_NAME", resolveFromSSM},
+	{"DD_KMS_API_KEY", resolveFromKMS},
+	{"DD_API_KEY", resolveFromEnv},
```
RaphaelAllier: I think we should be opinionated and enforce a secret-based API key approach here. This is more of a design discussion than code implementation though

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -0,0 +1,167 @@
...
+			})
+			if err != nil {
+				return fmt.Errorf("resolving API key from %s: %w", resolver.envVar, err)
+			}
+			c.APIKey = strings.TrimSpace(key)
+			slog.Debug("API key resolved", "source", resolver.envVar)
+			return nil
+		}
+	}
+	return errors.New("no API key configured: set DD_API_KEY_SECRET_ARN, DD_API_KEY_SSM_NAME, DD_KMS_API_KEY, or DD_API_KEY. See: https://docs.datadoghq.com/serverless/forwarder/")
```
RaphaelAllier: **💭 thought:** ‏Since we're probably going to have a bunch of errors, I wonder if we shouldn't define them as variable consts in a package so we have them all in one place

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -0,0 +1,167 @@
...
+	client := &http.Client{
+		Timeout: httpClientTimeout,
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: c.SkipSSLValidation,
+			},
+		},
+	}
+
+	url := fmt.Sprintf("%s/api/v1/validate?api_key=%s", c.APIURL, c.APIKey)
```
RaphaelAllier: **🥜 nitpick:** ‏String concatenation is faster with "+" then with fmt.Sprintf (It allocated some extra resources)

aws/logs_monitoring_go/internal/config/config.go
```diff
@@ -0,0 +1,111 @@
...
+
+	return cfg, nil
+}
+
+func (c *Config) deriveURLs() {
+	scheme := "https"
+	if c.NoSSL {
+		scheme = "http"
+	}
+	c.IntakeURL = envOrDefault("DD_URL", fmt.Sprintf("%s://http-intake.logs.%s", scheme, c.Site))
```
RaphaelAllier: Same nit comment about string formatting

aws/logs_monitoring_go/cmd/forwarder/main.go
ndakkoune: Will use `slog.Info("received event", slog.String("event", string(event))))`

aws/logs_monitoring_go/internal/config/apikey.go
ndakkoune: Good point.

aws/logs_monitoring_go/internal/config/apikey.go
ndakkoune: Smart. Let's make some of the non-recoverable errors (e.g. missing key) as sentinel errors for the future retry feature/DLQ

aws/logs_monitoring_go/internal/config/apikey.go
ndakkoune: Let's opt for a warning for now and keep the timeout short.

aws/logs_monitoring_go/internal/config/apikey.go
ndakkoune: Going to do with 
```	
defer func() {
    io.Copy(io.Discard, resp.Body)
	resp.Body.Close()
}()
```

aws/logs_monitoring_go/internal/config/apikey.go
ndakkoune: As discussed, we'll probably depreciate this flag due to the security problems it can lead to (e.g. MitM). We'll probably write to in the documentation to setup the `HTTP_PROXY` and `HTTPS_PROXY` flags, as in the CloudFormation template.

ndakkoune: We'll deal with that in the future if we change our minds.

aws/logs_monitoring_go/internal/config/secretsmanager_test.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/secretsmanager"
+)
+
+func GetSecretValueFromSecretsManager(ctx context.Context, api SecretsManagerGetSecretValueAPI, secretId string) (string, error) {
```
ndakkoune: The client mocking has not been fully configured/implemented. It does not compile currently.

aws/logs_monitoring_go/Makefile
```diff
@@ -1,15 +1,12 @@
-.PHONY: build package test lint clean sam-build sam-invoke sam-deploy build-ForwarderFunction
-
 BINARY_NAME := bootstrap
 ZIP_NAME := forwarder.zip
 
+# Go
+.PHONY: build package test lint clean audit
```
ViBiOh: **🥜 nitpick:** ‏ I'd avoid declaring all phone at once, but add a `.PHONY: <name>` before each command.

So that we don't forgot to update a phony on the top list.

aws/logs_monitoring_go/Makefile
```diff
@@ -19,15 +16,26 @@ test:
...
+	go tool staticcheck ./...
+	go tool govulncheck
+
 clean:
 	rm -f $(BINARY_NAME) $(ZIP_NAME)
 
+# SAM
+.PHONY: build-ForwarderFunction sam-build sam-invoke sam-deploy 
+
+build-ForwarderFunction:
```
ViBiOh, line 39: **💬 suggestion:** ‏ Are we allowed to change case here? Can we stick to kebab-case? It's a mix of PascalCase and kebab-case that is disturbing.

aws/logs_monitoring_go/Makefile
```diff
@@ -19,15 +16,26 @@ test:
...
+	GOOS=linux GOARCH=arm64 go build -o $(ARTIFACTS_DIR)/bootstrap ./cmd/forwarder/
```
ViBiOh: Add some option to the linker to reduce the binary size with `-s` and have a self-sufficient binary (bigger binary, but not dependent to external libs)

```suggestion
	GOOS=linux GOARCH=arm64 go build -ldflags="-s" -installsuffix nocgo -o $(ARTIFACTS_DIR)/bootstrap ./cmd/forwarder/
```

aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -8,49 +8,28 @@ package main
...
-		slogLevel = slog.LevelWarn
-	case "ERROR":
-		slogLevel = slog.LevelError
-	default:
-		slogLevel = slog.LevelInfo
+func main() {
+	ctx := context.Background()
+	cfg, err := config.Load(ctx)
+	if err != nil {
+		slog.Error("config load failed", slog.String("error", err.Error()))
```
ViBiOh: **💬 suggestion:** ‏ We might enrich `error` log display later, I'd recommend to log it as any for now, maybe later we'll add stacktrace or better description.

```suggestion
		slog.Error("config load failed", slog.Any("error", err.Error()))
```

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -7,161 +7,82 @@ package config
...
-	envVar  string
-	resolve apiKeyResolver
-}{
-	{"DD_API_KEY_SECRET_ARN", resolveFromSecretsManager},
-	{"DD_API_KEY_SSM_NAME", resolveFromSSM},
-	{"DD_KMS_API_KEY", resolveFromKMS},
-	{"DD_API_KEY", resolveFromEnv},
-}
+const (
+	httpClientTimeout = 3 * time.Second
```
ViBiOh: I think it's a bit optimistic, it was 5 seconds in the Python one, I think we can keep this value.

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -7,161 +7,82 @@ package config
...
-	if c.APIKey == "" || c.APIKey == "<YOUR_DATADOG_API_KEY>" {
-		return errors.New("missing Datadog API key. Set DD_API_KEY environment variable. See: https://docs.datadoghq.com/serverless/forwarder/")
+func (c *Config) validateAPIKey(ctx context.Context) error {
+	if c.APIKey == "" {
+		return fmt.Errorf("%w: set DD_API_KEY_SECRET_ARN, DD_API_KEY_SSM_NAME or DD_KMS_API_KEY. See: https://docs.datadoghq.com/serverless/forwarder/", ErrMissingAPIKey)
 	}
 
 	if len(c.APIKey) != 32 {
-		return fmt.Errorf("invalid Datadog API key format: expected 32 characters, got %d. Verify your API key at https://app.%s/organization-settings/api-keys", len(c.APIKey), c.Site)
+		return fmt.Errorf("%w: expected 32 characters, got %d. Verify your API key at https://app.%s/organization-settings/api-keys", ErrInvalidAPIKey, len(c.APIKey), c.Site)
```
ViBiOh: The wrapped error is usually at the end of the error string

```suggestion
		return fmt.Errorf("expected 32 characters, got %d. Verify your API key at https://app.%s/organization-settings/api-keys: %w", len(c.APIKey), c.Site, ErrInvalidAPIKey)
```

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -7,161 +7,82 @@ package config
...
-	client := &http.Client{
-		Timeout: httpClientTimeout,
-		Transport: &http.Transport{
-			TLSClientConfig: &tls.Config{
-				InsecureSkipVerify: c.SkipSSLValidation,
-			},
-		},
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, c.APIURL+"/api/v1/validate", nil)
+	if err != nil {
+		slog.Warn("failed to build API key validation request", slog.String("error", err.Error()))
```
ViBiOh: Same here regarding error field

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -7,161 +7,82 @@ package config
...
-		slog.Warn("API key validation failed after retries, continuing anyway", "attempts", maxRetries, "error", lastErr)
-	} else {
-		slog.Warn("API key validation failed after retries. Verify your API key is correct and DD_SITE matches your Datadog account region. See: https://docs.datadoghq.com/getting_started/site/",
-			"attempts", maxRetries,
-			"lastStatus", lastStatus,
-			"site", c.Site,
-		)
+	defer func() {
+		io.Copy(io.Discard, resp.Body)
+		resp.Body.Close()
```
ViBiOh: **💬 suggestion:** ‏ Handle the error on the `Close()` here, you can just log it, but sometimes it can help getting clue.

aws/logs_monitoring_go/Makefile
ndakkoune: This command is (and should only be) used by the `sam build` command, which use whatever is written after `build-` to create a function named ForwarderFunction in the cloudformation template (local).

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -0,0 +1,101 @@
...
+		client, err := c.createSecretsManagerAPIClient(ctx, awsCfg)
+		if err != nil {
+			return fmt.Errorf("creating Secrets Manager client: %w", err)
+		}
+		apiKey, err := resolveFromSecretsManager(ctx, client, v)
+		if err != nil {
+			return fmt.Errorf("resolving from secrets manager: %w", err)
+		}
+		c.APIKey = apiKey
+		return nil
```
ViBiOh, line 49: nitpick: usually after a error condition, leave a new line to "unpack" your code and make it easier to read.

```suggestion
		client, err := c.createSecretsManagerAPIClient(ctx, awsCfg)
		if err != nil {
			return fmt.Errorf("creating Secrets Manager client: %w", err)
		}

		apiKey, err := resolveFromSecretsManager(ctx, client, v)
		if err != nil {
			return fmt.Errorf("resolving from secrets manager: %w", err)
		}

		c.APIKey = apiKey
		return nil
```

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -0,0 +1,101 @@
...
+	if v, ok := os.LookupEnv("DD_KMS_API_KEY"); ok {
+		return c.resolveFromKMS(ctx, awsCfg, v)
+	}
+
+	return errors.New("no API key configured: set DD_API_KEY_SECRET_ARN, DD_API_KEY_SSM_NAME or DD_KMS_API_KEY. See: https://docs.datadoghq.com/serverless/forwarder/")
+}
+
+func (c *Config) validateAPIKey(ctx context.Context) error {
+	if c.APIKey == "" {
+		return fmt.Errorf("%w: set DD_API_KEY_SECRET_ARN, DD_API_KEY_SSM_NAME or DD_KMS_API_KEY. See: https://docs.datadoghq.com/serverless/forwarder/", ErrMissingAPIKey)
```
ViBiOh:
```suggestion
		return fmt.Errorf("set DD_API_KEY_SECRET_ARN, DD_API_KEY_SSM_NAME or DD_KMS_API_KEY. See: https://docs.datadoghq.com/serverless/forwarder/: %w", ErrMissingAPIKey)
```

aws/logs_monitoring_go/internal/config/config_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/google/go-cmp/cmp/cmpopts"
+)
+
+func TestLoadConfig(t *testing.T) {
+	tests := map[string]struct {
+		env  map[string]string
+		want Config
+	}{
+		"default": {
+			env: map[string]string{},
```
ViBiOh:
```suggestion
```

aws/logs_monitoring_go/internal/config/config_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+)
+
+func TestLoadConfig(t *testing.T) {
+	tests := map[string]struct {
```
ViBiOh, line 16:
```suggestion
func TestLoadConfig(t *testing.T) {
	t.Parallel()
	
	tests := map[string]struct {
```

aws/logs_monitoring_go/internal/config/config_test.go
```diff
@@ -0,0 +1,64 @@
...
+				IntakeURL: "https://custom-intake.example.com",
+				APIURL:    "https://api.datadoghq.com",
+				LogLevel:  "INFO",
+			},
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			for k, v := range tc.env {
```
ViBiOh, line 54:
```suggestion
		t.Run(name, func(t *testing.T) {
			t.Parallel()

			for k, v := range tc.env {
```

aws/logs_monitoring_go/internal/config/environment.go
```diff
@@ -0,0 +1,55 @@
...
+	"DD_FETCH_STEP_FUNCTIONS_TAGS",
+	"DD_COMPRESSION_LEVEL",
+	"DD_FORWARD_LOG",
+	"DD_NO_SSL",
+	"DD_PORT",
+	"DD_SKIP_SSL_VALIDATION",
+	"DD_TAGS_CACHE_TTL_SECONDS",
+	"DD_TRACE_INTAKE_URL",
+	"DD_USE_COMPRESSION",
+	"DD_USE_VPC",
```
ViBiOh, line 31: **🥜 nitpick:** ‏ Can you alpha-sort them?

aws/logs_monitoring_go/internal/config/environment.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	return fallback
+}
+
+func envOrDefaultBool(key string, fallback bool) bool {
+	v, ok := os.LookupEnv(key)
+	if !ok {
+		return fallback
+	}
+	return strings.EqualFold(v, "true")
```
ViBiOh: **💭 thought:** ‏ I'm wondering if we should support the "1" value etc?

You might prefere using [ParseBool](https://pkg.go.dev/strconv#ParseBool) that will handle everything for you.

aws/logs_monitoring_go/internal/config/environment_test.go
```diff
@@ -0,0 +1,64 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2026-Present Datadog, Inc.
+
+package config
+
+import "testing"
+
+func TestEnvOrDefault(t *testing.T) {
```
ViBiOh, line 10: Same comments here regarding usage of `t.Parallel`

aws/logs_monitoring_go/internal/config/secretsmanager.go
```diff
@@ -0,0 +1,58 @@
...
+func resolveFromSecretsManager(ctx context.Context, client SecretsManagerAPIClient, arn string) (string, error) {
+	result, err := client.GetSecretValue(ctx, &secretsmanager.GetSecretValueInput{
+		SecretId: aws.String(arn),
+	})
+	if err != nil {
+		return "", fmt.Errorf("fetching secret %s: %w", arn, err)
+	}
+	if result.SecretString == nil {
+		return "", fmt.Errorf("secret %s has no string value", arn)
+	}
```
ViBiOh, line 56: **🥜 nitpick:** ‏ When putting string value into error message, surround them by backtick so we know exactly where it starts/ends.

```suggestion
	if err != nil {
		return "", fmt.Errorf("fetching secret `%s`: %w", arn, err)
	}

	if result.SecretString == nil {
		return "", fmt.Errorf("secret `%s` has no string value", arn)
	}
```

aws/logs_monitoring_go/internal/config/secretsmanager.go
```diff
@@ -0,0 +1,58 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2026-Present Datadog, Inc.
+
+package config
+
+//go:generate go tool mockgen -source=secretsmanager.go -package=config -destination=secretsmanager_mockgen.go
```
ViBiOh, line 8: We'll move the mock to a dedicated package later I think, to not have be part of the production bundle. Ok for now

ViBiOh (approved):
Nothing blocker, good if you can do it before merging

aws/logs_monitoring_go/internal/config/config_test.go
ndakkoune: `t.Parallel()` cannot be used with `t.Setenv(k,v)` since the latter affects the whole process.

aws/logs_monitoring_go/internal/config/config_test.go
ndakkoune: Same as [the previous comment](https://github.com/DataDog/datadog-serverless-functions/pull/1084#discussion_r2993288624).

aws/logs_monitoring_go/internal/config/environment.go
ndakkoune: We lose nothing by doing so. Let's use it.

aws/logs_monitoring_go/internal/config/environment_test.go
ndakkoune: Same as [here](https://github.com/DataDog/datadog-serverless-functions/pull/1084#discussion_r2993288624.).

ViBiOh (approved):

PR #1088, opened 2026-03-27
[AWSX-3190] feat(go-forwarder): log model and parsing logic

ndakkoune:
### What does this PR do?
Set up log parsing and model logic.

### Motivation
Following the migration of the Lambda Log Forwarder to Go, the log model is a necessary and critical step that will dictate how the data flow. 

### Additional Notes
The natural divergence of both the log events received by the Lambda and the outbound payload sent to Datadog needs a model as generic as possible. The most noticeable design choices are :
- The `CloudwatchLogsContext` and `CloudwatchLogsLogEvent` types that closely mimic some of their equivalents in the `github.com/aws/aws-lambda-go/events` package.
- The `LogContent` interface is responsible to return its own `Message()`, required in the future steps (e.g. scrubbing) and `MarshalFields()`, responsible to return the top-level fields to match the intake API format (different for each event source for S3 invocations, especially for CloudTrail).
- `LogEntry.MarshalJSON()` is the custom marshaler that merges `Content.MarshalFields()` so that each `LogEntry` is serialized to the correct format.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


aws/logs_monitoring_go/internal/config/config.go
```diff
@@ -11,13 +11,18 @@ import (
 	"log/slog"
 )
 
+var ForwarderVersion = "dev"
```
ndakkoune, line 14: Used for `ldflags`

aws/logs_monitoring_go/internal/parsing/parsing.go
```diff
@@ -0,0 +1,55 @@
...
+	case invokationSourceCloudWatch:
+		parseCloudWatch(ctx, event, cfg, out)
+	default:
+		slog.Error("unsupported invokation source")
+	}
+}
+
+// detectInvokationSource inspects the raw JSON to determine the invokation type.
+func detectInvokationSource(event json.RawMessage) invokationSource {
+	var probe struct {
```
ndakkoune: Even though this is not very clean, I don't see another way to detect the invokation type apart from unmarshalling the whole JSON (way more expensive).

aws/logs_monitoring_go/internal/parsing/parsing.go
```diff
@@ -0,0 +1,55 @@
...
+		AWSLogs *json.RawMessage `json:"awslogs"`
+	}
+
+	if err := json.Unmarshal(event, &probe); err != nil {
+		slog.Error("failed unmarshal", slog.Any("error", err))
+		return invokationSourceUnknown
```
ndakkoune: Will log a double error from both the unmarshalling + unknown source. Fair ?

aws/logs_monitoring_go/internal/parsing/parsing.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"context"
+	"encoding/json"
+	"log/slog"
+
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/config"
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/model"
+)
+
+type invokationSource int
```
ndakkoune: Decided to replace the old `AWSEventSource` by `invokationSource`, which is way more explicit.

aws/logs_monitoring_go/internal/model/log.go
```diff
@@ -0,0 +1,52 @@
...
+	"encoding/json"
+)
+
+// LogEntry represents a single processed log.
+type LogEntry struct {
+	Source  string
+	Service string
+	Host    string
+	Tags    string
+	AWS     *AWSMetadata
```
ndakkoune: Pointer here because its shared between all log entries.

aws/logs_monitoring_go/internal/config/config.go
```diff
@@ -11,13 +11,18 @@ import (
...
-	UseFIPS   bool
+	APIKey     string
+	Site       string
+	IntakeURL  string
+	APIURL     string
+	LogLevel   string
+	UseFIPS    bool
+	Source     string
+	Host       string
+	CustomTags string
```
ViBiOh, line 25: **💬 suggestion:** ‏ We'll mainly used them as a slice of string, so it might be interesting to convert the string directly to a slice when loading the config.

aws/logs_monitoring_go/internal/model/log.go
ViBiOh: It doesn't need to be a pointer for this, we can copy it on each LogEntry. it will not escape to the heap and might be more efficient in the end.

aws/logs_monitoring_go/internal/parsing/parsing.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// Parse detects the invocation source and sends parsed log entries to the out channel.
+// Errors are logged and skipped, unless the lambda timeout is reached.
+func Parse(ctx context.Context, event json.RawMessage, cfg *config.Config, out chan<- model.LogEntry) {
+	switch detectInvocationSource(event) {
+	case invocationSourceCloudwatchLogs:
+		parseCloudwatchLogs(ctx, event, cfg, out)
+	default:
+		slog.Error("unsupported invocation source")
```
ViBiOh: **💬 suggestion:** ‏ Add the event as a log field, otherwise it will be complex to debug.

aws/logs_monitoring_go/internal/parsing/parsing.go
```diff
@@ -0,0 +1,55 @@
...
+func detectInvocationSource(event json.RawMessage) invocationSource {
+	var probe struct {
+		AWSLogs *json.RawMessage `json:"awslogs"`
+	}
+
+	if err := json.Unmarshal(event, &probe); err != nil {
+		slog.Error("failed unmarshal", slog.Any("error", err))
+		return invocationSourceUnknown
+	}
+
```
ViBiOh: Benchmark this and try the decoder approach

aws/logs_monitoring_go/internal/config/config.go
ndakkoune: I agree for `CustomTags` (and maybe `UseFIPS`) since we'll append all the specified tags (e.g. `env:prod,team:aws`) along with other tags such as `forwarder_version`. However, `Source` and `Host` belong other top-level fields (i.e. `ddsource:X` and `host:Y`) and I don't see any benefit from that.

aws/logs_monitoring_go/internal/parsing/parsing.go
ndakkoune: Will make a separate PR for this. Thank you!

aws/logs_monitoring_go/internal/model/cloudwatchlogs.go
```diff
@@ -0,0 +1,32 @@
...
+	LogStream string `json:"logStream"`
+	Owner     string `json:"owner"`
+}
+
+func (CloudwatchLogsContext) InvocationKey() string { return "awslogs" }
+
+type CloudwatchLogsContent struct {
+	ID        string `json:"id"`
+	Timestamp int64  `json:"timestamp"`
+	Msg       string `json:"message"`
```
ge0Aja: nit: prefer to have complete variable names

aws/logs_monitoring_go/internal/model/cloudwatchlogs.go
```diff
@@ -0,0 +1,32 @@
...
+type CloudwatchLogsContext struct {
+	LogGroup  string `json:"logGroup"`
+	LogStream string `json:"logStream"`
+	Owner     string `json:"owner"`
+}
+
+func (CloudwatchLogsContext) InvocationKey() string { return "awslogs" }
+
+type CloudwatchLogsContent struct {
+	ID        string `json:"id"`
```
ge0Aja: what's the ID here ?

aws/logs_monitoring_go/internal/model/cloudwatchlogs.go
```diff
@@ -0,0 +1,32 @@
...
+	Timestamp int64  `json:"timestamp"`
+	Msg       string `json:"message"`
+}
+
+func (c CloudwatchLogsContent) Message() string {
+	return c.Msg
+}
+
+func (c CloudwatchLogsContent) MarshalFields() (map[string]any, error) {
```
ge0Aja: nit: Marshal gives the impression that the output should be a json string, instead we're getting a map. we should either rename the function, or implement as it says

aws/logs_monitoring_go/internal/model/log.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"encoding/json"
+)
+
+type LogEntry struct {
+	Source  string
+	Service string
+	Host    string
+	Tags    []string
+	AWS     AWSMetadata
```
ge0Aja: please rename the variable to match the type suggestion: Metadata

aws/logs_monitoring_go/internal/model/log.go
```diff
@@ -0,0 +1,23 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2026-Present Datadog, Inc.
+
+package model
+
+type LogEntry struct {
+	ID        string      `json:"id,omitempty"`
+	Timestamp int64       `json:"timestamp,omitempty"`
```
ndakkoune: These two lines are CloudwatchLogs specific. We'll probably have to put them as pointers in the future, so that they won't appear with their zero values when Marshalling. The other possibility would be to detect the zero values in the backend to deal with them properly (e.g. `timestamp: 0` => impossible => generate current timestamp).

aws/logs_monitoring_go/internal/model/log.go
```diff
@@ -0,0 +1,23 @@
...
+	Service   string      `json:"service,omitempty"`
+	Host      string      `json:"hostname,omitempty"`
+	Tags      []string    `json:"ddtags"`
+	AWS       AWSMetadata `json:"aws"`
+}
+
+type AWSMetadata struct {
+	InvokedFunctionARN string                 `json:"invoked_function_arn"`
+	FunctionVersion    string                 `json:"function_version,omitempty"`
+	CloudwatchLogs     *CloudwatchLogsContext `json:"awslogs,omitempty"`
```
ndakkoune: Adding the context here avoids us to write a custom MarshalJSON. Since there is only two invocation sources (i.e. CW and S3), it seems a good compromise.

aws/logs_monitoring_go/internal/parsing/parsing.go
```diff
@@ -0,0 +1,52 @@
...
+	switch detectInvocationSource(event) {
+	case invocationSourceCloudwatchLogs:
+		parseCloudwatchLogs(ctx, event, cfg, out)
+	default:
+		slog.Error("unsupported invocation source", slog.Any("event", event))
+	}
+}
+
+func detectInvocationSource(event json.RawMessage) invocationSource {
+	var probe struct {
```
ndakkoune: Will benchmark this and probably try the `Decoder` strategy.

aws/logs_monitoring_go/internal/model/log.go
ge0Aja: regarding CloudwatchLogs specific. I thought we agreed to have a separate struct for cloudwatch logs, s3 records, and events. which should remove the need for optional fields.

also , did you check the timestamp format ? is it a linux timestamp or a different format ?

aws/logs_monitoring_go/internal/model/log.go
ge0Aja: if this is to be shared then I don't think we should adda a cloudwatch logs specific context here.

aws/logs_monitoring_go/internal/parsing/parsing.go
```diff
@@ -0,0 +1,52 @@
...
+	"log/slog"
+
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/config"
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/model"
+)
+
+type invocationSource int
+
+const (
+	invocationSourceUnknown invocationSource = iota
```
ge0Aja, line 21: could we have this case ? unknown invocation source ? I think the default case in the switch should be enough

aws/logs_monitoring_go/internal/parsing/parsing.go
ge0Aja: yes, we did that for cloudtrail events if you're looking for specific keywords at the beginning of the raw json

aws/logs_monitoring_go/internal/model/log.go
ndakkoune: The timestamps provided by AWS are RFC3339, the Datadog intake endpoints ones expect unix epoch times.

aws/logs_monitoring_go/internal/parsing/parsing.go
ndakkoune: Nothing prevent a customer to create a trigger from an event source we do not support (e.g. Alexa, API Gateway, CloudFront, AppSync).

ndakkoune: Btw the 0 value of iota would become the Cloudwatch one, which could be thought as a default case (we could `_ invocationSource = iota` to skip this).

aws/logs_monitoring_go/internal/model/log.go
ndakkoune: Rectifying: Cloudwatch events are also unix epoch time, but S3/SQS/... are still RFC3339

aws/logs_monitoring_go/internal/config/config.go
```diff
@@ -11,13 +11,18 @@ import (
...
-	APIURL    string
-	LogLevel  string
-	UseFIPS   bool
+	APIKey     string
+	Site       string
+	IntakeURL  string
+	APIURL     string
+	LogLevel   string
+	UseFIPS    bool
+	Source     string
```
ge0Aja, line 23: forwarder level source override (for all logs) if set by the customer

aws/logs_monitoring_go/internal/parsing/parsing.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	if probe.AWSLogs != nil {
+		return invocationSourceCloudwatchLogs
+	}
+
+	return invocationSourceUnknown
+}
+
+func parseCloudwatchLogs(ctx context.Context, event json.RawMessage, cfg *config.Config, out chan<- model.CloudwatchLogEntry) {
```
ge0Aja, line 104: suggestion: use emit function callback instead of passing the channel explicitly.

ge0Aja (approved):

PR #1094, opened 2026-04-01
[AWSINTS-3450] feat(go-forwarder): add CloudwatchLogEntry creation logic

ndakkoune:
### What does this PR do?
Add (Cloudwatch)LogEntry creation logic (source detection, overriding tags, etc.)

### Testing Guidelines
Any for now, quick feedback before writing the tests.

### Additional Notes
Except from the `CONTROL_MESSAGE` skipping, the behavior is roughly the same as the Python one.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


aws/logs_monitoring_go/internal/parsing/cloudwatchlogs.go
```diff
@@ -0,0 +1,131 @@
...
+
+	data, err := cwEvent.AWSLogs.Parse()
+	if err != nil {
+		slog.Error("failed to decompress cloudwatch data", slog.Any("error", err))
+		return
+	}
+
+	if data.MessageType == "CONTROL_MESSAGE" {
+		return
+	}
```
ndakkoune, line 50: Not present in the Python implementation. the control message are probing messages from cloudwatch to verify that the lambda is reachable => noise.

aws/logs_monitoring_go/internal/parsing/cloudwatchlogs.go
```diff
@@ -0,0 +1,131 @@
...
+	"log/slog"
+	"strings"
+
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/config"
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/model"
+	"github.com/aws/aws-lambda-go/events"
+	"github.com/aws/aws-lambda-go/lambdacontext"
+)
+
+func parseCloudwatchLogs(ctx context.Context, event json.RawMessage, cfg *config.Config, out chan<- model.CloudwatchLogEntry) {
```
ViBiOh: **💬 suggestion:** ‏The naming of the function is a bit confusing. Usually a parser is a "pure" function, you give an input, you get an output and an error.

Here, the error are logged directly and we output to the chan. I think we should make this function "pure" (easier to test, because currently untested) and let the `Parse` function that call it deal with the error and forwarding to the chan

Edit: I see below that we have the log entry in memory and we stream them, so maybe just an intermediate function `handleCloudwatchLogs` that read the parse, and then stream

aws/logs_monitoring_go/internal/parsing/cloudwatchlogs.go
```diff
@@ -0,0 +1,131 @@
...
+			Host:      host,
+			Tags:      tags,
+			AWS:       metadata,
+		}
+
+		select {
+		case out <- entry:
+		case <-ctx.Done():
+			return
+		}
```
ViBiOh: **🥜 nitpick:** ‏While it "would" work as expected, it's not deterministic. Have a look at the internal repository, there is a `SafeSender` method you can copy/paste in an util package here

aws/logs_monitoring_go/internal/parsing/cloudwatchlogs.go
```diff
@@ -0,0 +1,131 @@
...
+	var source string
+	if strings.Contains(logStream, "_CloudTrail_") {
+		source = "cloudtrail"
+	} else {
+		source = getSourceFromLogGroup(strings.ToLower(logGroup))
+	}
+
+	if strings.HasPrefix(logStream, "states/") {
+		source = "stepfunction"
+	}
```
ViBiOh: **💬 suggestion:** ‏ You can move this after the source override and change to a definitive `return` (doing a fast-exit)

aws/logs_monitoring_go/internal/parsing/cloudwatchlogs.go
```diff
@@ -0,0 +1,131 @@
...
+
+	if lambdacontext.FunctionVersion != "$LATEST" {
+		metadata.FunctionVersion = lambdacontext.FunctionVersion
+	}
+
+	if lc, ok := lambdacontext.FromContext(ctx); ok {
+		metadata.InvokedFunctionARN = lc.InvokedFunctionArn
+	} else {
+		slog.Warn("failed lambda context loading")
```
ViBiOh: **❓ question:** ‏What's the actionable thing to do there? We should either enrich with some field or remove the log

ndakkoune: This one is a safety net. The only ways the else case is hit is (1) when the code is not ran from AWS Lambda (e.g. testing) or (2) someone passes a plain `context.Background()`. It should NEVER be hit in production, and if it does, the default string value "" will be sent in the backend (we would notice it with a `status:skipped`).

aws/logs_monitoring_go/internal/parsing/cloudwatchlogs.go
ndakkoune: Great. I kept the channel approach with only one goroutine for cloudwatch. The S3 Handler will probably trigger more goroutines and leverage streaming.

aws/logs_monitoring_go/internal/model/tags.go
```diff
@@ -0,0 +1,17 @@
...
+// Copyright 2026-Present Datadog, Inc.
+
+package model
+
+import (
+	"encoding/json"
+	"strings"
+)
+
+type Tags []string
```
ndakkoune, line 13: This little workaround is necessary due to constraints in the backend side.

`json.MarshalJSON` converts a slice to a JSON array (expected behavior), but the backend needs a string of comma-separated key:value pairs, hence this custom MarshalJSON.

aws/logs_monitoring_go/internal/pipeline/pipeline.go
```diff
@@ -0,0 +1,40 @@
...
+	var g errgroup.Group
+
+	entries := make(chan T)
+
+	g.Go(func() error {
+		defer close(entries)
+		return handler(ctx, event, cfg, entries)
+	})
+
+	g.Go(func() error {
```
ge0Aja: What's the purpose of this ? if we want some debugging logs you could just do it in the handler. 
I'm also assessing the need for an error group in the first place

aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -26,10 +28,14 @@ func main() {
...
-// cfg not used for now, will be when forwarding logic added
 func handleRequest(cfg *config.Config) func(context.Context, json.RawMessage) error {
 	return func(ctx context.Context, event json.RawMessage) error {
-		slog.Info("received event", slog.String("event", string(event)))
-		return nil
+		switch parsing.DetectInvocationSource(event) {
+		case parsing.InvocationSourceCloudwatchLogs:
+			return pipeline.Run(ctx, event, cfg, parsing.HandleCloudwatchLogs)
+		default:
+			slog.Error("unsupported invocation source")
```
ViBiOh: **⚠️ issue:** ‏ You should store the dectected invocation source as a variable and put in the log. In current scheme, the log is not actionable / hard du debug.

aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -16,29 +16,45 @@ import (
...
-		}
+	for range numWorkers {
+		g.Go(func() error {
+			for {
+				body, ok, err := concurrent.SafeReader(ctx, in)
+				if err != nil {
+					return err
+				}
+				if !ok {
+					return nil
```
ViBiOh, line 47: **⚠️ issue:** ‏There is a concurrency issue in this function.

We have 3 workers. In any worker encounter an error, it exits and stop, stopping to dequeue from the input channel.

If the 3 workers fail, the `in` chan will be full, nothing will dequeue it, introducing a deadlock for the upstream goroutine.

The only call of Forward is [here](https://github.com/DataDog/datadog-serverless-functions/blob/master/aws/logs_monitoring_go/internal/pipeline/pipeline.go#L47) but the `ctx` is not managed by the `errgroup.Group`. So if the goroutine fails, there is no signal sent to other goroutine to stop.

We should fix the context cancellation in the pipeline file and maybe leave a TODO here to handle the retry mechanism (later in the migration).

aws/logs_monitoring_go/internal/parsing/cloudwatchlogs.go
ViBiOh: You should change the logged message to this explanation.

My point is: if a customer see that in 6 months, the log should explain the situation clearly.

aws/logs_monitoring_go/internal/parsing/cloudwatchlogs.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/config"
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/model"
+	"github.com/aws/aws-lambda-go/events"
+	"github.com/aws/aws-lambda-go/lambdacontext"
+)
+
+func HandleCloudwatchLogs(ctx context.Context, event json.RawMessage, cfg *config.Config, out chan<- model.CloudwatchLogEntry) error {
+	logEntries, err := parseCloudwatchLogs(ctx, event, cfg)
+	if err != nil {
+		slog.Error("failed parse cloudwatch logs", slog.Any("error", err))
```
ViBiOh: **❓ question:** ‏We don't return after the parsing error. Should we return the error and let the caller log it, or return nil.

The `logEntries` below will very likely to be nil/empty, so we will return `nil` in the end. I think we should return the error. If we can't parse the cloudwatch message, it's very likely to be a broad failure and not pretend everything was fine if you don't look at the logs.

aws/logs_monitoring_go/internal/parsing/cloudwatchlogs.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+	}
+
+	return nil
+}
+
+func parseCloudwatchLogs(ctx context.Context, event json.RawMessage, cfg *config.Config) ([]model.CloudwatchLogEntry, error) {
+	var cwEvent events.CloudwatchLogsEvent
+	if err := json.Unmarshal(event, &cwEvent); err != nil {
+		return nil, err
```
ViBiOh:
```suggestion
		return nil, fmt.Errorf("unmarshal: %w", err)
```

aws/logs_monitoring_go/internal/parsing/cloudwatchlogs.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	data, err := cwEvent.AWSLogs.Parse()
+	if err != nil {
+		return nil, err
```
ViBiOh:
```suggestion
		return nil, fmt.Errorf("parse: %w", err)
```

aws/logs_monitoring_go/internal/parsing/cloudwatchlogs.go
```diff
@@ -0,0 +1,142 @@
...
+		return "stepfunction"
+	}
+
+	var source string
+	if strings.Contains(logStream, "_CloudTrail_") {
+		source = "cloudtrail"
+	} else {
+		source = getSourceFromLogGroup(strings.ToLower(logGroup))
+	}
+	return source
```
ViBiOh: Same here regarding "fast-exit", if you have your answer, exit the function directly. This avoid nesting instruction and give a clear execution path of the function.

aws/logs_monitoring_go/internal/parsing/parsing.go
```diff
@@ -7,100 +7,83 @@ package parsing
...
-	"context"
 	"encoding/json"
-	"log/slog"
-
-	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/config"
-	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/model"
 )
 
-type invocationSource int
+//go:generate stringer -type InvocationSource -trimprefix InvocationSource
```
ViBiOh, line 13: **👏 praise:** ‏ 👍

aws/logs_monitoring_go/internal/parsing/tags.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/model"
+	"github.com/aws/aws-lambda-go/lambdacontext"
+)
+
+func getTagsAndService(cfg config.Config) (model.Tags, string) {
+	var tags model.Tags
+	var service string
+
+	if cfg.CustomTags != "" {
+		for _, tag := range strings.Split(cfg.CustomTags, ",") {
```
ViBiOh: I think your code editor will suggest to use SplitSeq

aws/logs_monitoring_go/internal/pipeline/pipeline.go
```diff
@@ -0,0 +1,53 @@
...
+
+const timeout = 10 * time.Second
+
+func Run[T any](
+	ctx context.Context,
+	event json.RawMessage,
+	cfg *config.Config,
+	handler func(context.Context, json.RawMessage, *config.Config, chan<- T) error,
+) error {
+	g, ctx := errgroup.WithContext(ctx) // Fragile because if one goroutine fails, all stages will stop gracefully
```
ViBiOh, line 28: **📝 note:** ‏ I don't think it's "fragile" in this case. It's the graceful shutdown management and ensure that there won't be any deadlock of one goroutine quitting listening to a chan , leaving the others stuck

ndakkoune: I wrote this comment as a self memory note for the future, where we probably won stop all the goroutines if, for example, a forwarding worker fails.

aws/logs_monitoring_go/internal/parsing/cloudwatchlogs.go
ndakkoune: I'm going for the `return err`.

aws/logs_monitoring_go/internal/parsing/tags.go
```diff
@@ -6,24 +6,27 @@
...
 	var tags model.Tags
 	var service string
 
 	if cfg.CustomTags != "" {
 		for tag := range strings.SplitSeq(cfg.CustomTags, ",") {
 			if strings.HasPrefix(tag, "service:") {
-				if service != "" {
-					continue
+				if service == "" {
+					service = tag[8:]
```
ndakkoune, line 27: More explicit with constants

ViBiOh (approved):

PR #1099, opened 2026-04-02
[AWSINTS-3462] feat(go-forwarder): add scrubbing

ndakkoune:
### What does this PR do?
Add the scrubbing implementation + tests

### Testing Guidelines
make test

### Additional Notes
- Kept the same regex matches as the Python implementation
- `func (s Scrubber) ScrubMessage(msg string) string` as decoupled as possible.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring_go/internal/config/config.go
+++ b/aws/logs_monitoring_go/internal/config/config.go
@@ -23,6 +23,14 @@ type Config struct {
 	Source     string
 	Host       string
 	CustomTags string
+	Scrubbing  ScrubbingConfig
+}
+
+type ScrubbingConfig struct {
+	ScrubIP    bool
+	ScrubEmail bool
+	CustomRule        string
+	CustomReplacement string
 }
 
 func Load(ctx context.Context) (*Config, error) {
@@ -54,6 +62,12 @@ func loadConfig() *Config {
 		Source:     envOrDefault("DD_SOURCE", ""),
 		Host:       envOrDefault("DD_HOST", ""),
 		CustomTags: envOrDefault("DD_TAGS", ""),
+		Scrubbing: ScrubbingConfig{
+			ScrubIP:    envOrDefaultBool("REDACT_IP", false),
+			ScrubEmail: envOrDefaultBool("REDACT_EMAIL", false),
+			CustomRule:        envOrDefault("DD_SCRUBBING_RULE", ""),
+			CustomReplacement: envOrDefault("DD_SCRUBBING_RULE_REPLACEMENT", ""),
+		},
 	}
 }
 
@@ -64,5 +78,8 @@ func (c Config) LogValue() slog.Value {
 		slog.String("apiUrl", c.APIURL),
 		slog.String("loglevel", c.LogLevel),
 		slog.Bool("fips", c.UseFIPS),
+		slog.Bool("redactIP", c.Scrubbing.ScrubIP),
+		slog.Bool("redactEmail", c.Scrubbing.ScrubEmail),
+		slog.Bool("customScrubbing", c.Scrubbing.CustomRule != ""),
 	)
 }
--- a/aws/logs_monitoring_go/internal/config/config.go
+++ b/aws/logs_monitoring_go/internal/config/config.go
@@ -27,8 +27,8 @@ type Config struct {
 }
 
 type ScrubbingConfig struct {
-	ScrubIP    bool
-	ScrubEmail bool
+	ScrubIP           bool
+	ScrubEmail        bool
 	CustomRule        string
 	CustomReplacement string
 }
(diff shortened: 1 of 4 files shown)


aws/logs_monitoring_go/internal/config/config.go
```diff
@@ -23,6 +23,14 @@ type Config struct {
...
 	Host       string
 	CustomTags string
+	Scrubbing  ScrubbingConfig
+}
+
+type ScrubbingConfig struct {
+	ScrubIP    bool
+	ScrubEmail bool
+	CustomRule        string
+	CustomReplacement string
```
ViBiOh, line 33: **🥜 nitpick:** ‏ Use `gofumpt` to format your code (align the type), same below

aws/logs_monitoring_go/internal/config/config.go
```diff
@@ -54,6 +62,12 @@ func loadConfig() *Config {
 		Source:     envOrDefault("DD_SOURCE", ""),
 		Host:       envOrDefault("DD_HOST", ""),
 		CustomTags: envOrDefault("DD_TAGS", ""),
+		Scrubbing: ScrubbingConfig{
+			ScrubIP:    envOrDefaultBool("REDACT_IP", false),
+			ScrubEmail: envOrDefaultBool("REDACT_EMAIL", false),
+			CustomRule:        envOrDefault("DD_SCRUBBING_RULE", ""),
+			CustomReplacement: envOrDefault("DD_SCRUBBING_RULE_REPLACEMENT", ""),
```
ViBiOh, line 69: **❓ question:** ‏The default for scrubbed value is blank? I can be a bit confusing but if it's the existing behavior I'm ok with that.

aws/logs_monitoring_go/internal/processing/scrubbing_test.go
```diff
@@ -0,0 +1,118 @@
...
+package transform
+
+import (
+	"testing"
+
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/config"
+)
+
+func TestNewScrubber(t *testing.T) {
+	tests := map[string]struct {
```
ViBiOh, line 17: Same below

```suggestion
func TestNewScrubber(t *testing.T) {
	t.Parallel()

	tests := map[string]struct {
```

aws/logs_monitoring_go/internal/config/config.go
ndakkoune: Thank you, I've configured my editor to do that on save

aws/logs_monitoring_go/internal/config/config.go
ndakkoune: Yes, the default replacement is an empty string (same existing Python behavior). If DD_SCRUBBING_RULE is set without DD_SCRUBBING_RULE_REPLACEMENT, matched patterns are simply removed from the log. 

Btw really important note that I did not highlight: the regex **have now to be RE2**.

What does that mean ? Basically, the underlying engine of Go's regexp package uses a linear-time and deterministic engine called RE2 which guarantees a linear time execution relative to the input length, unlike the Python one that uses backtracking and is exponential in worst-case scenarios. However, with RE2 you cannot use **lookaheads** ((?=...), (?!...)), **lookbehinds** ((?<=...), (?<!...)), or **backreferences** in the pattern itself (e.g., (abc)\1 to match "abcabc"). Note that capture groups and backreferences in replacements (${1}, ${2}) should still work fine. I wonder if some clients have used such advanced regexes up to now.

ge0Aja: imo we should not keep the same logic, if `DD_SCRUBBING_RULE` is not set (or set with an empty string) we should explicitly ignore it.

ge0Aja: Also I didn't get the comment about `RE2` why does it have to be RE2 ? 
There's a linked paragraph explaining performance diffs but it's just explaining what's happening underneath w/o stating why RE2 in the first place.

ndakkoune: > imo we should not keep the same logic, if `DD_SCRUBBING_RULE` is not set (or set with an empty string) we should explicitly ignore it.

It is already handled in `scrubbing.go:49`, or did you mean something else ?

ndakkoune: > Also I didn't get the comment about `RE2` why does it have to be RE2 ? There's a linked paragraph explaining performance diffs but it's just explaining what's happening underneath w/o stating why RE2 in the first place.

Go's regexp package only supports RE2. It's a language constraint. There's no alternative regex engine in the standard library.

ViBiOh (approved):

PR #1100, opened 2026-04-03
[AWSINTS-3449] feat(go-forwarder): add batching

ndakkoune:
### What does this PR do?
Add the batching logic.

### Testing Guidelines
`make test`

### Additional Notes
Increased batch size, items per batch and max item size to better reflect what's in the documentation/backend.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring_go/internal/processing/batching.go
+++ b/aws/logs_monitoring_go/internal/processing/batching.go
@@ -1,0 +1,82 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2026-Present Datadog, Inc.
+
+package processing
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"log/slog"
+
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/concurrent"
+)
+
+const (
+	maxItemSize      = 512 * 1024
+	maxBatchSize     = 4 * 1024 * 1024
+	maxItemsPerBatch = 400
+)
+
+func Batch[T any](ctx context.Context, in <-chan T, out chan<- []byte) error {
+	batch := make([][]byte, 0, maxItemsPerBatch)
+	batchSize := 0
+
+	flush := func() error {
+		if len(batch) == 0 {
+			return nil
+		}
+		payload := assembleBatch(batch)
+		batch = batch[:0]
+		batchSize = 0
+		return concurrent.SafeSender(ctx, out, payload)
+	}
+
+	for {
+		entry, ok, err := concurrent.SafeReader(ctx, in)
+		if err != nil {
+			return err
+		}
+		if !ok {
+			return flush()
+		}
+
+		data, err := json.Marshal(entry)
+		if err != nil {
+			slog.Warn("failed to marshal log entry, skipped", slog.Any("error", err))
+			continue
+		}
+
+		if len(data) > maxItemSize {
+			slog.Warn("log entry exceeds max item size, dropping",
+				slog.Int("size", len(data)),
+				slog.Int("max", maxItemSize),
+			)
+			continue
+		}
+
+		if batchSize+len(data) > maxBatchSize || len(batch) >= maxItemsPerBatch {
+			if err := flush(); err != nil {
+				return err
+			}
+		}
+
+		batch = append(batch, data)
+		batchSize += len(data)
+	}
+}
+
+func assembleBatch(entries [][]byte) []byte {
+	var buf bytes.Buffer
+	buf.WriteByte('[')
+	for i, entry := range entries {
+		if i > 0 {
+			buf.WriteByte(',')
+		}
+		buf.Write(entry)
+	}
+	buf.WriteByte(']')
+	return buf.Bytes()
+}
--- a/aws/logs_monitoring_go/internal/processing/batching.go
+++ b/aws/logs_monitoring_go/internal/processing/batching.go
@@ -15,32 +15,34 @@ import (
 )
 
 const (
-	maxItemSize      = 512 * 1024
-	maxBatchSize     = 4 * 1024 * 1024
-	maxItemsPerBatch = 400
+	maxItemSize      = 1 * 1024 * 1024
+	maxBatchSize     = 5 * 1024 * 1024
+	maxItemsPerBatch = 1000
 )
 
-func Batch[T any](ctx context.Context, in <-chan T, out chan<- []byte) error {
-	batch := make([][]byte, 0, maxItemsPerBatch)
-	batchSize := 0
+type Batcher[T any] struct {
+	batch     [][]byte
+	batchSize int
+	in        <-chan T
+	out       chan<- []byte
+}
 
-	flush := func() error {
-		if len(batch) == 0 {
-			return nil
-		}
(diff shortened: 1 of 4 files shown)


aws/logs_monitoring_go/internal/processing/batching.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"log/slog"
+
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/concurrent"
+)
+
+const (
```
ge0Aja, line 17: let's have this as a proper struct that accepts an input channel / interface with methods like Batch and flush. 
Also prefer to have flush as an independent method, not sure why you chose to create it as a var inside Batch

aws/logs_monitoring_go/internal/processing/batching.go
```diff
@@ -0,0 +1,82 @@
...
+	maxItemSize      = 512 * 1024
```
ge0Aja: could you double check on  the current limits that we have in the python crawler set for charcters or byte arrays ?

aws/logs_monitoring_go/internal/processing/batching.go
```diff
@@ -0,0 +1,82 @@
...
+		batchSize += len(data)
+	}
+}
+
+func assembleBatch(entries [][]byte) []byte {
+	var buf bytes.Buffer
+	buf.WriteByte('[')
+	for i, entry := range entries {
+		if i > 0 {
+			buf.WriteByte(',')
```
ge0Aja, line 91: this would still add a `,` at the end before the closing bracket `]` is this ok ?

ge0Aja:
general comment: if you've already merged the forwarding part please rebase this PR and use the batcher in the forward call

aws/logs_monitoring_go/internal/processing/batching.go
ndakkoune: > Also prefer to have flush as an independent method, not sure why you chose to create it as a var inside Batch

It was to leverage variable closures.

aws/logs_monitoring_go/internal/processing/batching.go
ndakkoune: It won't since the last instruction inside the loop is `buf.Write(entry)`.

ndakkoune:
> general comment: if you've already merged the forwarding part please rebase this PR and use the batcher in the forward call

@ge0Aja Imo don't think we should make the forwarding part explicitly call the batching one. By separating the two, we'll make them decoupled and let the future X forwarding (goroutine) workers read the `out` channel feeded by the Y (goroutines) batcher workers. The main pipeline stage will take care of wiring them together.

aws/logs_monitoring_go/internal/processing/batching.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/concurrent"
+)
+
+const (
+	maxItemSize      = 1 * 1024 * 1024
+	maxBatchSize     = 5 * 1024 * 1024
+	maxItemsPerBatch = 1000
+)
+
+type Batcher[T any] struct {
```
ge0Aja, line 23: Do we need this to be generic ? afaiu we only expect byte arrays

ndakkoune: For the next milestone, T will be either a CloudwatchLogEntry or a S3LogEntry, thus using generics.

ge0Aja:
> > general comment: if you've already merged the forwarding part please rebase this PR and use the batcher in the forward call
> 
> @ge0Aja Imo don't think we should make the forwarding part explicitly call the batching one. By separating the two, we'll make them decoupled and let the future X forwarding (goroutine) workers read the `out` channel feeded by the Y (goroutines) batcher workers. The main pipeline stage will take care of wiring them together.

I do think this is important as I failed to understand how you wanted to use the forwarding/batching parts together which could make some comments irrelevant. afaiu, we wanted to pass raw byte arrays to the batcher passing through the forwarder, but I'm a bit lost now since you're introducing a generic batcher so I'm not sure how's is this gonna be used. 
Maybe to clarify, I wanted to rebase this PR to understand how are these two parts will be put together. You don't have to necessarily to call one from the other as you mentioned if you want to keep them separate, which is fine.

aws/logs_monitoring_go/internal/processing/batching.go
ge0Aja: and what would be the benefit from that ?

aws/logs_monitoring_go/internal/processing/batching.go
ge0Aja: ok, I deliberately asked since the previous version expected byte arrays and batched them together. while in this version we're both marshalling and batch at once.

aws/logs_monitoring_go/internal/processing/batching.go
```diff
@@ -0,0 +1,98 @@
...
+
+func NewBatcher[T any](in <-chan T, out chan<- []byte) *Batcher[T] {
+	return &Batcher[T]{
+		batch: make([][]byte, 0, maxItemsPerBatch),
+		in:    in,
+		out:   out,
+	}
+}
+
+func (b *Batcher[T]) Batch(ctx context.Context) error {
```
ge0Aja, line 38: nit: rename to Start

aws/logs_monitoring_go/internal/processing/batching.go
ndakkoune: Np, I'm making a little drawing to show how the types/data flows through the stages to be on the same wave length for everyone

aws/logs_monitoring_go/internal/processing/batching.go
```diff
@@ -0,0 +1,98 @@
...
+	for {
+		entry, ok, err := concurrent.SafeReader(ctx, b.in)
+		if err != nil {
+			return err
+		}
+		if !ok {
+			return b.flush(ctx)
+		}
+
+		data, err := json.Marshal(entry)
```
ge0Aja, line 48: wondering where are we planning to perform compression since it seems this is batching raw json

ndakkoune: It is done just before the forwarding part. See https://github.com/DataDog/datadog-serverless-functions/blob/d7b9aed58039b2efc52e709eefc0db134d201a74/aws/logs_monitoring_go/internal/forwarding/forwarding.go#L45-L52

aws/logs_monitoring_go/internal/processing/batching.go
ge0Aja: yep it would be great (for a small tech design doc), I'll approve for now

ge0Aja (approved):

aws/logs_monitoring_go/internal/processing/batching.go
ndakkoune: By doing so we avoided the need of a generic struct and made the flushing logic in very few lines of code, but it's outdated.

aws/logs_monitoring_go/internal/processing/batching.go
ndakkoune: I deliberately increased the values to optimize the number of API calls we make to DD since network calls are by far the highest time consuming operations of the forwarder.

PR #1108, opened 2026-04-17
[AWSINTS-3506] feat(go-forwarder): add S3 pipeline (with multiline regex)

ndakkoune:
### What does this PR do?
Implements the S3 pipeline with multiline regex support.

### Additional Notes
- The Python Forwarder multiline regex behavior was only splitting on the provided regex AFTER it encountered a `\r\n\f` separator. This non-expected behavior has been fixed.
- Will declare high level constants for shared key between S3 and CW to enhance maintainability (as the `sourceCategory` constant) in a future PR the CloudTrail implementation (will probably move a lot of things)
 

### Types of changes

- [x] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


aws/logs_monitoring_go/internal/model/forwarder.go
ndakkoune, line 1: If you have a better name for this file and struct type. I hesitated with `lambda.go` (could be confusing since we're speaking of the forwarder specifically) and `context.go` (confusing with the context package).

aws/logs_monitoring_go/internal/model/s3.go
```diff
@@ -0,0 +1,25 @@
...
+// Copyright 2026-Present Datadog, Inc.
+
+package model
+
+type S3LogEntry struct {
+	Message        string     `json:"message"`
+	Source         string     `json:"ddsource"`
+	SourceCategory string     `json:"ddsourcecategory"`
+	Service        string     `json:"service"`
+	Tags           Tags       `json:"ddtags"`
```
ndakkoune, line 13: Will probably declare the struct tag as constants (shared with CW) in the future.

aws/logs_monitoring_go/internal/parsing/s3.go
```diff
@@ -0,0 +1,162 @@
...
+		return err
+	}
+
+	defer func() {
+		if err := body.Close(); err != nil {
+			slog.Warn("failed to close response body", slog.Any("error", err))
+		}
+	}()
+
+	var buf strings.Builder
```
ndakkoune, line 79: To append `\n` and `message` without big reallocations.

aws/logs_monitoring_go/internal/parsing/s3.go
```diff
@@ -0,0 +1,162 @@
...
+	scanner := bufio.NewScanner(body)
```
ndakkoune, line 80: The default buffer is 64KB. We may want to benchmark it we want to optimize, or put 1-5MB to handle big lines.
Will use a `json.Decoder` for the big CloudTrail events as single JSON line inside a bucket.

aws/logs_monitoring_go/internal/model/forwarder.go
ge0Aja: maybe `metadata` ?

aws/logs_monitoring_go/internal/model/forwarder.go
```diff
@@ -0,0 +1,34 @@
...
+	var metadata ForwarderMetadata
+
+	if lambdacontext.FunctionVersion != "$LATEST" {
+		metadata.Version = lambdacontext.FunctionVersion
+	}
+
+	if lc, ok := lambdacontext.FromContext(ctx); ok {
+		metadata.ARN = lc.InvokedFunctionArn
+	} else {
+		slog.Warn("failed to load lambda context, this should not happen in production. The code is either not running from AWS Lambda or context is broken.")
```
ge0Aja, line 30: I think we should return an error if we couldn't find the arn from the context

aws/logs_monitoring_go/internal/model/s3.go
ge0Aja: I didn't get this comment

ndakkoune: Instead of using json:"ddsource" in both Cw and S3 entries. I'll declare it as a const SourceTag = `ddsource` and replace it in both models.

aws/logs_monitoring_go/internal/parsing/s3.go
```diff
@@ -0,0 +1,162 @@
...
+		return fmt.Errorf("create S3 client: %w", err)
+	}
+
+	forwarderMetadata := model.GetForwarderMetadata(ctx)
+
+	for _, record := range s3Event.Records {
+		bucket := record.S3.Bucket.Name
+		key := record.S3.Object.URLDecodedKey
+
+		tags, service := getTagsAndService(*cfg)
```
ge0Aja, line 50: nit: we can pass cfg as is here if you're using it everywhere else w/o dereferencing

aws/logs_monitoring_go/internal/parsing/s3.go
```diff
@@ -0,0 +1,162 @@
...
+		if err := body.Close(); err != nil {
+			slog.Warn("failed to close response body", slog.Any("error", err))
+		}
+	}()
+
+	var buf strings.Builder
+	scanner := bufio.NewScanner(body)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if rc.multilineRegex != nil {
```
ge0Aja, line 83: nit: these could be on a single line

aws/logs_monitoring_go/internal/parsing/s3.go
```diff
@@ -0,0 +1,162 @@
...
+	defer func() {
+		if err := body.Close(); err != nil {
+			slog.Warn("failed to close response body", slog.Any("error", err))
+		}
+	}()
+
+	var buf strings.Builder
+	scanner := bufio.NewScanner(body)
+	for scanner.Scan() {
+		line := scanner.Text()
```
ge0Aja: if this breaks only on `\n` then I recall we had a support ticket back in the past as some clients used other symbols to break lines. 
Let's capture the python logic as is here

aws/logs_monitoring_go/internal/parsing/s3.go
```diff
@@ -78,8 +81,9 @@ func processS3Record(ctx context.Context, client S3APIClient, out chan<- model.S
 
 	var buf strings.Builder
 	scanner := bufio.NewScanner(body)
+	scanner.Split(split)
```
ge0Aja: nit: add a small code comment explain that this is setting

aws/logs_monitoring_go/internal/parsing/scanner.go
```diff
@@ -5,9 +5,52 @@
...
+	return s
+}
+
+func (s *Scanner) splitOnRegex(data []byte, atEOF bool) (advance int, token []byte, err error) {
+	if atEOF && len(data) == 0 {
+		return 0, nil, nil
+	}
+
+	loc := s.rg.FindIndex(data[1:])
+	if loc != nil {
```
ge0Aja, line 45: I think we should check index for out of bounds. Also let's add tests for longer matched regexes like sentences

aws/logs_monitoring_go/internal/parsing/scanner.go
ge0Aja: nit: you can return and skip the else and consider the default case as split on lines

ge0Aja (approved):

PR #1110, opened 2026-04-21
[AWSINTS-3514] feat(go-forwarder): add Kinesis invocation

ndakkoune:
### What does this PR do?
Adds the Kinesis handler responsible to wrap and redirect to the CloudWatch parsing logic.

### Additional Notes
- We encode back the `record.Kinesis.Data` in base64 because the json.unmarshal give us `events.KinesisRecord.Data` which are already base64 decoded but not ungzipped. We go this way to leverage `HandleCloudwatchLogs`.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring_go/internal/parsing/kinesis.go
+++ b/aws/logs_monitoring_go/internal/parsing/kinesis.go
@@ -24,21 +24,25 @@ func HandleKinesis(ctx context.Context, event json.RawMessage, cfg *config.Confi
 	}
 
 	for _, record := range kinesisEvent.Records {
-		cwEvent := events.CloudwatchLogsEvent{
-			AWSLogs: events.CloudwatchLogsRawData{
-				Data: base64.StdEncoding.EncodeToString(record.Kinesis.Data),
-			},
-		}
-
-		cwRaw, err := json.Marshal(cwEvent)
-		if err != nil {
-			return fmt.Errorf("marshal cloudwatch event from kinesis: %w", err)
-		}
-
-		if err := HandleCloudwatchLogs(ctx, cwRaw, cfg, out); err != nil {
+		if err := handleKinesisRecord(ctx, record, cfg, out); err != nil {
 			slog.WarnContext(ctx, "skipping kinesis record", "error", err)
 			continue
 		}
 	}
 	return nil
+}
+
+func handleKinesisRecord(ctx context.Context, record events.KinesisEventRecord, cfg *config.Config, out chan<- model.CloudwatchLogEntry) error {
+	cwEvent := events.CloudwatchLogsEvent{
+		AWSLogs: events.CloudwatchLogsRawData{
+			Data: base64.StdEncoding.EncodeToString(record.Kinesis.Data),
+		},
+	}
+
+	cwRaw, err := json.Marshal(cwEvent)
+	if err != nil {
+		return fmt.Errorf("marshal cloudwatch event from kinesis: %w", err)
+	}
+
+	return HandleCloudwatchLogs(ctx, cwRaw, cfg, out)
 }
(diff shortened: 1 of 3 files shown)


aws/logs_monitoring_go/internal/parsing/kinesis.go
```diff
@@ -0,0 +1,52 @@
...
+func HandleKinesis(ctx context.Context, event json.RawMessage, cfg *config.Config, out chan<- model.CloudwatchLogEntry) error {
+	var kinesisEvent events.KinesisEvent
+	if err := json.Unmarshal(event, &kinesisEvent); err != nil {
+		return fmt.Errorf("unmarshal: %w", err)
+	}
+
+	for _, record := range kinesisEvent.Records {
+		cwEvent := events.CloudwatchLogsEvent{
+			AWSLogs: events.CloudwatchLogsRawData{
+				Data: base64.StdEncoding.EncodeToString(record.Kinesis.Data),
```
ge0Aja: why do we have to re-encode here ?
also nit: I think this can be extracted to a function i.e. KinesisToCwEvent

aws/logs_monitoring_go/internal/parsing/kinesis.go
```diff
@@ -0,0 +1,52 @@
...
+			},
+		}
+
+		cwRaw, err := json.Marshal(cwEvent)
+		if err != nil {
+			return fmt.Errorf("marshal cloudwatch event from kinesis: %w", err)
+		}
+
+		entries, err := parseCloudwatchLogs(ctx, cwRaw, cfg)
```
ge0Aja: it feels weird to do a parse and collect behavior where everywhere else we're doing parse and emit. Why did we use HandleCloudwatchLogs here ?

ndakkoune: We're using `parseCloudwatchLogs` because the kinesis event is just a wrapped base64 gzipped cloudwatch event.

aws/logs_monitoring_go/internal/parsing/kinesis.go
ndakkoune: The `CloudwatchLogsRawData.Data` expects a base64 encoded string, but the one in `Kinesis.Data` is raw bytes. It forces us to base64 encode each record (I'll check if we can avoid that).

aws/logs_monitoring_go/internal/parsing/kinesis.go
ge0Aja: I don't think it's a good argument that we change the parsing behavior because of encoding. 
checking the python code it seems we're just restructuring the cloudwatch event using the same data bytes that we received from Kinesis. 
Here's it seems that we're re-encoding and not passing through the main entry point of cloudwatch logs parsing.

ge0Aja: `HandleCloudwatchLogs` will still call `parseCloudwatchLogs` internally so I'm not getting the difference here

ndakkoune: I see, I'll update that.

aws/logs_monitoring_go/internal/parsing/kinesis.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+func HandleKinesis(ctx context.Context, event json.RawMessage, cfg *config.Config, out chan<- model.CloudwatchLogEntry) error {
+	var kinesisEvent events.KinesisEvent
+	if err := json.Unmarshal(event, &kinesisEvent); err != nil {
+		return fmt.Errorf("unmarshal: %w", err)
+	}
+
+	for _, record := range kinesisEvent.Records {
+		cwEvent := events.CloudwatchLogsEvent{
```
ge0Aja: nit: extract this to a function

ge0Aja (approved):
lgtm, added a nit comment

PR #1116, opened 2026-05-03
[AWSINTS-3594] feat(go-forwarder): add CloudTrail

ndakkoune:
### What does this PR do?
Add CloudTrail events as supported event source.

### Additional Notes
Disparities with the Python Forwarder :
- Go's strictly typed language makes adding a CloudTrail aware field in `model.LogEntry` an ugly solution (e.g. adding a pointer embedded-struct along with a custom marshaller). The most elegant way I found to bypass this is to store the whole CloudTrail payload inside the `logEntry.Message` field instead of merging them and let the default parser inside the backend do the job (which works correctly after payload comparison).

### Types of changes

- [x] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


aws/logs_monitoring_go/internal/handling/cloudwatch.go
```diff
@@ -134,6 +134,13 @@ func (h CloudwatchHandler) newCloudwatchLogEntry(event events.CloudwatchLogsLogE
 	entry.Message = message
 	entry.ID = event.ID
 	entry.Timestamp = event.Timestamp
+
+	if h.cfg.Host == "" && entry.Source == sourceCloudtrail {
+		if host := cloudtrailHost(event.Message); host != "" {
```
RaphaelAllier: is there any value in chaining all of this ? instead of 2 "if"

aws/logs_monitoring_go/internal/handling/s3.go
```diff
@@ -75,32 +76,40 @@ func (h S3Handler) processRecord(ctx context.Context, client S3APIClient, out ch
...
+	isCloudTrail := cloudTrailRegex().MatchString(key)
+	if isCloudTrail {
+		records = decodeCloudTrail(body)
+	} else {
+		records = scan(body, h.cfg.S3MultilineLogRegex)
+	}
+
+	for message, err := range records {
+		if err != nil {
+			return err
```
RaphaelAllier, line 89: ❓ Do we want to error on the whole S3 record if there's one cloudtrail event that failed decoding in the whole message ?

aws/logs_monitoring_go/internal/handling/cloudtrail.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+const (
+	cloudTrailARNKey          = "arn"
+	cloudTrailUserIdentityKey = "userIdentity"
+)
+
+var cloudTrailRegex = sync.OnceValue(func() *regexp.Regexp {
+	return regexp.MustCompile(`\d+_CloudTrail(|-Digest|-Insight)_\w{2}(|-gov|-cn)-\w{4,9}-\d_(|.+)\d{8}T\d{4,6}Z(|.+)\.json\.gz$`)
+})
```
ViBiOh:
```suggestion
var cloudTrailRegex = regexp.MustCompile(`\d+_CloudTrail(|-Digest|-Insight)_\w{2}(|-gov|-cn)-\w{4,9}-\d_(|.+)\d{8}T\d{4,6}Z(|.+)\.json\.gz$`)
```

ndakkoune: @ViBiOh I lazy loaded this because we may have customers only forwarding cloudwatch (no S3) events.

ViBiOh: The cost of compiling the regex is not that high, and the compiler might optimizing it to do it at compile time. Adding a sync.OnceValue is adding more complexity that the supposed perf gain.

aws/logs_monitoring_go/internal/handling/cloudwatch.go
ndakkoune: I can't find a solution without this ugly bout of code. Since we need to access the message field from this function AND we lose type safety duo to the logEntry.Metadata any field.

aws/logs_monitoring_go/internal/handling/s3.go
ndakkoune: I assumed that if the first CloudTrail event is badly formatted, all the following ones will may also be (e.g. uploading a .json.gz cloudtrail-like file by hand). At the moment I get an error, I wouldn't have any idea about where my decoder /scanner is and continue proceeding will probably only lead to errors (if not panics).

PR #1140, opened 2026-05-29
feat(go-forwarder): add proxy and custom compression level

ndakkoune:
### What does this PR do?
Adds the custom compression level and the proxy support.

### Additional Notes
- Deprecating `DD_USE_COMPRESSION` in favor to `DD_COMPRESSION_LEVEL`
- Keep using the same logic for proxies as the Python forwarder (i.e. we let the underlying http library read from `HTTP_PROXY` and `HTTPS_PROXY` to configure it ootb; the CloudFormation Template as the responsibility to bind `DdHttpProxyURL` to the two variables).
- Renamed the package from client to sdkclients.
- Moved the WithCompression middleware logic inside Forwarder.Send() since we need a level and the key validation does not use it. 

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -7,39 +7,49 @@ package forwarding
...
+}
 
 type Forwarder struct {
-	cfg     *config.Config
-	client  *http.Client
-	storage string
+	cfg      *config.Config
+	client   *http.Client
+	storage  string
+	gzipPool *sync.Pool
```
ndakkoune, line 34: I did not find any other solution to use a gzip pool with a custom compression level if we use a package-level variable since we have to create the gzipWriter from the configuration. @ViBiOh

aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -22,6 +23,9 @@ func main() {
 	if err != nil {
```
ndakkoune, line 24: (will change this in the future by joining all the errors in the config, keeping it as is for now)

aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -22,6 +23,9 @@ func main() {
...
 		panic(err)
```
ge0Aja: nit: prefer log.fatal instead (recommended to do the same elsewhere)

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -66,27 +62,22 @@ func (c *Config) ValidateAPIKey(ctx context.Context) error {
 
 	slog.Debug("validating Datadog API key")
 
+	ctx, cancel := context.WithTimeout(ctx, httpclient.RequestTimeout)
+	defer cancel()
+
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, c.APIURL+"/api/v1/validate", nil)
 	if err != nil {
-		slog.Warn("failed to build API key validation request", slog.Any("error", err))
 		return nil
```
ge0Aja: I think we should return the error here not to silently drop it

aws/logs_monitoring_go/internal/config/config.go
```diff
@@ -89,8 +99,24 @@ func Load() (*Config, error) {
 
 func (c *Config) loadEnv() {
 	c.Site = envOrDefault(EnvSite, DefaultSite)
-	c.IntakeURL = envOrDefault(EnvURL, "https://http-intake.logs."+c.Site+"/api/v2/logs")
-	c.APIURL = envOrDefault(EnvAPIURL, "https://api."+c.Site)
+	c.SkipServerCertificate = envOrDefaultBool(EnvSkipServerCertificate, false)
+	c.UseHTTP = envOrDefaultBool(EnvUseHTTP, false)
+
+	scheme := "https"
```
ge0Aja: nit: i think protocol is a straight forward name

aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -7,39 +7,49 @@ package forwarding
...
+				w, _ := gzip.NewWriterLevel(nil, cfg.CompressionLevel)
+				return w
+			},
+		},
 	}
 }
 
 func (f Forwarder) Start(ctx context.Context, in <-chan model.LogEntry) error {
-	batches := make(chan []byte, MaxConcurrency)
+	batches := make(chan []byte, httpclient.MaxConcurrency)
```
ge0Aja, line 52: I didn't get why are we limiting the batches slice size to max concurrency ?

aws/logs_monitoring_go/internal/httpclient/client.go
```diff
@@ -15,25 +18,32 @@ const (
...
+	RequestTimeout       = 7 * time.Second
+	DefaultMaxAttempts   = 3
+	MaxConcurrency       = 5
 )
 
-var Client = newClient()
+var Client *http.Client
 
-func newClient() *http.Client {
+func Init(skipServerCertificate bool) {
```
ge0Aja: nit: prefer options pattern, easier to extend in the future if we want to pass other options

aws/logs_monitoring_go/internal/forwarding/forwarding.go
ge0Aja: I think it should be fine since we're using it in a single place

aws/logs_monitoring_go/internal/forwarding/forwarding.go
ndakkoune: My assumption was that the batcher (CPU bound) would produce more batches than the forwarding goroutines (network bound).

ge0Aja: ok, we could still use a larger buffer though doesn't have to be as strict

aws/logs_monitoring_go/internal/httpclient/tls.go
```diff
@@ -0,0 +1,18 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2026-Present Datadog, Inc.
+
+package httpclient
+
+import (
+	"crypto/tls"
+)
+
+type TLSOption func(*tls.Config)
```
ge0Aja, line 12: Imo it's an overkill to split that into a separate file, but just an opinion

ge0Aja (approved):
lgtm, a couple of nits

PR #1143, opened 2026-06-01
feat(go-forwarder): add S3 retry

ndakkoune:
### What does this PR do?
Adds S3 as a storing mechanism for failed events.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -30,11 +32,22 @@ var bufPool = sync.Pool{
 type Forwarder struct {
 	cfg      *config.Config
 	client   *http.Client
-	storage  string
```
ndakkoune, line 33: We can't make the Forwarder keep the storageTag anymore.
For "normal" events, we have the certainty that all the events share the same storage tag (e.g. s3 or cloudwatch or none (e.g. plain sns)).
However, for the retried events, each batch can be associated with any of these storage tag, requiring the forwarder to be created per batch in such case. To avoid that, I updated the send method to take the storageTag as input

aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -65,22 +79,24 @@ func (f Forwarder) Start(ctx context.Context, in <-chan model.LogEntry) error {
 	var mu sync.Mutex
 
 	for {
-		body, ok, err := concurrent.SafeReader(ctx, batches)
-		if err != nil {
-			mu.Lock()
-			errs = append(errs, err)
-			mu.Unlock()
-		}
```
ndakkoune, line 73: As discussed with @ViBiOh, if the channel is closed, we don't want to gather a context canceled error.

aws/logs_monitoring_go/internal/pipeline/pipeline.go
```diff
@@ -48,9 +50,45 @@ func Start(
...
+	forwarder := forwarding.NewForwarder(cfg, httpclient.Client, storage)
+
+	keys, listErr := storage.List(ctx)
+	if listErr != nil {
+		slog.WarnContext(ctx, "failed to list stored batches", slog.Any("error", listErr))
+		return
+	}
+
+	slog.InfoContext(ctx, "retrying stored batches", slog.Int("count", len(keys)))
+	for _, key := range keys {
```
ndakkoune: Will probably add concurrency there in the future since the whole retry logic is mostly network bound.

aws/logs_monitoring_go/internal/storing/s3.go
```diff
@@ -0,0 +1,109 @@
...
+	return &S3{client: client, bucket: bucket}
+}
+
+func (s *S3) Put(ctx context.Context, batch []byte, storageTag string) error {
+	invocationID := "unknown"
+	if lc, ok := lambdacontext.FromContext(ctx); ok {
+		invocationID = lc.AwsRequestID
+	}
+
+	datetime := time.Now().UTC().Format("2006/01/02/150405000")
```
ndakkoune, line 42: Using this new format for event storage allow us to retry older batches first (ListObjectsV2 returns keys in UTF-8 sorting order). The millisecond granularity should be enough to have batches implicitly sorted.

aws/logs_monitoring_go/internal/config/config.go
```diff
@@ -120,6 +124,8 @@ func (c *Config) loadEnv() {
 	c.UseFIPS = envOrDefaultBool(EnvUseFIPS, false)
 	c.Source = envOrDefault(EnvSource, "")
 	c.Host = envOrDefault(EnvHost, "")
+	c.S3RetryBucketName = envOrDefault(EnvS3RetryBucketName, "")
```
ge0Aja: nit: for string values Go default is already "" I think this applies for the rest of the configs as well

aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -30,11 +32,22 @@ var bufPool = sync.Pool{
...
-func NewForwarder(cfg *config.Config, client *http.Client, storage string) Forwarder {
+func NewForwarder(cfg *config.Config, client *http.Client, storage storing.Storage) Forwarder {
+	header := http.Header{
+		"DD-API-KEY":            []string{cfg.APIKey},
+		"DD-EVP-ORIGIN":         []string{"aws_forwarder"},
+		"DD-EVP-ORIGIN-VERSION": []string{config.ForwarderVersion},
+		"Content-Type":          []string{"application/json"},
+	}
+
+	if storage == nil {
```
ge0Aja: would it be ever the case that storage is not set ?

aws/logs_monitoring_go/internal/forwarding/forwarding.go
ge0Aja: couldn't we do a header override per batch if we store the storage information along the batch payload ?

aws/logs_monitoring_go/internal/forwarding/forwarding.go
ndakkoune: Afaik, the failed event storage is enabled by default (on the CloudFormation template) but can be disabled through the `DD_STORE_FAILED_EVENTS` environment variable.

aws/logs_monitoring_go/internal/config/config.go
ndakkoune: I kept the `envOrDefault` function for all the environment from which we expect strings because some may default to a value different than the zero value of the string type (e.g. port, url).

Though I agree with you that there may be a better solution (probably a mix of variable initialisation + `cmp.Or` or functional option pattern). I'll do that at the end of the milestone when I will make a FIPS refactor.

PR #1144, opened 2026-06-03
refactor(go-forwarder): decoupling config.Config 

ndakkoune:
### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring_go/cmd/forwarder/main.go
+++ b/aws/logs_monitoring_go/cmd/forwarder/main.go
@@ -66,8 +66,6 @@ func handleRequest(cfg *config.Config) func(ctx context.Context, event json.RawM
 		filterer := filtering.NewFilterer(cfg.FilterInclude, cfg.FilterExclude)
 		scrubber := scrubbing.NewScrubber(cfg.ScrubbingRegex, cfg.ScrubbingReplacement, cfg.ScrubIP, cfg.ScrubEmail)
 		handlerCfg := handling.Config{
-			Scrubber:            scrubber,
-			Filterer:            filterer,
 			Host:                cfg.Host,
 			Service:             cfg.Service,
 			Source:              cfg.Source,
@@ -94,6 +92,6 @@ func handleRequest(cfg *config.Config) func(ctx context.Context, event json.RawM
 			storage,
 		)
 
-		return pipeline.New(handlerCfg, forwarder).Start(ctx, parsed)
+		return pipeline.New(handlerCfg, scrubber, filterer, forwarder).Start(ctx, parsed)
 	}
 }
--- a/aws/logs_monitoring_go/cmd/forwarder/main.go
+++ b/aws/logs_monitoring_go/cmd/forwarder/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/config"
 	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/filtering"
 	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/forwarding"
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/handling"
 	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/httpclient"
 	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/parsing"
 	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/pipeline"
@@ -62,10 +63,37 @@ func handleRequest(cfg *config.Config) func(ctx context.Context, event json.RawM
 			return errors.New("no events to process")
 		}
 
-		filter := filtering.NewFilter(cfg.FilterInclude, cfg.FilterExclude)
+		filterer := filtering.NewFilterer(cfg.FilterInclude, cfg.FilterExclude)
 		scrubber := scrubbing.NewScrubber(cfg.ScrubbingRegex, cfg.ScrubbingReplacement, cfg.ScrubIP, cfg.ScrubEmail)
-		forwarder := forwarding.NewForwarder(cfg, httpclient.Client, storing.NewStorage(ctx, cfg))
+		handlerCfg := handling.Config{
+			Scrubber:            scrubber,
+			Filterer:            filterer,
+			Host:                cfg.Host,
+			Service:             cfg.Service,
+			Source:              cfg.Source,
+			Tags:                cfg.Tags,
+			S3MultilineLogRegex: cfg.S3MultilineLogRegex,
+			UseFIPS:             cfg.UseFIPS,
+		}
 
-		return pipeline.New(cfg, filter, scrubber, forwarder).Start(ctx, parsed)
+		forwarderCfg := forwarding.Config{
+			APIKey:           cfg.APIKey,
+			IntakeURL:        cfg.IntakeURL,
+			CompressionLevel: cfg.CompressionLevel,
+		}
+
+		storageOpts := storing.Options{Enabled: cfg.StoreOnFail, FIPS: cfg.UseFIPS, S3Bucket: cfg.S3RetryBucketName}
+		storage, err := storing.NewStorage(ctx, storageOpts)
+		if err != nil {
+			return fmt.Errorf("new storage: %w", err)
+		}
+
+		forwarder := forwarding.NewForwarder(
+			forwarderCfg,
+			httpclient.Client,
+			storage,
+		)
+
+		return pipeline.New(handlerCfg, forwarder).Start(ctx, parsed)
 	}
 }
--- a/aws/logs_monitoring_go/cmd/forwarder/main.go
+++ b/aws/logs_monitoring_go/cmd/forwarder/main.go
@@ -8,13 +8,18 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 
 	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/config"
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/filtering"
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/forwarding"
 	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/httpclient"
 	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/parsing"
 	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/pipeline"
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/scrubbing"
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/storing"
 
 	"github.com/aws/aws-lambda-go/lambda"
 )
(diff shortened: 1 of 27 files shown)


aws/logs_monitoring_go/internal/pipeline/pipeline.go
```diff
@@ -9,36 +9,55 @@ import (
...
 	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/parsing"
-	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/storing"
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/scrubbing"
 	"golang.org/x/sync/errgroup"
 )
 
-func Start(
+type Pipeline struct {
+	cfg       *config.Config
+	filter    *filtering.Filter
```
ge0Aja: nit:filterer, but not a big deal

aws/logs_monitoring_go/internal/pipeline/pipeline.go
```diff
@@ -9,36 +9,55 @@ import (
...
+func (p *Pipeline) Start(
 	ctx context.Context,
 	parsedEvents []parsing.ParsedEvent,
-	cfg *config.Config,
 ) error {
-	if len(parsedEvents) == 0 {
-		return errors.New("no events to process")
+	contentType := parsedEvents[0].ContentType
+	if contentType == parsing.ContentTypeRetry {
+		p.forwarder.Retry(ctx)
```
ge0Aja: we should return here

aws/logs_monitoring_go/internal/scrubbing/scrubber.go
```diff
@@ -6,53 +6,46 @@
...
-	ipReplacement    = "xxx.xxx.xxx.xxx"
-	emailReplacement = "xxxxx@xxxxx.com"
-)
-
 type scrubbingRule struct {
 	regex       *regexp.Regexp
 	replacement string
 }
 
+var ipRule = scrubbingRule{
```
ge0Aja, line 17: what's the reason behind making them global variables ?

aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -53,11 +58,14 @@ func handleRequest(cfg *config.Config) func(ctx context.Context, event json.RawM
 			return fmt.Errorf("parse: %w", err)
 		}
 
-		if len(parsed) == 1 && parsed[0].ContentType == parsing.ContentTypeRetry {
-			pipeline.Retry(ctx, cfg)
-			return nil
+		if len(parsed) == 0 {
```
ge0Aja, line 62: nit: we can do a nil check for early exit

aws/logs_monitoring_go/internal/config/config.go
```diff
@@ -48,66 +49,62 @@ const (
...
-	err = errors.Join(err, scrubbingErr)
-
-	filter, filteringErr := filtering.NewFilter(
-		envOrDefault(EnvIncludeAtMatch, ""),
-		envOrDefault(EnvExcludeAtMatch, ""),
-	)
-	err = errors.Join(err, filteringErr)
-	if err != nil {
-		return nil, err
+	var errs []error
```
ge0Aja, line 79: why are we collecting errors here ? Imo we should exit at first error

aws/logs_monitoring_go/internal/pipeline/pipeline.go
```diff
@@ -9,36 +9,55 @@ import (
...
 	eg, ctx := errgroup.WithContext(ctx)
 
-	forwarder := forwarding.NewForwarder(cfg, httpclient.Client, storing.NewStorage(ctx, cfg))
 	entries := make(chan model.LogEntry)
 
 	eg.Go(func() error {
 		defer close(entries)
 		for _, parsedEvent := range parsedEvents {
-			handler, err := handling.NewHandler(ctx, parsedEvent.ContentType, cfg)
+			handler, err := handling.NewHandler(ctx, parsedEvent.ContentType, p.cfg)
```
ge0Aja: let's pass the needed info only and not the whole config

aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -108,7 +109,36 @@ func (f Forwarder) Start(ctx context.Context, in <-chan model.LogEntry, storageT
 	return errors.Join(append(errs, <-producerErrCh)...)
 }
 
-func (f Forwarder) Send(ctx context.Context, payload []byte, storageTag string) error {
+func (f *Forwarder) Retry(ctx context.Context) {
+	// keys, listErr := f.storage.List(ctx)
+	// if listErr != nil {
+	// 	slog.WarnContext(ctx, "failed to list stored batches", slog.Any("error", listErr))
```
ge0Aja: return the error

aws/logs_monitoring_go/internal/config/config.go
ndakkoune: If the customer provides multiple wrong configurations, I think it's relevant to gather and return all of them so that he can update them simultaneously (instead of going back and forth to potentially figure out other misconfigurations).

aws/logs_monitoring_go/internal/scrubbing/scrubber.go
ndakkoune: To avoid compiling these regexes for each invocation (we moved the scrubber initialization in the request handler).

aws/logs_monitoring_go/cmd/forwarder/main.go
ndakkoune: I don't see why we would need to check if the parsed slice is nil since we can use len which handle both nil and zero-length slice ootb

aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -109,33 +115,32 @@ func (f *Forwarder) Start(ctx context.Context, in <-chan model.LogEntry, storage
...
-	// 		slog.WarnContext(ctx, "failed to delete successfully sent batch from storage, will lead to log duplication", slog.String("key", key), slog.Any("error", deleteErr))
-	// 		continue
-	// 	}
-
-	// 	slog.DebugContext(ctx, "batch sent successfully", slog.String("key", key))
-	// }
+func (f *Forwarder) Retry(ctx context.Context) error {
+	keys, listErr := f.storage.List(ctx)
+	if listErr != nil {
+		slog.WarnContext(ctx, "failed to list stored batches", slog.Any("error", listErr))
```
ndakkoune: (forgot to deleted this)

aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -62,10 +63,37 @@ func handleRequest(cfg *config.Config) func(ctx context.Context, event json.RawM
...
+			Source:              cfg.Source,
+			Tags:                cfg.Tags,
+			S3MultilineLogRegex: cfg.S3MultilineLogRegex,
+			UseFIPS:             cfg.UseFIPS,
+		}
+
+		forwarderCfg := forwarding.Config{
+			APIKey:           cfg.APIKey,
+			IntakeURL:        cfg.IntakeURL,
+			CompressionLevel: cfg.CompressionLevel,
```
ndakkoune, line 80: I'm wondering if I should put the storage alongside these fields since it is also derived from the main config

ge0Aja: let's keep them apart for now, on the other hand, why are you passing httpClient in the NewForwarder call ? I don't see any changing behavior from main for the client

aws/logs_monitoring_go/internal/forwarding/forwarding.go
ge0Aja: yeah we either log or return the error no need to do both

aws/logs_monitoring_go/internal/handling/handler.go
```diff
@@ -9,33 +9,52 @@ import (
...
 	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/sdkclient"
 )
 
 type Handler interface {
 	Handle(ctx context.Context, event json.RawMessage, out chan<- model.LogEntry) error
 }
 
-func NewHandler(ctx context.Context, ct parsing.ContentType, cfg *config.Config) (Handler, error) {
+type Config struct {
+	Scrubber            *scrubbing.Scrubber
```
ge0Aja: suggestion: keep the actors apart from the config as technically they're not

aws/logs_monitoring_go/cmd/forwarder/main.go
ndakkoune: https://github.com/DataDog/datadog-serverless-functions/blob/ead61e5008b259da129b62ca7f5db6cef84d4361/aws/logs_monitoring_go/cmd/forwarder/main.go#L38

We configured the TLS certificate skip in `main.go`. But more precisely, I passed the client here to have the `NewForwarder` function easily testable.

aws/logs_monitoring_go/internal/config/config.go
ge0Aja: if we want to keep the compile and collect pattern, I think to have more informative errors in the case of a collection it's better to add a prefix to each error we append to the slice.

aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -30,25 +30,32 @@ var bufPool = sync.Pool{
...
 
-func NewForwarder(cfg *config.Config, client *http.Client, storage storing.Storage) Forwarder {
+type Config struct {
+	APIKey           string
+	IntakeURL        string
+	CompressionLevel int
+}
+
+func NewForwarder(cfg Config, client *http.Client, storage storing.Storage) *Forwarder {
+	if storage == nil {
```
ge0Aja, line 47: can this be true ?

ndakkoune: In case the customer set `DD_FAILED_STORED_EVENT` to false.

aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -53,11 +59,39 @@ func handleRequest(cfg *config.Config) func(ctx context.Context, event json.RawM
...
+		}
+
+		storageOpts := storing.Options{Enabled: cfg.StoreOnFail, FIPS: cfg.UseFIPS, S3Bucket: cfg.S3RetryBucketName}
```
ge0Aja, line 83: nit: can be passed directly

aws/logs_monitoring_go/internal/config/apikey.go
```diff
@@ -79,7 +79,7 @@ func (c *Config) ValidateAPIKey(ctx context.Context) error {
 	defer httpclient.DrainClose(resp)
 
 	if resp.StatusCode == http.StatusForbidden {
-		slog.Warn("invalid Datadog API key", slog.String("url", "https://app."+c.Site+"/organization-settings/api-keys"))
+		slog.Warn("invalid Datadog API key")
```
ge0Aja, line 82: I think this should be logged as an error

aws/logs_monitoring_go/internal/forwarding/forwarding.go
ge0Aja, line 88: let's keep a controllable concurrency level with Go routines pool

aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -85,7 +92,7 @@ func (f Forwarder) Start(ctx context.Context, in <-chan model.LogEntry, storageT
 		}
 
 		eg.Go(func() error {
-			err := f.Send(ctx, body, storageTag)
+			err := f.send(ctx, body, storageTag)
```
ge0Aja, line 95: we can return error at first failure to send as we have a retry mech with backoff or set a threshold and return after that

aws/logs_monitoring_go/internal/forwarding/forwarding.go
ge0Aja, line 104: call backup function() where if there's no storage it will be a noop

ge0Aja (approved):
added comments to be addressed in the next PR

aws/logs_monitoring_go/internal/config/apikey.go
ndakkoune: Opted for an error return directly.

PR #1147, opened 2026-06-05
refactor(go-forwarder): stop sending on first error

ndakkoune:
### What does this PR do?
Refactors the storage creation into main and stop on first intake error.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/aws/logs_monitoring_go/internal/forwarding/forwarding.go
+++ b/aws/logs_monitoring_go/internal/forwarding/forwarding.go
@@ -9,6 +9,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
@@ -96,31 +97,19 @@ func (f *Forwarder) Start(ctx context.Context, in <-chan model.LogEntry, storage
 					onceSent.Do(func() { sentErr = err })
 				}
 
-				if err := f.store(ctx, batch, storageTag); err != nil {
-					return fmt.Errorf("store: %w", err)
+				if f.storage == nil {
+					break
+				}
+
+				if err := f.storage.Put(ctx, batch, storageTag); err != nil {
+					return fmt.Errorf("put: %w", err)
 				}
 			}
 			return nil
 		})
 	}
 
-	if err := eg.Wait(); err != nil {
-		return err
-	}
-
-	return sentErr
-}
-
-func (f *Forwarder) store(ctx context.Context, batch []byte, storageTag string) error {
-	if f.storage == nil {
-		return nil
-	}
-
-	if err := f.storage.Put(ctx, batch, storageTag); err != nil {
-		return fmt.Errorf("put: %w", err)
-	}
-
-	return nil
+	return errors.Join(eg.Wait(), sentErr)
 }
 
 func (f *Forwarder) Retry(ctx context.Context) error {
--- a/aws/logs_monitoring_go/internal/forwarding/forwarding.go
+++ b/aws/logs_monitoring_go/internal/forwarding/forwarding.go
@@ -14,7 +14,6 @@ import (
 	"log/slog"
 	"net/http"
 	"sync"
-	"sync/atomic"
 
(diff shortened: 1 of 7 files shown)


aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -70,49 +66,59 @@ func NewForwarder(cfg Config, client *http.Client, storage storing.Storage) *For
...
-			break
+		if err := batching.New().Batch(ctx, in, batches); err != nil {
+			return fmt.Errorf("batch: %w", err)
 		}
+		return nil
+	})
 
+	var stopSending atomic.Bool
+	for range httpclient.MaxConcurrency {
 		eg.Go(func() error {
```
ge0Aja, line 84: suggestion you can use https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit

aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -67,6 +66,7 @@ func NewForwarder(cfg Config, client *http.Client, storage storing.Storage) *For
 
 func (f *Forwarder) Start(ctx context.Context, in <-chan model.LogEntry, storageTag string) error {
 	eg, ctx := errgroup.WithContext(ctx)
+	eg.SetLimit(1 + httpclient.MaxConcurrency)
```
ndakkoune, line 70: eg.Go() blocks when the limit is set, had to workaround

aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -77,36 +77,39 @@ func (f *Forwarder) Start(ctx context.Context, in <-chan model.LogEntry, storage
...
+			return nil
 		})
 	}
 
-	return eg.Wait()
+	if err := eg.Wait(); err != nil {
+		return err
+	}
+
+	return sentErr
```
ndakkoune: I prioritized storing / batching error over the sending error. I can also join them if we want to highlight that both steps failed.

aws/logs_monitoring_go/internal/forwarding/forwarding.go
ge0Aja: I think this is the way to go

aws/logs_monitoring_go/internal/forwarding/forwarding.go
ge0Aja: I think we need to join return these errors

aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -77,36 +77,38 @@ func (f *Forwarder) Start(ctx context.Context, in <-chan model.LogEntry, storage
...
+					onceSent.Do(func() { sentErr = err })
 				}
 
 				if err := f.store(ctx, batch, storageTag); err != nil {
 					return fmt.Errorf("store: %w", err)
 				}
 			}
-
-			return sendingErr
+			return nil
```
ge0Aja, line 108: I think we don't need this return any more, we're stetting the error in the once and that's it no ? I just also realised that we could continue processing even if we don't have the store set, so we need to break out early when it's the case

ndakkoune: Got it for the storage check. However, the nil return is required due to the signature of the goroutines (send error => give them a try to store/save the batches, store error => let's stop everybody, we cannot do anything now). The `eg.Wait()` error is either a context expiration, a batcher error (i.e. marshal) or a storage failure.

ge0Aja: for a second pr, we can keep the ok, err pattern and break when we don't have a storage set

ge0Aja (approved):
lgtm, comment for next pr

PR #1151, opened 2026-06-09
refactor(go-forwarder): decouple key resolution, cache clients and share AWS config

ndakkoune:
This PR separates API key validation from config loading, uniformly initialize AWS SDK client configs and adds sdk clients caching.

It makes API key resolution logic easily testable.


aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -37,16 +39,11 @@ func main() {
...
-		log.Fatal(err)
-	}
-	err = cfg.ValidateAPIKey(context.Background())
-	if err != nil {
-		log.Fatal(err)
+	if err = apikey.Validate(context.Background(), httpclient.Client, cfg.APIURL, cfg.APIKey); err != nil {
+		if !cfg.StoreOnFail {
+			log.Fatal(fmt.Errorf("no failed event storage set, validate API key: %w", err))
+		}
+		slog.Error("API key validation", slog.Any("error", err))
```
ndakkoune, line 46: An invalid error can be due to a bad API key configuration, a network resolution failure, or a down validation endpoint.

In that case (and given that the failed event storage is set storage), I logged an error.

The other (probably better) way we could do is to validate the API key in the handleRequest: if it returns an error AND failed event storage is set, we cache the error keep the default execution flow (which would end up in the failed storage) and at the end we return the cached error to explicitly show that an error happened.

ge0Aja: let's not cache anything for now and keep the api validation call per invocation

aws/logs_monitoring_go/internal/config/config.go
```diff
@@ -126,3 +140,37 @@ func (c *Config) loadEnv() {
...
+
+	for _, resolver := range resolvers {
+		v, ok := os.LookupEnv(resolver.env)
+		if !ok {
+			continue
+		}
+
+		apiKeyConfigured = true
+		key, err := resolver.resolve(ctx, v)
+		if err != nil {
```
ge0Aja, line 159: do we need to still keep resolving once we have an error ?

aws/logs_monitoring_go/internal/sdkclient/awsconfig.go
```diff
@@ -0,0 +1,23 @@
...
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+)
+
+var AWSConfig = sync.OnceValues(func() (aws.Config, error) {
+	cfg, err := awsconfig.LoadDefaultConfig(context.Background())
```
ge0Aja, line 18: this should be used in init only

ge0Aja (approved):
lgtm, some comments on error handling

PR #1152, opened 2026-06-09
feat(go-forwarder): add SQS as failed event storage

ndakkoune:
This PR adds SQS as a failed event storage.

A refactor was needed on `storing.Storage`, which previously was not abstract enough to support diverse storage services. A new `storing.Batch` struct has been created to represent failed batches from the forwarder. `Batch.DeleteHandle` is a unique identifier of the underlying stored batch mechanism (key for S3, receiptHandler for SQS), used to delete the batch if the retry succeeds.

Diverging behaviors from the Python behavior are :
- Using `sqs.SendMessageBatch`, a batch equivalent of `sqs.SendMessage` to minimize API calls on `Store()`.
- `retry_prefix` which would equals to either logs, metrics or traces, and stored as a message attribute is deleted since we know only deals with logs.
- `function_prefix` which would equal to the sha1 of the forwarder's arn and stored as a message attribute is deleted. It was used to differentiate between message sent from the forwarder, from other the consumers. Sharing the same SQS queue between unrelated use-cases is a bad design and would highly enhance API costs by receiving noisy messages.

Side notes :
- We now have `DD_STORE_FAILED_EVENTS`, `DD_S3_BUCKET_NAME` and `DD_SQS_QUEUE_URL` which wraps the same logic around the retry mechanism: storing failed event. Imo, we should only keep one of these (since is one or the other), or a new `DD_FAILED_EVENT_STORAGE` which would directly take either an S3 bucket name or an SQS queue URL, if such identification is feasible.


aws/logs_monitoring_go/internal/batching/batcher.go
```diff
@@ -17,79 +16,137 @@ import (
...
+	for _, opt := range opts {
+		opt(b)
+	}
+
+	b.batch = make([]json.RawMessage, 0, b.maxItemsPerBatch)
+	return b
 }
 
-func (b *Batcher) Batch(ctx context.Context, in <-chan model.LogEntry, out chan<- []byte) error {
+func (b *Batcher) Start(ctx context.Context, in <-chan model.LogEntry, out chan<- json.RawMessage) error {
```
ndakkoune: If we want to fully decouple the batcher from the model, we could use any or json.RawMessage (if we move the marshaling to the caller) for the `in` channel type.
Tried to make the `Batcher` generic for type safety, but quite complicated because of the functional option pattern (i.e would have to make generic options which is quite verbose).

aws/logs_monitoring_go/internal/storing/sqs.go
```diff
@@ -0,0 +1,217 @@
...
+func (s *SQS) Store(ctx context.Context, storedBatch Batch) error {
+	var logs []json.RawMessage
+	if err := json.Unmarshal(storedBatch.Data, &logs); err != nil {
+		return fmt.Errorf("unmarshal: %w", err)
+	}
+
+	batcher := batching.New(
+		batching.WithMaxItemSize(maxSizePerSQSMessage),
+		batching.WithMaxBatchSize(maxSizePerSQSBatch),
+		batching.WithMaxItemsPerBatch(math.MaxInt),
```
ndakkoune: There is no (count) limit of items inside an message (l.114).

aws/logs_monitoring_go/internal/batching/batcher.go
```diff
@@ -17,79 +16,137 @@ import (
...
+			if err = concurrent.SafeSender(ctx, out, batch); err != nil {
 				return err
 			}
+			_ = b.add(item)
+		}
+	}
+	return nil
+}
+
+func (b *Batcher) StartSlice(items []json.RawMessage) ([]json.RawMessage, error) {
```
ndakkoune: Refactoring that because we could end up with twice the memory of all the cumulated logs. Going with an `iter.Seq` to yield values per values.

aws/logs_monitoring_go/internal/batching/batcher.go
```diff
@@ -6,90 +6,167 @@
...
+	defaultMaxItemsPerBatch = 1000
 )
 
 type Batcher struct {
-	batch     [][]byte
-	batchSize int
+	maxItemSize      int
+	maxBatchSize     int
+	maxItemsPerBatch int
+	batch            []json.RawMessage
```
ndakkoune: More explicit than [][]byte

aws/logs_monitoring_go/internal/batching/batcher.go
ndakkoune: It would also make the testing easier: no need to inject `model.LogEntry` anymore: only their. json representation

aws/logs_monitoring_go/internal/storing/s3_test.go
ndakkoune, line 1: Previously did a tons of state verification: this is not the purpose of mocking. Only verifying behaviors now (Is this function called X times ?)

aws/logs_monitoring_go/internal/storing/sqs.go
```diff
@@ -0,0 +1,115 @@
...
+
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/batching"
+	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/sdkclient"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/sqs"
+	"github.com/aws/aws-sdk-go-v2/service/sqs/types"
+)
+
+const (
+	maxSizePerSQSMessage = 1000 * 1024 // Overhead for other attributes than message body
```
ndakkoune, line 24: The maximum SQS message size is 1MiB.
However, we have to take all the attributes of `sqs.SendMessageInput` into account, so we're forced to keep an overhead.

aws/logs_monitoring_go/internal/batching/batcher.go
```diff
@@ -6,90 +6,167 @@
...
 }
 
-func (b *Batcher) Batch(ctx context.Context, in <-chan model.LogEntry, out chan<- []byte) error {
+func (b *Batcher) Start(ctx context.Context, in <-chan model.LogEntry, out chan<- json.RawMessage) error {
 	for {
-		entry, ok, err := concurrent.SafeReader(ctx, in)
-		if err != nil {
-			return err
-		}
+		v, ok, _ := concurrent.SafeReader(ctx, in)
```
ge0Aja, line 47: are we discarding the error because it could only be context cancellation ?

aws/logs_monitoring_go/internal/storing/storage.go
```diff
@@ -7,24 +7,40 @@ package storing
...
 }
 
 type Options struct {
 	S3Bucket string
+	SQSQueue string
 }
 
-func NewStorage(ctx context.Context, opts Options) (Storage, error) {
+func NewStorage(opts Options) (Storage, error) {
+	if opts.SQSQueue != "" {
```
ge0Aja, line 36: so it's going to be either sqs or s3 right ? O think we might want to indicate that in the readme

aws/logs_monitoring_go/internal/batching/batcher.go
```diff
@@ -6,90 +6,167 @@
...
 		}
 
-		if b.batchSize+len(data) > maxBatchSize || len(b.batch) >= maxItemsPerBatch {
-			if err := b.flush(ctx, out); err != nil {
+		if ok := b.add(item); !ok {
+			batch, constructed, err := b.construct()
+			if err != nil {
+				return err
+			}
+			_ = b.add(item)
```
ge0Aja, line 80: I didn't get that, we couldn't add the item so we entered this block and now we're adding it again while disregarding the result ?

aws/logs_monitoring_go/internal/batching/batcher.go
```diff
@@ -6,90 +6,167 @@
...
+
+			if err = concurrent.SafeSender(ctx, out, batch); err != nil {
 				return err
 			}
 		}
+	}
+	return nil
+}
+
+func (b *Batcher) StartYield(items []json.RawMessage) iter.Seq2[json.RawMessage, error] {
```
ge0Aja: nit: let's not use python terminology here Start is clear enough also I don't think you have to give the function a name in the return statement

aws/logs_monitoring_go/internal/batching/batcher.go
ndakkoune: Exactly. The caller would already have added the context cancellation error by itself, so we just turn off here.

aws/logs_monitoring_go/internal/batching/batcher.go
```diff
@@ -6,90 +6,167 @@
...
+	return func(yield func(json.RawMessage, error) bool) {
```
ge0Aja: I didn't get why we're returning a function that calls a function ?

aws/logs_monitoring_go/internal/batching/batcher.go
```diff
@@ -6,90 +6,167 @@
...
 
-		b.batch = append(b.batch, data)
-		b.batchSize += len(data)
+			if ok := b.add(item); !ok {
+				batch, constructed, err := b.construct()
+				if err != nil {
+					yield(nil, err)
+					return
+				}
+				_ = b.add(item)
```
ge0Aja: same questions about re-adding the item here

aws/logs_monitoring_go/internal/batching/batcher.go
ndakkoune: We know that the item is valid (otherwise it would have been dropped from `b.valid(...)`), but couldn't be added to the batch because it was near its full capacity. After `b.construct`, the data has been `b.reset()`: we know that the batch is valid and can now be added to the current (empty) data.

aws/logs_monitoring_go/internal/storing/storage.go
ndakkoune: Yes (that's the current behavior in the Python version too): SQS has the priority over S3 (even though customers shouldn't configure both).

aws/logs_monitoring_go/internal/batching/batcher.go
ndakkoune: That's also the convention when using an iterator in Go (see [here](https://pkg.go.dev/iter#hdr-Iterators)).

aws/logs_monitoring_go/internal/batching/batcher.go
ndakkoune: Basically the `iter.Seq2` is type which is a function that either returns true (if the sender sends a value) or false (if there is nothing to be sent anymore). The caller can `range` over it as if it was a slice of (batch, error) pairs.

```
type (
	Seq[V any]     func(yield func(V) bool)
	Seq2[K, V any] func(yield func(K, V) bool)
)
```

aws/logs_monitoring_go/internal/batching/batcher.go
ge0Aja: in this case why does it have to be surfaced to the caller ? your repeating yourself here and below I tihnk this can be all done in a single function right ?

aws/logs_monitoring_go/internal/batching/batcher.go
ndakkoune: Side note: I could have have reused the `batcher.Start` function that reads logs from a channel, batch them and outputs them to a channel, but for the SQS point of view if looks as if it was both the producer and consumer of the batcher:
1. Create two channels
2. Send all the logs inside the first
3. Receive all the batch in the second

aws/logs_monitoring_go/internal/batching/batcher.go
ge0Aja: discussed IRL let's try the two channel approach

aws/logs_monitoring_go/internal/batching/batcher.go
```diff
@@ -6,90 +6,167 @@
...
+func (b *Batcher) add(item json.RawMessage) bool {
+	if (b.maxItemsPerBatch != 0 && len(b.batch) >= b.maxItemsPerBatch) || b.batchSize+len(item)+1 > b.maxBatchSize {
+		return false
 	}
 
-	payload := assembleBatch(b.batch)
-	b.batch = b.batch[:0]
-	b.batchSize = 0
+	b.batch = append(b.batch, item)
+	b.batchSize += len(item) + 1 // ','
```
ge0Aja: nit: remove comment

aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -69,10 +69,10 @@ func (f *Forwarder) Start(ctx context.Context, in <-chan model.LogEntry, storage
 	eg, ctx := errgroup.WithContext(ctx)
 	eg.SetLimit(1 + httpclient.MaxConcurrency)
 
-	batches := make(chan []byte)
+	batches := make(chan json.RawMessage) // we should use a buffered channel here since we use a worker pool
```
ge0Aja: it seems it's not the case

aws/logs_monitoring_go/internal/forwarding/forwarding.go
```diff
@@ -113,33 +113,28 @@ func (f *Forwarder) Start(ctx context.Context, in <-chan model.LogEntry, storage
 }
 
 func (f *Forwarder) Retry(ctx context.Context) error {
-	keys, listErr := f.storage.List(ctx)
-	if listErr != nil {
-		return fmt.Errorf("list: %w", listErr)
+	if f.storage != nil {
```
ge0Aja:
```suggestion
	if f.storage == nil {
```

aws/logs_monitoring_go/internal/storing/s3.go
```diff
@@ -55,51 +57,63 @@ func (s *S3) Put(ctx context.Context, batch []byte, storageTag string) error {
...
 	if err != nil {
-		return nil, "", err
+		return Batch{}, fmt.Errorf("read body: %w", err)
 	}
 
-	return batch, out.Metadata[metadataStorageTagKey], nil
+	return Batch{
+		Data:       batch,
+		StorageTag: out.Metadata[metadataStorageTagKey],
+		DeleteKey:  aws.ToString(object.Key),
```
ge0Aja, line 109: isn't the key known to the getter ? why are we returning it with the batch ?

ndakkoune: We would have to return either return `(body io.Reader, storageTag string, err error)` or manually add the `storedBatch.DeleteKey=...` inside the loop.
Since we are forced to encapsulate the `GetObject` call into a function (since we have to `defer.Close()` the `out.Body`), I found it more simpler to encapsulate the whole batching logic inside `getBatch` instead to additionally set the `DeleteKey` inside the loop.

aws/logs_monitoring_go/internal/batching/batcher.go
```diff
@@ -13,40 +13,37 @@ import (
...
-	b := &Batcher{
-		maxItemSize:      defaultMaxItemSize,
-		maxBatchSize:     defaultMaxBatchSize,
-		maxItemsPerBatch: defaultMaxItemsPerBatch,
-		batchSize:        2, // '[' and ']'
+func NewConfig(maxItemSize, maxBatchSize, maxItemsPerBatch int) Config {
+	return Config{
+		maxItemSize:      maxItemSize,
+		maxBatchSize:     maxBatchSize,
+		maxItemsPerBatch: maxItemsPerBatch,
```
ndakkoune, line 27: I had three solution to reuse the `Batcher.Start()` method :
1. Keep the old functional option pattern and change the `in` channel from `model.LogEntry` to `any`. Too much restrictive: we would lose type safety and change the `Handler` signature.
2. Keep the old functional option pattern, make the Batcher and the Option both generic. Too much verbose, e.g. `batching.WithMaxItemSize[model.LogEntry](1*1024*1024)` three times and it does not provide any value since only the `Start` method leverage genericity.
3. Delete the old functional option pattern, use a config struct directly and make the batcher generic.

Opted for 3.

aws/logs_monitoring_go/internal/forwarding/forwarding.go
ge0Aja: @ndakkoune this is not a buffered channel are we changing it in this PR?

ndakkoune: I'll delete the comment but keep that in my personal notes.
I can explain you the reasoning I had when wrote the comment (its about context switching).

ge0Aja: let's just keep it as todo :p

aws/logs_monitoring_go/internal/storing/sqs.go
```diff
@@ -41,32 +43,51 @@ func (s *SQS) Store(ctx context.Context, batch Batch) error {
...
-		batching.WithMaxItemSize(maxSizePerSQSMessage),
-		batching.WithMaxBatchSize(maxSizePerSQSMessage),
-		batching.WithMaxItemsPerBatch(0),
-	)
-	for message, err := range batcher.StartYield(logs) {
-		if err != nil {
-			return fmt.Errorf("batching: %w", err)
+	eg, ctx := errgroup.WithContext(ctx)
+
+	in := make(chan json.RawMessage)
```
ge0Aja: we can't guarantee ordering on the intake side so let's not optimize for that reason

ge0Aja (approved):
lgtm i trust you on the comments

PR #1162, opened 2026-07-01
[AWSINTS-3801] feat(go-forwarder): add configuration for multiple lambda targets

ndakkoune:
Adds the DD_ADDITIONAL_TARGET_LAMBDAS environment variable support.
Diverging behavior from Python: it will now make the invocation fail if an error occurs (e.g. inexistant target, missing permissions).


aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -49,6 +50,12 @@ func main() {
 
 func handleRequest(cfg *config.Config) func(ctx context.Context, awsevent json.RawMessage) (any, error) {
 	return func(ctx context.Context, awsevent json.RawMessage) (any, error) {
+		if cfg.AdditionalLambdaTargets != nil {
```
ge0Aja: I think we can invoke additional targets in a separate go routine and log when error instead of blocking the whole invocation

aws/logs_monitoring_go/internal/forwarding/forwarding_test.go
```diff
@@ -119,62 +118,62 @@ func TestForwarder_Start(t *testing.T) {
...
-
-			if tc.wantErr != nil {
-				require.ErrorIs(t, err, tc.wantErr)
-				return
-			}
-			require.NoError(t, err)
-		})
-	}
-}
+// func TestForwarder_Start_Context(t *testing.T) {
```
ndakkoune, line 121: Muted this test for now

aws/logs_monitoring_go/cmd/forwarder/main.go
ndakkoune: Went for separate goroutine + return the error at the end of the invocation if an additional target is wrongly configured

aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -49,6 +52,20 @@ func main() {
...
 func handleRequest(cfg *config.Config) func(ctx context.Context, awsevent json.RawMessage) (any, error) {
 	return func(ctx context.Context, awsevent json.RawMessage) (any, error) {
+		var err error
+		var mu sync.Mutex
+		var wg sync.WaitGroup
+
+		if cfg.AdditionalTargets != nil {
+			wg.Go(func() {
+				if invokeErr := sdkclient.InvokeTargets(ctx, cfg.AdditionalTargets, awsevent); invokeErr != nil {
+					mu.Lock()
```
ge0Aja: let's note use mutex here we can set a separate err var and join the errors after we return from wg wait

ge0Aja (approved):

PR #1163, opened 2026-07-02
fix(go-forwarder): reduce binary size

ndakkoune:
See https://docs.aws.amazon.com/lambda/latest/dg/golang-package.html#golang-package-mac-linux.

_"You can use the optional lambda.norpc tag to exclude the Remote Procedure Call (RPC) component of the [lambda](https://github.com/aws/aws-lambda-go/tree/master/lambda) library. The RPC component is only required if you are using the deprecated Go 1.x runtime. Excluding the RPC reduces the size of the deployment package."_

### DIFF
--- a/aws/logs_monitoring_go/Makefile
+++ b/aws/logs_monitoring_go/Makefile
@@ -5,7 +5,7 @@ ZIP_NAME := forwarder.zip
 
 .PHONY: build
 build:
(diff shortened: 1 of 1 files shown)


RaphaelAllier (approved):

Sergio-Na (approved):

ndakkoune:
(previously broken tests)

PR #1165, opened 2026-07-03
tests(go-forwarder): add approval testing on non-external AWS API events

ndakkoune:
This PR adds approval testing. It reuses relevant test cases from the Python forwarder and adds new ones.

Note that approval testing was and is still not present for S3 events since we don't have the object content in-memory (we need to fetch it). It is doable but would require to either overwrite the GetS3 global variable during tests, or find another tricky solution (e.g. pass all clients to handleRequest through a struct and find a way to enrich it at runtime due to lazyloading).

### DIFF
--- a/aws/logs_monitoring_go/cmd/forwarder/main.go
+++ b/aws/logs_monitoring_go/cmd/forwarder/main.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"log"
 	"log/slog"
+	"net/http"
 	"sync"
 
 	"github.com/DataDog/datadog-serverless-functions/aws/logs_monitoring_go/internal/apikey"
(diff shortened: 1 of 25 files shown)


aws/logs_monitoring_go/cmd/forwarder/main.go
```diff
@@ -47,10 +48,10 @@ func main() {
 		slog.Error("API key validation", slog.Any("error", err))
 	}
 
-	lambda.Start(handleRequest(cfg))
+	lambda.Start(handleRequest(cfg, httpclient.Client))
 }
 
-func handleRequest(cfg *config.Config) func(ctx context.Context, awsevent json.RawMessage) (any, error) {
+func handleRequest(cfg *config.Config, client *http.Client) func(ctx context.Context, awsevent json.RawMessage) (any, error) {
```
ndakkoune: passing the client as parameter so that we can inject the httptest.Server.Client() during testing

aws/logs_monitoring_go/cmd/forwarder/main_test.go
ndakkoune, line 1: Use the same strategy as [go-fmt](https://cs.opensource.google/go/go/+/master:src/cmd/gofmt/gofmt_test.go) for approval testing.

aws/logs_monitoring_go/cmd/forwarder/main.go
ge0Aja: let's return a client instead of using a global var

ge0Aja (approved):

PR #1178, opened 2026-07-21
feat(go-forwarder): add minimal lambda enrichment

ndakkoune:
Adds lambda source identification and minimal enrichment to match previous lambda log structure.
Tags enrichment is done in the backend.


aws/logs_monitoring_go/internal/model/log.go
```diff
@@ -13,15 +13,16 @@ import (
...
+	Host           string     `json:"host,omitempty"`
+	ID             string     `json:"id,omitempty"`
+	Timestamp      int64      `json:"timestamp,omitempty"`
+	Message        string     `json:"message,omitempty"`
+	Service        string     `json:"service,omitempty"`
+	Source         string     `json:"ddsource"`
+	SourceCategory string     `json:"ddsourcecategory"`
+	Tags           Tags       `json:"ddtags"`
+	Metadata       any        `json:"aws"`
+	Lambda         *LambdaLog `json:"lambda,omitempty"`
```
ndakkoune, line 20: Minimal leak of lambda information inside the main struct here.

Could have used type embedding but it would require to change the `Handler.Handle` interface signature to a `chan any`.

aws/logs_monitoring_go/internal/model/lambda.go
```diff
@@ -0,0 +1,10 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2026-Present Datadog, Inc.
+
+package model
+
+type LambdaLog struct {
+	ARN string `json:"arn"`
```
ge0Aja, line 9: do we expect other information here ? why did you chose a struct instead of a simple string ?

aws/logs_monitoring_go/internal/handling/cloudwatch.go
```diff
@@ -177,3 +186,44 @@ func CloudwatchSource(logGroup, logStream string) string {
...
+		return strings.ToLower(m[1])
+	}
+
+	if _, name, found := strings.Cut(logGroup, logGroupLambda+"/"); found && name != "" {
+		return strings.ToLower(name)
+	}
+	return ""
+}
+
+func hasTag(tags model.Tags, prefix string) bool {
```
ge0Aja: if we're looking for the tag key then we could embed the `:` in the function and not pass it in the call

aws/logs_monitoring_go/internal/handling/cloudwatch.go
```diff
@@ -177,3 +186,44 @@ func CloudwatchSource(logGroup, logStream string) string {
...
 	return sourceCloudwatch
 }
+
+func enrichLambdaLog(entry *model.LogEntry, forwarderARN, logGroup, logStream string, cfg *Config) {
+	name := lambdaName(strings.ToLower(logGroup), logStream)
+	if name == "" {
+		return
+	}
+
+	prefix, _, found := strings.Cut(forwarderARN, "function:")
```
ge0Aja: could this happen here ?

aws/logs_monitoring_go/internal/handling/cloudwatch.go
```diff
@@ -177,3 +186,44 @@ func CloudwatchSource(logGroup, logStream string) string {
 	}
 	return sourceCloudwatch
 }
+
+func enrichLambdaLog(entry *model.LogEntry, forwarderARN, logGroup, logStream string, cfg *Config) {
+	name := lambdaName(strings.ToLower(logGroup), logStream)
```
ge0Aja: nit could be replaced with a string,bool return as done below and return when !ok

aws/logs_monitoring_go/internal/handling/cloudwatch.go
```diff
@@ -130,18 +133,24 @@ func (h cloudwatchHandler) newCloudwatchBaseEntry(data events.CloudwatchLogsData
...
 	}
+	source := cloudwatchSource(strings.ToLower(logGroup), logStream)
 
 	entry := model.NewLogEntry()
-	entry.Source = cmp.Or(h.cfg.Source, CloudwatchSource(strings.ToLower(logGroup), logStream))
+	entry.Source = cmp.Or(h.cfg.Source, source)
 	entry.Host = cmp.Or(h.cfg.Host, logGroup)
 	entry.Metadata = metadata
+
+	if entry.Source == sourceLambda {
```
ge0Aja, line 142: should we actually enrich if we used the override source from the config ? I think we should only enrich when we figure that the extracted source is lambda and disregard the source override

aws/logs_monitoring_go/internal/handling/cloudwatch.go
```diff
@@ -177,3 +186,44 @@ func CloudwatchSource(logGroup, logStream string) string {
 	}
 	return sourceCloudwatch
 }
+
+func enrichLambdaLog(entry *model.LogEntry, forwarderARN, logGroup, logStream string, cfg *Config) {
```
ge0Aja: if we're passing the entry, I assume we already have the concatenated tags appended same goes for the host. Do we need to pass the config here ?

aws/logs_monitoring_go/internal/model/lambda.go
ndakkoune: 👋 The output we expect is this one :
```json
{
  "aws": {
    "awslogs": {
      "logGroup": "",
      "logStream": "",
      "owner": ""
    },
    "invoked_function_arn": ""
  },
  "ddsource": "",
  "ddtags": "",
  "host": "",
  "id": "",
  "lambda": {
    "arn": ""
  },
  "service": ""
}
```

The `arn` is under a `lambda` key, so we need a struct to have this nesting.

aws/logs_monitoring_go/internal/handling/cloudwatch.go
ndakkoune: It shouldn't (`found` should always be true).

aws/logs_monitoring_go/internal/handling/cloudwatch.go
ndakkoune: The Python version bypasses the lambda logs enrichment when the source is overriden because of [this line](https://github.com/DataDog/datadog-serverless-functions/blob/3395a3b2d003aea3ceb913edab3141a2eb566426/aws/logs_monitoring/steps/handlers/awslogs_handler.py#L61-L62).

aws/logs_monitoring_go/internal/handling/cloudwatch.go
ndakkoune: good catch

aws/logs_monitoring_go/internal/handling/cloudwatch.go
ge0Aja: then maybe there's no need for that extra check

aws/logs_monitoring_go/internal/handling/cloudwatch.go
ge0Aja: but we're not bypassing it here are we ?

aws/logs_monitoring_go/internal/handling/cloudwatch.go
ndakkoune: I'll keep it as a safe guard

aws/logs_monitoring_go/internal/handling/cloudwatch.go
ndakkoune: This [line](https://github.com/DataDog/datadog-serverless-functions/pull/1178/changes#diff-e0af235db15f7ffcd64b1a401c0708589335715ac45438a8d8b84e26123869ffL135) prioritize the custom source, so we wont enter the if check.

ge0Aja (approved):

PR #1186, opened 2026-07-30
[AWSINTS-3784] feat(go-forwarder): add CloudFormation template

ndakkoune:
Adds Cloudformation Go template.

Most noticeable change is the deletion of the layer path. We will have to see with the team if we opt for another way of releasing (e.g. upload compiled zip to S3 in each region?).


aws/logs_monitoring_go/template.yaml
```diff
@@ -274,8 +190,9 @@ Parameters:
     Description: Set to true to enable a scheduled forwarder invocation (via AWS EventBridge) to process stored failed events.
   DdScheduleRetryInterval:
     Type: Number
-    Default: 6
-    Description: Interval in hours for scheduled forwarder invocation (via AWS EventBridge).
+    Default: 360
+    MinValue: 2
+    Description: Interval in minutes for scheduled forwarder invocation (via AWS EventBridge).
```
ndakkoune, line 195: Updated for smaller granularity here

aws/logs_monitoring_go/template.yaml
```diff
@@ -442,43 +311,24 @@ Rules:
...
+      Handler: bootstrap
+      Code:
+        S3Bucket: !Ref SourceZipBucket
+        S3Key: !If
+          - SetSourceZipKey
+          - !Ref SourceZipKey
+          - !Sub
             - "aws-dd-forwarder-${DdForwarderVersion}.zip"
             - {DdForwarderVersion: !FindInMap [Constants, DdForwarder, Version]}
-        - ZipFile: " "
```
ndakkoune, line 479: Not possible for non Python code

